PR #17, opened 2017-06-02
Adding parse order capture

### LINKED ISSUE #8: Inter-argument order
The poster [here](https://softwarerecs.stackexchange.com/questions/31766/boostprogram-options-command-line-parser-alternative-that-can-process-argume) describes a specific issue that should be easy to solve with a small modification to CLI11.

Planned solution: store a vector of pointers to Options as it parses. Then this could be solved like this:

```cpp
for(Option* opt : app.parse_order()) {
    if(opt == arg1)
        \\ use value of arg1
}
```

henryiii:
This fixes #8 by adding the original parse order.


henryiii:
Example added solving https://softwarerecs.stackexchange.com/questions/31766/boostprogram-options-command-line-parser-alternative-that-can-process-argume (IMO) fairly elegantly.

PR #24, opened 2017-08-22
Fix #23: Respect fallthrough_ in _valid_subcommand

### LINKED ISSUE #23: [Bug] Parser assumes positional arguments which happen to match subcommand names are subcommands
Consider this demo program:
```c++
#include <string>
#include <iostream>

#include "CLI11.hpp"

using std::string;
using std::cout;
using std::endl;

    CLI::App app{"Buggy demo app"};
    app.require_subcommand();

    auto foo = app.add_subcommand("foo", "Foo Description", false);
    auto bar = app.add_subcommand("bar", "Bar Description", false);
    string baz;
    bar->add_option("baz", baz, "Baz Description", true)->required();

    foo->set_callback([] {
        cout << "Bad" << endl;
    });
    bar->set_callback([&baz] {
        cout << "Good; baz was " << baz << endl;
    });

    try {
        app.parse(argc, argv);
    } catch (CLI::ParseError& e) {
        return app.exit(e);
    }
}
```

The app has two subcommands, `foo` and `bar`, where `bar` takes a required positional argument named `baz`. Suppose I want to run `bar` with `baz` set to "qux":
```
$ ./demo bar qux
Good; baz was qux
```
This works great! But suppose instead I want `baz` set to "foo":
```
$ ./demo bar foo
ERROR: RequiredError: baz
Foo Description
Usage: ./a.out foo
```
The parser has seen "foo" and given up parsing arguments for `bar`, but it instead context switched to the `foo` subcommand.

There might be a use case for running multiple subcommands in the same invocation of the program, but the behavior we're seeing here is clearly incorrect, especially since there is no apparent way to fix it: adding `app.fallthrough(false)` and/or `bar->fallthrough(false)` do not solve the issue.

The only workaround I see so far is to convert `baz` to a named argument:
```c++
    bar->add_option("--baz", baz, "Baz Description", true)->required();
```
This allows me to slip it by the parser correctly:
```
$ ./demo bar --baz foo
Good; baz was foo
```

nathanielhourt:
_valid_subcommand checks whether its argument appears to be a valid subcommand name or not; however, if it doesn't recognize the name, it always checks if its parent does. As described in in issue #23, this can cause incorrect behavior. To avoid this, check if fallthrough is disabled first, and do not consult the parent's known subcommands if fallthrough is disabled.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -857,7 +857,7 @@ class App {
         for(const App_p &com : subcommands_)
             if(com->check_name(current))
                 return true;
-        if(parent_ != nullptr)
+        if(parent_ != nullptr && fallthrough_)
             return parent_->_valid_subcommand(current);
         else
             return false;


henryiii:
I have an idea, let me try that first...

PR #33, opened 2017-09-10
Adding simple flag callback

### LINKED ISSUE #32: Version flag/option
So far to get a simple flag that triggers it's callback first time it's detected I had to do this workaround.

```
auto ver = app.add_option("-v,--version", (CLI::callback_t)show_version, "Show version");
ver->set_custom_option("", 0); 
```

It would be much more user friendly to have it similar to `add_flag` syntax but just with a callback.
No sure if it qualifies to be a separate function as `add_config`.

```
Option *add_version(std::string name, callback_t callback, std::string description = "") {
    Option *opt = add_option(name, callback, description, false);
    opt->set_custom_option("", 0);
    return opt;
}
```

henryiii:
This should fix #32 in a general way, I think.
```cpp
.add_flag_functional("--name", [](size_t count){...});
```

C++14 shortcut:
```cpp
.add_flag("--name", [](size_t count){...});
```

This can't be added to C++11 since the compiler can't tell the difference between `std::string` and `std::function<void(size_t)>` for a `const char *` argument until C++14 (or C++11 mode on a C++14 compiler, since this was back-fixed as a bug, but that's too specific to provide `#if`'s for).


infinity0n3:
I can confirm that it works now. Would have been nicer to have the add_flag syntax but I have to use c++11.
Good thing is that it's a one liner now.

Thank you!

PR #34, opened 2017-10-27
File example

henryiii:
This is based on the Gitter discussion from @lczech. It shows an example of how you could break a subcommand into separate files. I've cleaned it up a bit with shared pointers (unique would be just fine in C++14, note added).


lczech:
Nice, thanks for picking up this example!

henryiii:
Thanks! I'll look into the copy vs. move capture of lambdas to see if there's anything I can do there, and drop the extra C++14 comment.

Issue #36, opened 2017-11-04
allow_extras/prefix_command and subcommands


lambdafu:
allow_extras/prefix_command and subcommands don't mix well. The main problem is that allow_extras/prefix_command has to be set on the parent, rather than the subcommand, for it to have _any_ effect. This implies:

* If no subcommand is given, extra arguments are ignored and parsing stops at the first unknown positional.  There is no way to get allow_extra/prefix_command only for subcommands but not for the parent.
* You can't have some subcommands that are the prefix of an operation, and some subcommands that are not.

To regain composability, allow_extras/prefix_command should take affect only at the level they are given.

henryiii:
Can you put up a little pseudo code example?

lambdafu:
Sure!

```
#include <CLI/CLI.hpp>

int main(int argc, char** argv) {
  CLI::App app{"test"};
  // [1] This will allow "foo -v" as well as "cmd -w foo" etc.                                                                                                                                         
  // app.allow_extras();
  // [2] Even weirder:
  // app.prefix_command();                                                                                                                                                             

  int v = 0;
  app.add_flag("-v", v);
  CLI::App* cmd = app.add_subcommand("cmd", "Invoke cmd");
  // [5] cmd->set_callback([](){ std::cout << "cmd\n"; throw CLI::Success(); });                                                                                                                       
  int w = 0;
  cmd->add_flag("-w", w);
  // [3] No effect: cmd->allow_extras();                                                                                                                                                               
  // [4] No effect: cmd->prefix_command();                                                                                                                                                             

  try {
    std::vector<std::string> rest = app.parse(argc, argv);
    for (auto x : rest)
      std::cout << "REST: " << x << "\n";
    std::cout << "v: " << v << "\n";
    std::cout << "w: " << w << "\n";
  } catch (const CLI::ParseError &e) {
    return app.exit(e);
  }
}
```

Without modifications, this program will behave as expected:

```
$ ./a.out -v cmd -w
v: 1
w: 1
$ ./a.out -v foo cmd -w
ERROR: ExtrasError: [foo]
$ ./a.out -v cmd -w foo
ERROR: ExtrasError: [foo]
```

Say I want to allow extra positional arguments to the subcommand, but not to the main program. My intuitive approach would be to set "allow_extras" on the subcommand ([3] above). I'd then expect this to work:

```
$ ./a.out -v cmd -w foo
ERROR: ExtrasError: [foo]
```

[...]

PR #37, opened 2017-11-16
Extras handling refactor

henryiii:
In the interests of solving #36, this PR refactors the way extras are handled. The implementation:
* `parse` no longer returns a vector (not well supported anyway, for example with `CLI11_PARSE`
* Added `remaining()` and `remaining_size()`, which provide access into the unparsed options. They support a "recurse" bool flag that causes them to collect all subcommand arguments, too. They are not "reversed".
* `allow_extras` and `prefix_command` are now valid on subcommands, and inherit the setting from the parent upon creation (like the other options)

Oddities that remain:
* The vector form of parse still expects a "reversed" vector; it updates itself with the remaining options, however. Reordering might be unintentionally possible with multiple subcommands.

Will need to maybe add more tests, and needs docs and changelog.


henryiii:
@lambdafu , let me know how this works (and maybe the other two PRs, too) :)

lambdafu:
Looks good!  I'll implement this in a larger program and see if there are any issues remaining. Thanks!

PR #42, opened 2017-11-20
Allow to set a footer in the description

lambdafu:
Let's see what you think :)  Setting a footer for a bug report address or other notes is common enough that I think it should be supported without subclassing.  If you agree, here is a simple implementation that does the trick for me.


henryiii:
I think these both look good. Will check and merge tomorrow. Don't worry about the formatting check, I can run clang-format.

henryiii:
Thanks! I already have a use case, I believe, in a program I'm working on! 👍

PR #43, opened 2017-11-20
Allow to customize help flag.

lambdafu:
Another little helper: Allows to customize the help flag without fully exposing help_ptr_.  This makes it possible to translate the string, for example. Inspired by taywee/args.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,8 +132,6 @@ app.add_set(option_name,
 
 app.add_set_ignore_case(... // String only
 
-app.add_help_flag(name, optional_discription);
-
 App* subcom = app.add_subcommand(name, discription);
 ```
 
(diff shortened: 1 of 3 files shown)


henryiii:
Does it really need to throw if the help pointer is already set? Couldn't it just deregister the old help pointer instead? That way, someone who just wants to change the help string or flag name doesn't have to know about the `false` argument to `App`.

henryiii:
Other than the answer to that question, it looks good. I added a `///` comment, those are rendered into Doxygen API docs, and moved to using this internally as well.

lambdafu:
Deregistering would be better, but I don't know how to do that :)

henryiii:
Okay, I can add that, just a second..

henryiii:
Thanks!

PR #46, opened 2017-11-20
Support groups for subcommands.

lambdafu:
I like myself some groups for subcommands :).  If you look at "git --help", for example, that obviously looks better than "svn --help".  I chose to not order them alphabetically, but in insertion order.  I _think_ that this would also be better for option groups (instead of ordering them alphabetically), but this pull request does not change that.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -121,6 +121,9 @@ class App {
     /// -1 for 1 or more, 0 for not required, # for exact number required
     int require_subcommand_ = 0;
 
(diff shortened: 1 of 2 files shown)


lambdafu:
By the way, this also allows one to override the hard-coded "Subcommands" string. I am slowly eating away at those hard-coded untranslatable strings as a side-effect :)

henryiii:
This is very cool! It also works really well for my planned work on the inheritance of values. I'll open an issue describing my plans in a minute.

PR #49, opened 2017-11-21
Fix ExtrasError.

lambdafu:
I noticed that ExtrasError wasn't showing the extra argument in a subcommand.  So I dug deeper, and I could identify a bug in remaining(true), which was overwriting the miss_list instead of extending it.  So, I added some tests and fixed it.
But then I noticed some bargled logic in triggering ExtrasError: Firstly, there is a mismatch between remaining(true) and remaining_size() (without true). Secondly, it didn't handle allow_extras_ and prefix_command_ on a per-subcommand basis.  I tried a couple of things, but the easiest fix was to just let each subcommand fail on its own, by not using recurse=true at all.
As recurse=true is not used anymore (and because it does not handle allow_extras_ and prefix_command_), the recurse=true case could be removed from the code base.  My patch doesn't do that, though, to keep the change minimal, in case there are some other plans or cases I missed.  My recommendation would be to remove it if it doesn't serve any other purpose.  I am not firm on the logic behind the fallthrough_ cases, so it may be that I am missing something.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -954,7 +954,7 @@ class App {
         if(recurse) {
             for(const App_p &sub : subcommands_) {
                 std::vector<std::string> output = sub->remaining(recurse);
-                miss_list.assign(std::begin(output), std::end(output));
+                std::copy(std::begin(output), std::end(output), std::back_inserter(miss_list));
             }
         }
         return miss_list;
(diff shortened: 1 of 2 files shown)


henryiii:
I'll look into this in more detail when I get a chance in the next day or two, but not surprised - that code was added very recently. I really need a series of more complex subcommsnd tests that mix features...

henryiii:
I've added a fix for the subcommand ordering, as well.

henryiii:
I think `recurse=true` is still useful if someone just wants to collect all extras, possibly for debugging. I've added a test that makes sure it can rebuild the command line in order - which also fixes the callback order. I'm not sure if it has a practical use other than for diagnostics, since I would think users would always want to have the subcommands, as well. I agree with you that it should not be used if `allow_extras` is false - that should fail. Let me know if you have any thoughts!

By the way, the main purpose for fallthrough is to allow something like this:

```bash
./run_model model1 --common-opt --model1-opt
```

to work, where common options can be set on `run_model` but be given to subcommands. I'm not sure if the interaction with other models like prefix commands is perfect yet.

henryiii:
Thanks!

henryiii:
FYI, Windows requires `#include <iterator>` to use `back_inserter`. Fixed in master. :)

PR #61, opened 2017-11-30
Add test target for Clang 5.0 and C++14.

lambdafu:
I got a build failure in neopg for clang5 and c++14:

```
CLI11/include/CLI11.hpp:234:12: error: 
      no member named 'enable_if_t' in namespace 'std'; did you mean
      'enable_if'?
using std::enable_if_t;
      ~~~~~^~~~~~~~~~~
           enable_if
/usr/bin/../lib/gcc/x86_64-linux-gnu/4.8/../../../../include/c++/4.8/type_traits:1766:12: note: 
      'enable_if' declared here
    struct enable_if 
           ^
```

Let's see if we can trigger this here, too! :fire: :fire_engine: :potable_water:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,17 @@ matrix:
   - compiler: clang
     addons:
       apt:
+        sources:
+        - ubuntu-toolchain-r-test
+        - llvm-toolchain-trusty-5.0
+        packages:
+        - clang++-5.0
+    env:
+    - COMPILER=5.0
+    - CLI_CXX_STD=14
+  - compiler: clang
+    addons:
+      apt:
         packages:
         - clang-3.9
         - clang-format-3.9
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,7 @@ matrix:
     - COMPILER=4.7
   - os: osx
     before_install:
-    - python -m ensurepip --default-pip
+    - python -m ensurepip --user
 install:
 - python -c 'import sys; print(sys.version_info[:])'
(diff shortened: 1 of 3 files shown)


henryiii:
You are building with Clang 5 but linking to a GCC 4.8 stdlib? GCC 4.8 doesn't have `enable_if_t`, but c++14 mode expects it.

lambdafu:
oops

henryiii:
I'll check and see if there's a good option here, the check against the C++ version is what is causing it to expect `enable_if_t` to be there.

lambdafu:
I'll see what's going on in the travis target, and why it doesn't pick up libc++ by default.  I certainly didn't intend to use gcc 4.8 std.  Blah!

lambdafu:
I can't even reproduce this in an Ubuntu trusty docker container, so this seems to be a travis-specific problem. Sorry for the noise.  Shall we close here?

henryiii:
It's working now, I added a fix for the macOS builds now missing pip.

lambdafu:
Ok, thanks! I am new to navigating all these subtle toolchain challenges, so thank you for your patience!

lambdafu:
Here is some more information for the archive why this happened: My understanding is that on all major GNU/Linux distributions, clang will use the GNU C++ standard library by default to maintain ABI compatibility.  Basically, you can't mix libraries compiled against different standard libraries, so they can't just switch over.  And the toolchain builds on the Travis CI images use a specific version of libstdc++ that is very old.  For example, std::regex is also not supported there.

On MacOS since Maverick, clang will use libc++.

Issue #65, opened 2017-12-15
Adding the option of printing the help text as comments when calling "config_to_str()"


SkyToGround:
I am using CLI11 for an application with a lot of options thus I am making use of the ini-file feature to store the options. To further ease the use of a lot of options it would be nice to have the help text printed as comments when calling "App::config_to_str()". I would be willing to add this functionality if you accept the feature.

henryiii:
Sure, it would probably be best to have it optional. And it would need to handle multi line help. But sounds good!

SkyToGround:
I have a simple (and naive) implementation, with two unit tests, that you can [find here](https://github.com/SkyToGround/CLI11/commit/97b1def525c42db2257359e2b6d8800e9376b94a). This implementation does not take into account any kind of line breaking as it simply writes what is in ``Option::description_``. If you can give me an example of the expected output from a multi line description I will implement that as well before submitting a pull request.

henryiii:
Feel free to make it a WIP pull request. (Add WIP to the title; it doesn't block merging like it would on GitLab, but it still makes it obvious it's not ready to merge.)

It looks good. As a simple first approximation, I'd just search through the string, find the multilines, and add `; ` to each. This is not really handled yet for the regular help output; ideally, the leader to add would be settable, and the normal help could add spaces as needed.

Please make the WIP pull request, then I might be able to add the function I just described.

SkyToGround:
Yes I was slightly confused by the lack of handling of printing of multiple help lines in the current implementation. I am still not entirely sure of what you mean. Thus do you suggest that we:

1. Replace every occurrence of ``\n`` in the description with ``\n;``.

or

2. Do the above and also determine where to do line breaking (based on the length of the line) and insert ``\n;`` at the line break locations.

henryiii:
I am currently only suggesting 1. 2 isn't important for a file. It's not even very important for the terminal, though it does help a bit there.

I'll try to look at this soon. I'm currently on vacation for Christmas and a bit limited in when I can work on this. 

Henry

SkyToGround:
I am closing this issue seeing as you have accepted the pull request. Thanks for your help on this issue!

PR #66, opened 2017-12-19
Writing argument description to config string (issue #65)

SkyToGround:


include/CLI/App.hpp
```diff
@@ -809,35 +809,46 @@ class App {
...
+                    value = std::to_string(opt->count());
 
                     // Flag, not present
                 } else if(opt->count() == 0 && default_also) {
-                    out << name << "=false" << std::endl;
+                    value = "false";
+                }
+                
+                if (value.size() != 0) {
+                    if (write_description and opt->description_.size() != 0) {
```
henryiii: `and` is not valid here, should be `&&` (Only msvc cares)

SkyToGround: This has now been changed.

henryiii:
When you remove the WIP I'll merge, LGTM.

SkyToGround:
Ah, sorry. Forgot about that as I have been taking time off the past week.

PR #68, opened 2017-12-25
Add check if lexical_cast fully consumed input on integer conversion

SX91:
Would be even better to add checks for integer overflow on `long long` to `T` conversions.


SX91:
Complex numbers parsing is broken now.
With `istringstream` `lexical_cast` is able to parse complex numbers [with streaming operators](http://en.cppreference.com/w/cpp/numeric/complex/operator_ltltgtgt).

henryiii:
Thanks!

Issue #63, opened 2017-12-12
Validator ExistingFileOrDirectory


lczech:
For a program that I am currently writing, it would be nice to have a validator that checks whether a path is existing. Kind of an `or` between `ExistingFile` and `ExistingDirectory`:

```cpp
/// Check for an existing file or directory
inline std::string ExistingFileOrDirectory(const std::string &filename) {
    struct stat buffer;
    bool exist = stat(filename.c_str(), &buffer) == 0;
    if(!exist) {
        return "Path does not exist: " + filename;
    }
    return std::string();
}
```

Could also be called `ExistingPath`, assuming that all kinds of files are allowed, e.g., block devices etc, see https://linux.die.net/man/2/stat

henryiii:
Sure, that could be useful. Could you make a PR?

henryiii:
Now Validators directly support  `&` and `|`, so `ExistingFile | ExistingDirectory` is also possible. But ExistingPath was still useful because it supports other things besides just files and directories. Plus the type is displayed as `PATH` while or-ing `FILE` and `DIR` just gives `TEXT`.

PR #73, opened 2018-01-07
Add path exists validator

lczech:
Fixes issue #63


lczech:
Oh. Just realized that this also contains a small other PR... Sorry for that, hope that's okay.

Issue #76, opened 2018-02-11
"is required" displayed twice in case of required option is not found  


RecklessTodd:
In method App::_parse : 
    RequiredError(opt->single_name() + " is required"); 

In class RequiredError : 
    RequiredError(std::string name) : RequiredError(name + " is required", ExitCodes::RequiredError)

It would make sense to remove the "is required" string from App::_parse method and let it in the RequiredError() class.

henryiii:
Should be fixed, thanks!

henryiii:
Fixed in the just-released 1.4.

PR #77, opened 2018-02-11
#76 | Remove "is required" string passed as an argument to RequiredEr…

RecklessTodd:
…ror() constructor to avoid double printing.

The string "is required" is already added internally by RequiredError() class.


henryiii:
Thanks! Was left over from the old way of adding strings.

PR #78, opened 2018-02-15
Allow MakeSingleHeader.py script to run from non-git source code

dvj:
Currently, if `CLI_SINGLE_FILE` is true, `MakeSingleHeader.py` runs which checks for a tag in the current git repo. This fails if you are not in a repo. This would be the case if, for example, you use the source code from an archive.

This PR just puts a check around the call and adds a nondescript message instead.


henryiii:
Thanks!

PR #79, opened 2018-02-15
Add CMake install command

dvj:
A normal process for CMake based projects is:
```
 cmake ..
 make
 make install
```

where the `make install` step would bundle all the project outputs and copy them to the `CMAKE_INSTALL_PREFIX`.

This PR adds the cmake install commands to do so if `CUR_PROJ` is true


henryiii:
Thanks again!

Issue #87, opened 2018-03-16
Hard to read custom options


AbcAeffchen:
Hi,

I current experiment a little with CLI11 and tried to read a `std::pair<int, float>` as a custom option.
All works fine. I used `set_custom_option("INT FLOAT", 2)`, to enforce that there are exactly two arguments and to reflect the types I expect. But the help text that is generated looks weird.

```
-p INT FLOAT x 2      Some text
```

It gets even worst if I also print default values (depends a little on how you stringify a pair).

```
-p INT FLOAT=(1, 0.00000) x 2      Some text
```

I looked a while into the documentation and what methods are provided, but I couldn't find a solution to this.

So I took a quick look into the code and I added a quick fix for this. See the commit 1561a1a3fb5e5090ec12ccd68c56eebae7767ef5 in my fork. If you are interested I can send a Pull Request. 

But please let me know if I just didn't saw the right options.

Best,
Alex

henryiii:
I like the idea, but do you think the following solution might be better than adding a flag that hides the number print:

* Options would gain a new property "size" that tells CLI11 how "large" one option is. Complex would be 2, most options would be 1, flags would be 0. Size would be passed through `set_custom_option` in instead of expected.
* The number of options expected would then become 1 in your case (1 * 2 = two arguments parsed), so the `x 2` would go away. If you asked for 2, then the x 2 would be printed and 4 options would be expected.

What do you think? Would that be general enough?

AbcAeffchen:
I think this is general enough for my use case. At the moment I think it covers all possibilities for a custom input type. If it gets much more complex one should split it up in multiple options. 
So for me it is fine. 

Will you implement this or should I look at this? I'm not sure how much code needs to be changed for that. You mentioned the complex type.

henryiii:
I don't think it would be that hard - very similar to what you've done already. A new `size_t` would be added, maybe "`option_size_`", instead of the bool you originally added. The loop to collect options would be `option_size_*expected_`. I'll be a while before I can try to implement it, so feel free to have a go at it! I'll help if you have questions or problems.

AbcAeffchen:
`expected_` is also involved in some checking I think. 
If I change `set_custom_option` such that it stores 
```
expected_ = expected < 0 ? -1 : expected * size
```
[...]

PR #92, opened 2018-03-26
Adding size instead of expected + unchangable

henryiii:
Working on #87.

- [x] Move to using two values instead of one for `expected`: `expected` and `type_size`. Drop `unchangable`, which was really just a hack for not knowing the type size.
- [x] Expand policies to support multiple options
- [x] Check/expand support for expected `-N` where `N>1`.
- [ ] Check/fix help printout for `typeval > 1`. Add test. (@AbcAeffchen)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,11 @@ Other, non-user facing changes:
 * Informational printout now added to CTest [#95]
 * Better single file generation [#95]
 * Added support for GTest on MSVC 2017 (but not in C++17 mode, will need next version of GTest)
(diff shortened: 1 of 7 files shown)


henryiii:
@AbcAeffchen, can you check and see if printout works (and patch if it doesn't), and maybe add a test to help test that checks for your expected printout? Any tests you can think of are appreciated, really. :)

Otherwise, I think this is about ready.

henryiii:
Going to go ahead and merge, feel free to add a test/fixes for your use case and PR!

PR #103, opened 2018-04-10
Better mixing of positionals and unlimited options

### LINKED ISSUE #102: Unlimited positional and unlimitied option
Hi, have a question. I have a situation where there's one unlimited positional and one unlimited option. Like this:

```cpp
std::vector<std::string> t1, t2;
app->add_option("test1", t1);
app->add_option("--test2", t2);
```

If you give these arguments:  `./test a b c --test2 d e f`, then `a b c` and `e f` will go to `t1`, and only `d` will go to `t2`.

I've seen in the source code:

https://github.com/CLIUtils/CLI11/blob/b7c031cc3e758b4ea7728c818e0edcbfe7f96c4c/include/CLI/App.hpp#L1465

Could you please explain why you design like this? I think in this situation, `d e f` should go to `t2`.

henryiii:
Since `--` can end an unlimited option, it no longer makes sense to force short-circuiting of unlimited options over positionals. Should give more predictable behavior (closes #102).

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ Extra positional arguments will cause the program to exit, so at least one posit
 If you set `.allow_extras()` on the main `App`, you will not get an error. You can access the missing options using `remaining` (if you have subcommands, `app.remaining(true)` will get all remaining options, subcommands included).
 
 You can access a vector of pointers to the parsed options in the original order using `parse_order()`.
-If `--` is present in the command line,
+If `--` is present in the command line that does not end an unlimited option, then
 everything after that is positional only.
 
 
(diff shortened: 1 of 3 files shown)


harrywong:
Thanks! It's exactly what I want.

henryiii:
This was supposed to have been changed in 1.5, I just forgot about removing the old short circuit behavior.

If it is in 1.5.1 instead of 1.6, copy and pasting the release statement here:

This fixes an old workaround with unlimited positionals vs. unlimited options; Version 1.5 added `--` to end an unlimited option, allowing unlimited options and unlimited positionals to work together naturally without a special short-circuit trigger.

Issue #128, opened 2018-05-19
Support repeating multi-argument options


seanfisk:
First of all, thank you for creating CLI11! It's much better than anything else out there. I've used Boost.Program_Options, which was ok, and QCommandLineParser, which was awful. CLI11 is a breath of fresh air and has saved me a lot of time, so, thank you!

I am looking to implement a command-line interface that allows a repeating option that takes 2 arguments. Here is an example Python program using argparse that implements it:
```python
#!/usr/bin/env python3.6
# repeating_multi_arg_option.py

import argparse

parser = argparse.ArgumentParser(
    description='Repeating options that take multiple arguments')
parser.add_argument('--entry', nargs=2, metavar=('KEY', 'VALUE'),
                    action='append', help='key & value')
print(parser.parse_args())
```
Example runs:
```
$ ./repeating_multi_arg_option.py --entry key1 val1 --entry key2 val2
Namespace(entry=[['key1', 'val1'], ['key2', 'val2']])

$ ./repeating_multi_arg_option.py --entry
usage: repeating_multi_arg_option.py [-h] [--entry KEY VALUE]
repeating_multi_arg_option.py: error: argument --entry: expected 2 arguments

$ ./repeating_multi_arg_option.py --entry key1 --entry key2 val2
usage: repeating_multi_arg_option.py [-h] [--entry KEY VALUE]
repeating_multi_arg_option.py: error: argument --entry: expected 2 arguments

$ ./repeating_multi_arg_option.py --help
usage: repeating_multi_arg_option.py [-h] [--entry KEY VALUE]

Repeating options that take multiple arguments

optional arguments:
  -h, --help         show this help message and exit
  --entry KEY VALUE  key & value
```
It's almost possible with a combination of `…->expected(2)->join()`, but what I'm really looking for is a result that shows accurate error messages and preserves the nested structure (a list of lists).

If there is a path forward, I may also be able to help implement. Thank you!

henryiii:
This was added in 1.5. See #92. You can now do:

```cpp
...->set_custom_option("KEY VALUE", -2);
```

If you really want to go all out and capture the structure too, something like this should work:

```cpp
#include<map>

std::map<std::string, std::string> options;
app.add_option("--entry",
               [&options](CLI::results_t vals){ // results_t is just a vector of strings
    for(size_t i=0; i<vals.size()/2; i++) // will always be a multiple of 2
        options[vals.at(i*2)] = vals.at(i*2+1);
    return true;
    },
[...]

PR #132, opened 2018-05-24
Add virtual destructor to Config class

seanfisk:
Clang++/LLVM 6.0 complains about this from warning `-Wdelete-non-virtual-dtor` which is included when using `-Wall`. This also follows virtual destructor usage in `App` and `FormatterBase`.

By the way, I am still working on #128, I just saw this particular warning when I bumped the compiler version in my own project and figured I'd fix it!


henryiii:
Thanks! Not the first time I've forgotten those...

PR #140, opened 2018-06-26
Add failing test for repeating multi-argument options

seanfisk:
Hi @henryiii, I still wasn't able to get #128 working. But I've finally had some time to work on this, so here is a failing test that showcases what I understand to be the proper success and failure behaviors. The success test passes but the `ValidationError` one does not.

Let me know what you think and thank you for working on this issue!

### DIFF
--- a/include/CLI/Option.hpp
+++ b/include/CLI/Option.hpp
@@ -560,12 +560,18 @@ class Option : public OptionBase<Option> {
             local_result = !callback_(partial_result);
 
         } else {
-            // For now, vector of non size 1 types are not supported but possibility included here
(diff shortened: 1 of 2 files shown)


henryiii:
Added a missing `app.reset()` in your test, and corrected the logic to catch mismatch in typesize and result vector. We'll see if that works!

seanfisk:
@henryiii Thanks much! That is working as intended now — you're the best!

Sorry, I know I had the `app.reset()` at one point; don't know how it got removed…

Do you mind if I add a few more tests for related behavior in the future?

henryiii:
I don't think I'd ever refuse more tests! 👍 

I might be able to drop the need for `app.reset()`, parse should reset the app before running, I believe.

Edit: #141.

PR #145, opened 2018-07-06
Work around Windows min/max macro problem

dvj:
This small change works around the very problematic default feature of `windows.h` declaring macros for `min` and `max`.  See https://stackoverflow.com/questions/5004858/stdmin-gives-error

There are other solutions to this problem (Such as defining `#NOMINMAX` on the project developer's side), but this change keeps the code in CLI11 unambiguous.

This would mean that any further invocations of `std::min` or `std::max` in this project would also need to use the explicit template type version for this change to be useful. I would understand if you do not wish to do that. If you *do* think it is worthwhile, a test could also be created that includes `windows.h` such that the macros would be defined and cause errors when building on Windows if min or max is used ambiguously in the future.

### DIFF
--- a/include/CLI/Option.hpp
+++ b/include/CLI/Option.hpp
@@ -546,7 +546,8 @@ class Option : public OptionBase<Option> {
 
         // Num items expected or length of vector, always at least 1
         // Only valid for a trimming policy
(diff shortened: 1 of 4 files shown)


henryiii:
Thanks!

PR #159, opened 2018-09-03
Warning fix

henryiii:
This fixes several warnings and auto-config errors. Also upgrading to GoogleTest 1.8.1, removing several warning fixes that were accepted into GoogleTest master and included in 1.8.1.


chfast (commented):
I cannot test it directly, but this for sure will fix the issue.

PR #174, opened 2018-10-12
Add the is_assignable to a string condition to return 'TEXT' as type name of a type

OlivierHartmann:
This fixes an overload ambiguity when adding a user new type_name function.

### DIFF
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -76,6 +76,11 @@ constexpr const char *type_name() {
     return "TEXT";
 }
 
+template <typename T, enable_if_t<std::is_enum<T>::value, detail::enabler> = detail::dummy>
+constexpr const char *type_name() {
+    return "ENUM";
+}
+
 // Lexical cast
 
 /// Signed integers / enums
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -32,6 +32,14 @@ template <typename T> struct is_bool { static const bool value = false; };
 /// Check to see if something is bool (true if actually a bool)
 template <> struct is_bool<bool> { static bool const value = true; };
 
+/// Check to see if something is a string (fail check by default)
+template <typename T> struct is_string { static const bool value = false; };
+
+/// Check to see if something is string (true if actually a std::string or const char*)
+template <> struct is_string<std::string> { static bool const value = true; };
+template <> struct is_string<const char*> { static bool const value = true; };
+template <> struct is_string<      char*> { static bool const value = true; };
+
 namespace detail {
 // Based generally on https://rmf.io/cxx11/almost-static-if
 /// Simple empty scoped class
@@ -70,8 +78,8 @@ constexpr const char *type_name() {
 }
 
 template <typename T,
-          enable_if_t<!std::is_floating_point<T>::value && !std::is_integral<T>::value && !is_vector<T>::value && std::is_assignable<T &, std::string>::value,
-                      detail::enabler> = detail::dummy>
+          enable_if_t<!std::is_floating_point<T>::value && !std::is_integral<T>::value && !is_vector<T>::value
+                      && is_string<T>::value, detail::enabler> = detail::dummy>
 constexpr const char *type_name() {
     return "TEXT";
 }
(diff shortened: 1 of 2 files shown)


henryiii:
Let me know if you have a question or need some help. Don't worry about formatting passing, that I can run trivially. I don't have time to look over the patch yet, but can answer questions.

OlivierHartmann:
Hi,
Finally this patch is not needed. I tried to turn it all over the place by using `std::enable_if`, like that:

```cpp
/// Check to see if something is a chrono duration (fail check by default)
template <typename T> struct is_chrono { static const bool value = false; };

/// Check to see if something is a chrono duration (true if actually a chrono duration)
template <class T, class R> struct is_chrono<std::chrono::duration<T,R>> { static bool const value = true; };

template <typename T, typename std::enable_if<is_chrono<T>::value && std::is_same<T,std::chrono::hours>::value, void*>::type = nullptr>
constexpr const char *type_name()
{
	return "TIME [H]";
}
```

But then there is an ambiguous overload with the default `type_name() {"TEXT"}` and the compilo does not know which one to use ! 

But finally, there just need to specialize !!!

```cpp
template <>
constexpr const char *type_name<std::chrono::nanoseconds>()
{
	return "TIME [NSEC]";
}

template <>
constexpr const char *type_name<int>()
{
	return "my nice integer";
}
```

Now it's working perfectly. The thing is I need to do something not really nice like that:

```cpp

namespace CLI
{
	template <typename T, typename R>
	std::istringstream &operator>>(std::istringstream &in, std::chrono::duration<T,R> &val)
	{
		T v;
		in >> v;
		val = std::chrono::duration<T,R>(v);
		return in;
	}

	template <typename T, typename R>
	std::stringstream &operator<<(std::stringstream &in, std::chrono::duration<T,R> &val)
	{
		in << val.count();
		return in;
	}
 }

#include <CLI/CLI.hpp>

namespace CLI
{
	namespace detail
	{
		template <>
		constexpr const char *type_name<std::chrono::hours>()
		{
			return "TIME [H]";
		}

		template <>
		constexpr const char *type_name<std::chrono::minutes>()
		{
			return "TIME [MIN]";
		}
        }
}
```

In order to give to this PR a reason to its life, maybe you can add this to the CLI tutorial, in advanced topics/custom converters section. By the way, the Validators' example structure does not have its constructor with the good name.

Anyway, sorry for the mess.

OlivierHartmann:
Maybe do you think it would be a good thing to have `std::chrono::durations` objects nested in CLI11 ? I can reopen and PR at least that.

henryiii:
I think we could have it in a separate optional file `CLI/Chrono.hpp`. Would that work?

Agree that it should be in the tutorial, as well. ~~Where is the poor name for Validators?~~ Nevermind, I found it. Thanks!

OlivierHartmann:
The thing is the `operator>>` and `operator<<` overloads must be done before CLI.hpp call and `type_name` specializations after. 
I do not see how to do so with only ONE file to call. Any idea ?
Maybe a compilation flag /#define that has to be activated by the user before calling CLI.hpp ?

```cpp
#define CLI_CHRONO_ENABLE
#include "CLI/CLI.hpp"
```

henryiii:
I see that, don't have a good solution yet for a file that could be placed before or after. I think you could at least put the specializations in the before part, as long as you include the declarations in there too - then you would have to include it before.

Adding the fixes to the tutorial: <https://gitlab.com/CLIUtils/CLI11Tutorial/merge_requests/1/diffs> Look okay?

OlivierHartmann:
What do you mean by "include the declarations in there too" ? 
Like the following ?  does not work :( again ambiguous overloads.

```cpp
namespace CLI
{
	template <typename T, typename R>
	std::istringstream &operator>>(std::istringstream &in, std::chrono::duration<T,R> &val) {...}
 
	namespace detail
	{
                template <typename T>
		constexpr const char *type_name(); // prototype declaration

		template <>
		constexpr const char *type_name<std::chrono::hours>() {...}
                
		template <>
		constexpr const char *type_name<std::chrono::minutes>() {...}
              
        }
}

#include <CLI/CLI.hpp>
```

The tutorial seems good, thanks.

Issue #172, opened 2018-10-09
Removing option with a link causes an error


OlivierHartmann:
Hello,

Thank you for this work.
I have found an issue while removing an option that has a link with another one.
Here is a minimalist example causing the bug:

	#include <CLI/CLI.hpp>

	int main(int argc, char **argv) {

	    CLI::App app{"App to demonstrate a bug."};

	    int foo, bar;
	    auto foo_option = app.add_option("--foo,-f", foo, "An argument");
	    app.add_option("--bar", bar, "Another argument")->excludes(foo_option);

	    app.remove_option(foo_option);

	    CLI11_PARSE(app, argc, argv);
	}

Then when running with the "help" flag, we get the following error:

	   terminate called after throwing an instance of 'std::bad_alloc'
	     what():  std::bad_alloc
	   Abandon (core dumped)

The solution can therefore simply be to delete the links at the same time as deleting the option.

Thank you

Olivier

henryiii:
This just needs to be checked as part of `remove_option`, should be pretty easy to add.

PR #179, opened 2018-11-12
Dropping links if option removed

henryiii:
This should address #172.


OlivierHartmann:
Thank you for this patch !

PR #183, opened 2018-12-14
Prefer BUILD_TESTING over CLI11_TESTING if defined

ayum:
CMake variable BUILD_TESTING is used ubiquitously in cmake projects and infrastructure. For example `cget` defines it when building to avoid building tests. If someone greps BUILD_TESTING in vcpkg ports she would get vast number of project using it.

Consider prefer it to CLI11_TESTING if defined. If not defined it will have no effect.


henryiii:
I think you might be a bit more backward compatible than I would have been. :) I'll drop `CLI11_TESTING` in a release or two, then. I might switch to using `include(CTest)` instead, it adds `BUILD_TESTING`.

PR #186, opened 2019-01-05
Add single string parsing

phlptp:
If merged this pull request will add a function to parse a single string containing all the arguments and optionally the program name.  The function splits the single string into a vector and then uses the existing parse function.    Basically a wrapper around the split function to deal with the program name and some edge cases that can arise when everything is in the same string.  

The PR also includes test code for the functionality and
a modification of the split_up function to handle backquote '`' characters and quoted strings

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1212,6 +1212,8 @@ class App {
         }
 
         auto args = detail::split_up(std::move(commandline));
+        // remove all empty strings
+        args.erase(std::remove(args.begin(), args.end(), std::string()), args.end());
         std::reverse(args.begin(), args.end());
 
         parse(args);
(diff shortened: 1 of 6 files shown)


phlptp:
I didn't see any place in the readme where it would be appropriate to add a discussion of the function.  So I didn't do anything regarding that or in the What's new file.  I can add something if you would like.

henryiii:
I have a few minor consistency changes in mind, but I'll go ahead and merge, then make them. I'll add at least a note in the change log.

henryiii:
One suggestion: Most of the contents of the `parse(string` function can be moved to a helper function that is not part of App. This makes it easier to write unit tests for, since you have string in, vector of strings out, and keeps the already long App.hpp from growing too much.

PR #187, opened 2019-01-07
add options to handle windows style command line options

phlptp:
If merged this pull request will add an option for CLI11 to handle windows style options 
such as /d /D /file:some_file.ext  

This is enabled through a new app `function allow_windows_style_options()`

The modifications include some refactoring of the parse_arg function to take a Classifier enum instead of a boolean and the addition of a parse_windows helper function.  

The Classifer enum was renamed to Classifier  (not entirely sure if that was a misspelling or on purpose?)  If it was on purpose I can change it back.

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -191,6 +191,33 @@ struct Range : public Validator {
     template <typename T> explicit Range(T max) : Range(static_cast<T>(0), max) {}
 };
 
+namespace detail {
+/// split a string into a program name and command line arguments
+/// the string is assumed to contain a file name followed by other arguments
+/// the return value contains is a pair with the first argument containing the program name and the second everything
+/// else
+inline std::pair<std::string, std::string> split_program_name(std::string commandline) {
+    // try to determine the programName
+    std::pair<std::string, std::string> vals;
+    trim(commandline);
+    auto esp = commandline.find_first_of(' ', 1);
+    while(!ExistingFile(commandline.substr(0, esp)).empty()) {
+        esp = commandline.find_first_of(' ', esp + 1);
+        if(esp == std::string::npos) {
+            // if we have reached the end and haven't found a valid file just assume the first argument is the
+            // program name
+            esp = commandline.find_first_of(' ', 1);
+            break;
+        }
+    }
+    vals.first = commandline.substr(0, esp);
+    rtrim(vals.first);
+    // strip the program name
+    vals.second = (esp != std::string::npos) ? commandline.substr(esp + 1) : std::string{};
+    ltrim(vals.second);
+    return vals;
+}
+} // namespace detail
 /// @}
 
 } // namespace CLI
--- a/tests/StringParseTest.cpp
+++ b/tests/StringParseTest.cpp
@@ -26,30 +26,6 @@ TEST_F(TApp, ExistingExeCheck) {
     EXPECT_EQ(str3, "\"quoted string\"");
 }
 
-TEST_F(TApp, ExistingExeCheckWithSpace) {
-
-    TempFile tmpexe{"Space File.out"};
-
-    std::string str, str2, str3;
-    app.add_option("-s,--string", str);
-    app.add_option("-t,--tstr", str2);
-    app.add_option("-m,--mstr", str3);
-
-    {
-        std::ofstream out{tmpexe};
-        out << "useless string doesn't matter" << std::endl;
-    }
-
-    app.parse(std::string("./") + std::string(tmpexe) +
-                  " --string=\"this is my quoted string\" -t 'qstring 2' -m=`\"quoted string\"`",
-              true);
-    EXPECT_EQ(str, "this is my quoted string");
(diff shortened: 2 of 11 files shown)


henryiii:
I'll try to review on Thursday. Thanks for catching Classifier!

phlptp:
I am not totally clear what to do about the coverage,  it seems to be no recording the the closing brace on a lambda function, which seems strange to me.

phlptp:
A little background:  I tried a couple different variations the first had a number of different functions to control the parsing.  allow_slash_for_short(),  allow_slash_for_long(),  But that ended up with a lot of conditions and controls in both App and Option.  I also looked at allowing the option definition to allow "/option" in addition to the existing one and have the control be through the options.  But the parsing is done in the App and that got complex in a hurry.   

So I went with a single control in App to simplify matters then nothing would need to be changed in the Options.    It is possible to allow the use /option=XXX but that is not typical for windows style options, it would require a couple lines in the parse_windows function.   So it was left off.  
Also typical windows options do not allow chaining of the short form options,   That restriction fell out naturally from the parsing.  It could be allowed with a little bit of additional logic, but because it is not typical in windows applications seems natural to leave the restriction in place.    I tried to convey those limitations in the readme.  

My main reason for needing this is even though I prefer the unix style option parsing we have a couple cases where our application is layered with windows specific applications that need to forward command line options, and would require mixing the options style and some odd positional parsing if Windows Style options were not allowed.  So allowing that style of options allows everything to be consistent.

tests/StringParseTest.cpp
```diff
@@ -26,30 +26,6 @@ TEST_F(TApp, ExistingExeCheck) {
     EXPECT_EQ(str3, "\"quoted string\"");
 }
 
-TEST_F(TApp, ExistingExeCheckWithSpace) {
```
henryiii, line 29: Why is this being removed?

include/CLI/Validators.hpp
```diff
@@ -191,6 +191,33 @@ struct Range : public Validator {
     template <typename T> explicit Range(T max) : Range(static_cast<T>(0), max) {}
 };
 
+namespace detail {
+/// split a string into a program name and command line arguments
+/// the string is assumed to contain a file name followed by other arguments
+/// the return value contains is a pair with the first argument containing the program name and the second everything
+/// else
+inline std::pair<std::string, std::string> split_program_name(std::string commandline) {
```
henryiii, line 199: Why is this is Validators.hpp? Shouldn't it be in one of the other headers, like string tools?

henryiii (approved):
Overall this looks fantastic. I had a couple of questions. The non-100% coverage issue probably is not the final line of a lambda, but rather a nearby line. For some reason on this and the last PR the coverage line reporting has been off. After it is merged I think we'll see the real line that is causing problems.

henryiii:
Merged, we can look at the coverage after the CI completes on master, and hopefully track down the problem line.

henryiii:
Question: what happens if there's an option named `-a` and `--a`? Who wins?

And, note to self, add to changelog.

include/CLI/Validators.hpp
phlptp: The main reason was it uses the validators to check for an existing file, and if it was included in one of the other tools that header would then have to include the validators header,  which I wasn't sure how you felt about.

tests/StringParseTest.cpp
phlptp: This was mainly testing the file name separation, which is then redundant with the new Validators test since the function that handles that was moved out of the parse function.

phlptp:
The long options are checked first so the long option would take precedence in that case
App.hpp line 1949  `return opt->check_lname(name) || opt->check_sname(name);`

include/CLI/Validators.hpp
henryiii: Okay, we'll leave it there for now. Maybe should be moved to a new header at some point.

tests/StringParseTest.cpp
henryiii: I think having an integration-style test in addition to the unit-style test is not a bad idea, at least as long as it already exists. I might add it back in.

henryiii:
> The long options are checked first 

No, the options are checked one by one in definition order. The first match counts, even if it is `opt->check_sname`. So the first defined option with that name is used. That's fine, it just should be mentioned and tested (doing that now).

henryiii:
@phlptp, would it make sense to set the default for `allow_windows_style_options` to be true on windows, and false otherwise? It would be easier for user code to write `->allow_windows_style_options(false)` if they really did not want windows users to use windows style options. Would there be any drawbacks?

phlptp:
The potential drawbacks would be a slight performance degradation to have an additional check for /options and if someone had a relative path using / to start as part of a positional argument that would probably break it if they didn't turn the check off.    I wouldn't have any problem with that being the default for my applications but it could cause some unexpected behavior in specific situations,  but those would likely be rare.

PR #191, opened 2019-01-12
Clear up Wshadow warnings from gcc 4.9

phlptp:
I ran some integration tests with some of our applications ended up with a couple hundred warnings  from CLI11.  We are not at the point in that application of Zero warnings but that was a bit too much to use.  
The warnings were on gcc 4.9 which seems to be a little pickier, particularly about Wshadow warning.  
See [Travis build](https://travis-ci.org/GMLC-TDC/HELICS-FMI/builds/478608750#0-2201) starting line 1166

This pull request will clear up the warnings from GCC 4.9.  It also adds a few lines of documentation, one or two spelling issues, and adds a few std::move assignments in some member functions where they might be useful.    I didn't thoroughly look for such places but came across a few obvious ones when fixing the Wshadow warnings.

### DIFF
--- a/include/CLI/Formatter.hpp
+++ b/include/CLI/Formatter.hpp
@@ -6,7 +6,7 @@
 #include <string>
 
 #include "CLI/App.hpp"
-#include "CLI/Formatter.hpp"
+//#include "CLI/Formatter.hpp"
 
 namespace CLI {
 
@@ -152,7 +152,7 @@ inline std::string Formatter::make_subcommands(const App *app, AppFormatMode mod
     for(const std::string &group : subcmd_groups_seen) {
         out << "\n" << group << ":\n";
         std::vector<const App *> subcommands_group = app->get_subcommands(
-            [&group](const App *app) { return detail::to_lower(app->get_group()) == detail::to_lower(group); });
+            [&group](const App *sub_app) { return detail::to_lower(sub_app->get_group()) == detail::to_lower(group); });
         for(const App *new_com : subcommands_group) {
             if(mode != AppFormatMode::All) {
                 out << make_subcommand(new_com);
--- a/include/CLI/Formatter.hpp
+++ b/include/CLI/Formatter.hpp
@@ -6,7 +6,7 @@
 #include <string>
 
 #include "CLI/App.hpp"
-//#include "CLI/Formatter.hpp"
+#include "CLI/FormatterFwd.hpp"
 
 namespace CLI {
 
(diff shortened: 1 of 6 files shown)


include/CLI/Formatter.hpp
```diff
@@ -6,7 +6,7 @@
 #include <string>
 
 #include "CLI/App.hpp"
-#include "CLI/Formatter.hpp"
+//#include "CLI/Formatter.hpp"
```
phlptp: The file seemed to be including itself not sure why?  Maybe should have been CLI/FormatterFwd.hpp?

include/CLI/Optional.hpp
```diff
@@ -15,19 +15,25 @@
 #if defined(CLI11_CPP17) && __has_include(<optional>) && \
      !defined(CLI11_STD_OPTIONAL)
 #define CLI11_STD_OPTIONAL 1
+#elif !defined(CLI11_STD_OPTIONAL)
+#define CLI11_STD_OPTIONAL 0
```
phlptp, line 19: I was getting a bunch of warnings about undefined symbols when this is used in #if XXXX statements

include/CLI/Formatter.hpp
henryiii: Yes, it was. Thanks!

include/CLI/Optional.hpp
henryiii: It was supposed to be checked in `ifdef`, but it was checked in an `if` just below, so this is probably best. The main user interaction with these is optionally defining 1 or 0 before including, and if they do define 0, it would have been inconsistent before this patch, so I like it.

henryiii:
I'm surprised there were still warnings, I worked through them at one point, though I do think newer compilers are more forgiving for the case were you set a parameter directly to a member, where it's logically more acceptable to match the names.

Thank you!

phlptp:
I think newer compilers realized that in nearly all cases having a local variable name the same as a function name really couldn't conflict so they don't warn on that any more.    But gcc 4.9 is still in pretty common use yet.  
Thanks.

include/CLI/Formatter.hpp
phlptp: Just a note that the header was commented out in this PR,  you probably want to either remove it or change it to CLI/FormatterFwd.hpp in a subsequent revision.  It probably still works fine since the header is included in other locations before including this one.

phlptp: never mind I see that was done already

PR #201, opened 2019-01-24
Fix for GCC 8 warning

henryiii:
This seems to be a bug in 1.7.0 introduced in #187; the `==` evaluates before the `?`. GCC notices this and gives a warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,9 @@
 [![Conan.io][conan-badge]][conan-link]
 [![Try CLI11 1.6 online][wandbox-badge]][wandbox-link]
 
+[What's new](./CHANGELOG.md) •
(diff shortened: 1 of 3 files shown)


henryiii:
@phlptp, does this look right?

phlptp:
yes the fix looks right.

phlptp:
probably indicates there may be a missing test that would check the false branch of that if statement.  If that statement was always evaluating to true (which would be by far the most common case) then it should be possible to craft a failing test case before the fix.

henryiii:
I’m traveling till Tueday, could you make a test for it?

henryiii:
Thanks!

PR #204, opened 2019-01-25
add_subcommand exception safety

phlptp:
The add subcommand function was not exception safe.  It added the subcommand to the vector before checking the already added option.  This would result in duplicate subcommands being in place in the subcommands_ vector if the error were thrown.  The modifications make it exception safe and remove what I think was an unnecessary check for pointer duplication, that as far as I can tell was always true since it was comparing a newly created pointer directly to previously created ones.


henryiii:
Good point, thanks.

Issue #181, opened 2018-12-07
Callback methods with differing signatures


helmesjo:
I suppose this is what's already mentioned in the readme:

> In C++14, you could have a set of callback methods with differing signatures (tested in a branch). Not deemed worth having a C++14 variation on API and removed.

Which would be:
```
app.add_option("--run", [&](bool doRun){
    handle_run(doRun);
}, "Set to 1 or 0");
```

IMO this would be very powerful, and avoids a lot of boilerplate code. Currently we can kinda do above, but will have to do the conversion in each callback:
```
app.add_option("--run", [&](std::vector<std::string> results){
    bool doRun;
    if (detail::lexical_cast(results[0], doRun))
        handle_run(doRun);
    return true;
}, "Set to 1 or 0");
```

The latter adds _a lot_ of repetition with just two or three options and is very error prone.

Thoughts? It really shouldn't be too much hassle getting this in, spontaneously thinking, but adds great value.

henryiii:
Quick thoughts:

1. That's not what was mentioned on the Readme - that's actually about a different style of adding options that was in version 0.1. I should remove this or make it clearer.
2. What you are asking is quite a bit simpler to add.
3. It would require a C++14 compatible compiler (though technically wouldn't require C++14 mode, though, as most compilers consider this a bugfix and include it in C++11 mode too). Probably should be restricted to C++14 mode just in case.
4. Unless it's a pull request, I'm not likely to have time to add it until I work on the type system a bit (not soon).

henryiii:
What are the signature(s) that you would like to have?

helmesjo:
> What are the signature(s) that you would like to have?

Basically all that are supported using `add_option` (bool, int, float, std::string etc.).

What I have done in my wrapper allows the following:
```
app.add_option<bool>("--run", [&](bool doRun){
    // ...
}, "Set to 1 or 0");
```
_The type can unfortunately not be deduced from a lambda argument like above unless explicitly instantiated as an std::function object (from what I know), so `<bool>` is required._

phlptp:
In some of my early initial explorations with using CLI  having something like add_option("....",CustomType &T, std::function<CustomType(std::string &arg)>  func)  might be useful, particularly in cases where I have a bunch of arguments to convert from strings.    I suspect this can be done with the existing interfaces but requires the resulting location be specified in the lambda.  Splitting it out would allow custom types behave like double or bool or string.    Just a thought

phlptp:
[...]

PR #205, opened 2019-01-25
Add option_function

phlptp:
@henryiii I wanted to let you look at this for comment
@helmesjo 

I have to add the documentation and more tests yet

If merged this pull request would add a add_option_function which would pass the resultant option to a callback function with appropriate signature.  

I think it would address #181 but could use some feedback.  

I needed some things like this since the results of some parse are fed directly into setter functions so I don't have any variable to set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,10 @@ app.add_option(option_name,
                help_string="",
                default=false)
 
+app.add_option_function<type>(option_name,
+               function <void(const type &value)>, // int, float, vector, or string-like
+               help_string="")
+
 app.add_complex(... // Special case: support for complex numbers
 
 app.add_flag(option_name,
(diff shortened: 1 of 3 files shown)


henryiii:
I didn't bring my computer this weekend but I think this looks about right from my iPad. I'll revisit this again next week. I've been drafting a copy of CLI::Set that looks promising and might be able to eventually replace the various `add_*_set_*` functions with just two.

~~I'd probably use `fn` instead of `function` for consistency with `add_flag_fn`. Though I could be talked into a rename.~~ Edit: nevermind, I ended up going with the full spelling. Great, I like that better.

phlptp:
Does this example in the readme work? 
```cpp
app.add_option("--fancy-count", [](std::vector<std::string> val){
    std::cout << "This option was given " << val.size() << " times." << std::endl
    });
```

helmesjo:
> Does this example in the readme work?
> 
> ```c++
> app.add_option("--fancy-count", [](std::vector<std::string> val){
>     std::cout << "This option was given " << val.size() << " times." << std::endl
>     });
> ```

Yep, that works. I use it in my current wrapper around CLI11 to get "type-correct" callbacks, something like this:
```
std::function<void(T)> callback;
auto opt = impl_->cli->add_option(option, [callback](std::vector<std::string> results){
    T value;
    if (CLI::detail::lexical_cast(results[0], value))
    {
        callback(value);
    }
    else
    {
        // Throw...
    }
    return true;
});
```

helmesjo:
Well, basically what @henryiii did in here:
```
        CLI::callback_t fun = [func](CLI::results_t res) {
            T variable;
            bool result = detail::lexical_cast(res[0], variable);
            if(result) {
                return func(variable);
            }
            return result;
        };
```

phlptp:
ok, I see now,  I think the example in the readme needs a return true; statement to make it correct then to match the function signature.

phlptp:
@henryiii  when you get around to it take a look.  I think I got the tests running and coverage complete.  Along with some documentation edits and updates.  I also collapsed the commits to a single one.

henryiii:
I think it looks good. I'll try to push a 1.7.1 release in a day or so before merging just to avoid making a branch (yet) for 1.7.

henryiii:
I've merged your readme fixes (except the new function) to 1.7.1. I'll merge the rest to master, probably for 1.8.0, but maybe 1.7.2.

henryiii:
Thanks again!

PR #206, opened 2019-01-31
add some more tests of custom parsers and adjustments to the README

phlptp:
Add a test that creates and uses a custom parser to store a value.

If merged this PR will add a few more tests using an overload of CLI::defail::lexical_cast and a few statements in the readme.

It includes an alternative for getting a std::complex as an option output.

### DIFF
--- a/tests/NewParseTest.cpp
+++ b/tests/NewParseTest.cpp
@@ -97,3 +97,133 @@ TEST_F(TApp, BuiltinComplexFail) {
 
     EXPECT_THROW(run(), CLI::ArgumentMismatch);
 }
+
+// an example of custom converter that can be used to add new parsing options
+// On MSVC and possibly some other new compilers this can be a free standing function without the template
+// specialization but this is compiler dependent
+namespace CLI {
+namespace detail {
+
+template <>
+bool lexical_cast<std::pair<std::string, std::string>>(std::string input, std::pair<std::string, std::string> &output) {
+
+    auto sep = input.find_first_of(':');
+    if((sep == std::string::npos) && (sep > 0)) {
+        return false;
+    }
+    output = {input.substr(0, sep), input.substr(sep + 1)};
+    return true;
+}
+} // namespace detail
+} // namespace CLI
+
+TEST_F(TApp, custom_string_converter) {
+    std::pair<std::string, std::string> val;
+    app.add_option("-d,--dual_string", val);
+
+    args = {"-d", "string1:string2"};
+
+    run();
+    EXPECT_EQ(val.first, "string1");
+    EXPECT_EQ(val.second, "string2");
+}
+
+TEST_F(TApp, custom_string_converterFail) {
+    std::pair<std::string, std::string> val;
+    app.add_option("-d,--dual_string", val);
+
+    args = {"-d", "string2"};
+
+    EXPECT_THROW(run(), CLI::ConversionError);
+}
+
+// an example of custom complex number converter that can be used to add new parsing options
+#if defined(__has_include)
+#if __has_include(<regex>)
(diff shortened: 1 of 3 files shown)


tests/NewParseTest.cpp
```diff
@@ -97,3 +97,109 @@ TEST_F(TApp, BuiltinComplexFail) {
...
+    std::pair<std::string, std::string> val;
+    app.add_option("-d,--dual_string", val);
+
+    args = {"-d", "string2"};
+
+    EXPECT_THROW(run(), CLI::ConversionError);
+}
+
+#if defined(__has_include)
+#if __has_include(<regex>)
```
henryiii, line 142: Does this work if you are using libstdc++ from GCC 4.8? libstdc++ 4.8 does have `<regex>`, but it's non-functional. Also, if you mix stdlibs, this will fail as well, even on a newer g++ or clang. CLI11 is forgiving if you mismatch stdlibs and compilers. See <https://stackoverflow.com/a/41186162/2402816>. It might be better to optionally enable this test from CMake, etc.

phlptp: For GCC the check worked since gcc 4.8 doesn't have __has_include.  I think the clang that used an older standard library was failing due to that issue.  I changed the check on the complex test to use the CLI11_CPP14 check.  That isn't exactly accurate since it does work on C++11 but the check for the other conditions is probably too complex to be worthwhile

i tried on ubuntu 14.04 with Clang 3.5 and gcc 4.8.  I ran into a few issues with cmake 3.5 though in the build process.    Did you want those as a separate PR or just include in this one?

phlptp:
Is the travis clang test on clang-5 linking against an old libstd++ even when set to C++14?

henryiii:
I hope so, there is supposed to be one (at least for C++11).

henryiii:
That looks like a good solution.

phlptp:
Interesting,  I had never attempted to link a non c++14 compliant standard library with a compiler using C++14 mode before.  I think I had always assumed that would fail in linking stage.

phlptp:
I am not entirely clear why it is failing on the clang build yet.

tests/NewParseTest.cpp
henryiii: Separate PR would be nice, incase this takes a while to get right. But same PR is also acceptable.

phlptp:
I will merge a lot of these commits together once I figure out what is going on, so it looks a little cleaner in the end.

henryiii:
No problem. Unless there's a reason for more than one commit (such as a large addition), I can also squash and merge.

henryiii:
Feel free to add a `CLI11_HAS_REGEX` or  `CLI11_TESTS_HAS_REGEX` option to CMake, and we can manually turn it on for the test jobs that it will work on. Also, if you use the current method, please go back to has_include + the runtime test, rather than checking for C++14. (After you isolate the issue, anyway)

Also, FYI, the first few development versions (up to 0.3 or so?) of CLI11 had parsing based on regex, but the requirement was dropped due to the issues with older compilers and even older stdlibs.

henryiii:
That looks pretty good. Is it ready to merge?

phlptp:
I believe so, As far as I can tell the issue had something to do with scoping and the const regex string was not available when actually used.  But I am not sure what travis was doing since it worked fine on all the clang versions I tested with.  Anyway putting the const in the function seemed to have resolved it.    And I merged all the commits together so If you are comfortable with the modifications I think it is good to go.

henryiii:
I need to see if https://cliutils.gitlab.io/CLI11Tutorial/chapters/advanced-topics.html should be updated, too. Will look into it later.

PR #208, opened 2019-01-31
Fixes some cmake issues when using earlier cmake versions 

phlptp:
specifically Boost::boost is not defined before boost 3.5.2 so 3.4 doesn't work.  Also the AND condition for MSVC didn't work on older cmakes for addGoogleTest file.


henryiii:
Thanks!

PR #209, opened 2019-01-31
Support user defined delimiter when parsing list

rafiw:
Currently user can only parse list with spaces
E.g app -o 1 2 3
now use can define his own one char delimiter e.g comma

### DIFF
--- a/tests/AppTest.cpp
+++ b/tests/AppTest.cpp
@@ -1961,3 +1961,94 @@ TEST_F(TApp, BeforeRequirements) {
     // args = {"-b", "-a", "extra"};
     // EXPECT_THROW(run(), CLI::CallForHelp);
 }
+
+// #209
+TEST_F(TApp, CustomUserSepParse) {
+
+    std::vector<int> vals = {1, 2, 3};
+    args = {"--idx", "1,2,3"};
+    auto opt = app.add_option("--idx", vals, "", ',');
+    run();
+    EXPECT_EQ(vals, std::vector<int>({1, 2, 3}));
+
+    app.remove_option(opt);
+
+    app.add_option("--idx", vals, "", true, ',');
+    run();
+    EXPECT_EQ(vals, std::vector<int>({1, 2, 3}));
+}
+
+// #209
+TEST_F(TApp, DefaultUserSepParse) {
+
+    std::vector<int> vals = {1, 2, 3};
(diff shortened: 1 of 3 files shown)


henryiii:
Please add a few tests, including at least the following cases:

```
-v 1 2
-v 1,2
-v 1,2,
-v 1, 2
-v "1,2"
-v "1, 2 "
```

I believe space will always be a separator, since this is what gets passed to CLI11 from argc, argv.

henryiii:
I can fix the style pretty trivially if needed. Is there a version with defaults to add too?

tests/AppTest.cpp
```diff
@@ -339,7 +339,7 @@ TEST_F(TApp, doubleVectorFunction) {
     });
     args = {"--val", "5", "--val", "6", "--val", "7"};
     run();
-    EXPECT_EQ(res.size(), 3);
+    EXPECT_EQ(res.size(), (size_t)3);
```
harrywong: I think `EXPECT_EQ(res.size(), 3u);` is better here :)

henryiii: That might be a good idea generally, but for now, this is consistent.

henryiii: Okay, this can be changed now. I've converted to using `#u` everywhere in testing. Thanks!

henryiii:
This should be added in the 

```cpp
Option *add_option(std::string option_name,
                       std::vector<T> &variable, ///< The variable vector to set
                       std::string description,
                       bool defaulted)
```
 one too, right below, and matching tests copied and pasted in.

rafiw:
i will update next week

rafiw:
@henryiii there are some style issues not related to my changes, besides that i think it's ok

henryiii:
Okay, I'll rerun the formatter soon.

henryiii:
Looks like the 5.0.1 version of clang-format is a bit finicky too. I'll play with cleanup tomorrow, and will probably add a couple of clang-format-off directives (or switch to a docker clang-format on Travis and use unibeatify's version).

henryiii:
By the way, I believe the API will change a bit with #240: Instead of an extra argument in add_option, you would do it like this now:

```cpp
add_option("--thing", vector)->delimiter(',');
```

As a bonus, the delimiter is now supported by validators and more.

rafiw:
Great work

PR #210, opened 2019-01-31
add validator for ip address and positive number

rafiw:


henryiii:
This is great, thanks!

henryiii:
I’ll fix the two comments tomorrow, and the tname for positive number should be all caps.

Issue #215, opened 2019-02-03
Allow subcommands Apps to be extracted and merged


phlptp:
A bit of background.  
I am working on moving to CLI11 from boost::program_options.  The main reason being to move away from a dependency on a compiled boost library which was giving us problems in terms of portability.  

So CLI11 is what we chose to move to.  This required adding some additional capabilities to CLI11 to match the required functionality, We think CLI11 will give us the needed capabilities and quite a few that will ultimately make our code simpler and more capable.  

In some of our programs, the command line processing happens in several layers.  An outer layer doing some things that build other objects pulling in objects from several different components that are mostly independent of each other,  With boost::program_options this mainly meant just passing the command line arguments between all the different options or just passing the resulting parse map to the different options to deal with processing.    The nature of CLI11 make this paradigm a little more complicated in some aspects but also opportunities for cleaning things up and simplifying.   One needed feature was a way to merge parse options together from various different sources.    One option is to pass a main App to the different component to add their own options into,  but this means the objects are dependent on some higher level which might not always be the case.    So what I want to do is have the component objects build a CLI::App then have a higher level object merge them if it wants to, or the modules can parse options or option strings directly.  

I will submit a PR with one option for this in a couple days but I wanted to start a discussion on this and see if there was other things that might be impacted or considered.   The idea is basically to replace the unique_ptr storage of subcommands with shared_ptr.   Add a set of get_subcommand_ptr functions just like get_subcommand to actually retrieve the shared_ptrs, and an additional function to add a subcommand from a shared_ptr.    Closely related is an ability to add nameless subcommands, that behave similarly to groups in some ways basically commands that don't match in the main App fall down to all nameless subcommands.  I think this will mean no changes to existing functions, just internal changes and some additional overloads and functionality.   

Basically talking about adding two things that will make an ability to modularize the options and App for parsing options.  
 - ability to extract add, and reuse subcommands
 - ability to add subcommands that don't have a name so don't trigger on a subcommand argument but can be triggered from any option that doesn't match in the parent.

henryiii:
Thank you for the details. Overall, this sounds like an excellent addition.

[...]

PR #216, opened 2019-02-04
Add unnamed subcommand

phlptp:
If merged this pull request will allow unnamed subcommands that get checked for options if no matches occur in the main App.  it also allows the extraction and insertion of subcommands via shared_ptr.  

See issue #215.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -49,7 +49,7 @@ std::string help(const App *app, const Error &e);
 
 class App;
 
-using App_p = std::unique_ptr<App>;
+using App_p = std::shared_ptr<App>;
 
 /// Creates a command line program, with very few defaults.
 /** To use, create a new `Program()` instance with `argc`, `argv`, and a help description. The templated
@@ -77,8 +77,11 @@ class App {
     /// If true, allow extra arguments in the ini file (ie, don't throw an error). INHERITABLE
     bool allow_config_extras_{false};
 
-    ///  If true, return immediately on an unrecognised option (implies allow_extras) INHERITABLE
+    ///  If true, return immediately on an unrecognized option (implies allow_extras) INHERITABLE
     bool prefix_command_{false};
+
+    /// if set to true the name was automatically generated from the command line vs a user set name
+    bool has_automatic_name_{false};
 
     /// This is a function that runs when complete. Great for subcommands. Can throw.
     std::function<void()> callback_;
@@ -244,6 +247,7 @@ class App {
     /// Set a name for the app (empty will use parser to set the name)
     App *name(std::string app_name = "") {
         name_ = app_name;
+        has_automatic_name_ = false;
         return this;
     }
 
@@ -1124,17 +1128,29 @@ class App {
     ///@{
 
     /// Add a subcommand. Inherits INHERITABLE and OptionDefaults, and help flag
-    App *add_subcommand(std::string subcommand_name, std::string description = "") {
-        CLI::App_p subcom(new App(description, subcommand_name, this));
-        for(const auto &subc : subcommands_)
-            if(subc->check_name(subcommand_name) || subcom->check_name(subc->name_))
-                throw OptionAlreadyAdded(subc->name_);
+    App *add_subcommand(std::string subcommand_name = "", std::string description = "") {
+        CLI::App_p subcom = std::shared_ptr<App>(new App(description, subcommand_name, this));
+        return add_subcommand(std::move(subcom));
+    }
+
+    /// Add a previously created app as a subcommand
+    App *add_subcommand(CLI::App_p subcom) {
+        if(!subcom)
+            throw IncorrectConstruction("passed App is not valid");
+        if(!subcom->name_.empty()) {
+            for(const auto &subc : subcommands_)
+                if(subc->check_name(subcom->name_) || subcom->check_name(subc->name_))
+                    throw OptionAlreadyAdded(subc->name_);
+        }
+        subcom->parent_ = this;
         subcommands_.push_back(std::move(subcom));
         return subcommands_.back().get();
     }
-
     /// Check to see if a subcommand is part of this command (doesn't have to be in command line)
+    /// returns the first subcommand if passed a nullptr
     App *get_subcommand(App *subcom) const {
+        if(subcom == nullptr)
+            throw OptionNotFound("nullptr passed");
         for(const App_p &subcomptr : subcommands_)
             if(subcomptr.get() == subcom)
                 return subcom;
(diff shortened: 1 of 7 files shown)


phlptp:
A couple remaining issues
Should there be a way to remove and or disable subcommands?  
Need to update the Readme yet.

include/CLI/App.hpp
```diff
@@ -1124,17 +1131,31 @@ class App {
...
+        if(subcom) {
+            if(!subcom->name_.empty()) {
+                for(const auto &subc : subcommands_)
+                    if(subc->check_name(subcom->name_) || subcom->check_name(subc->name_))
+                        throw OptionAlreadyAdded(subc->name_);
+            }
+            subcom->parent_ = this;
+            subcommands_.push_back(std::move(subcom));
+        }
+        return this;
```
phlptp: I am not entirely certain what this function should return.  Right now it returns a pointer to the main app so it can be chained,  but to match the other add_subcommand functions it should return a pointer to the added subcommand.  

Not sure which is preferred here.

include/CLI/App.hpp
```diff
@@ -77,9 +77,15 @@ class App {
     /// If true, allow extra arguments in the ini file (ie, don't throw an error). INHERITABLE
     bool allow_config_extras_{false};
 
-    ///  If true, return immediately on an unrecognised option (implies allow_extras) INHERITABLE
+    ///  If true, return immediately on an unrecognized option (implies allow_extras) INHERITABLE
     bool prefix_command_{false};
 
+    /// set to true if an unnamed subcommand were added
+    bool has_unnamed_subcommands_{false};
```
henryiii: Can't this be queried? Not fond of adding too many values, and if we allow deleting/removing subcommands, this would be hard to track. Looping over a handful of subcommands is not too expensive.

include/CLI/App.hpp
henryiii: It either needs to match the first one or have a different name. I don't think chaining is important here.

include/CLI/App.hpp
```diff
@@ -1231,8 +1283,10 @@ class App {
     /// This must be called after the options are in but before the rest of the program.
     void parse(int argc, const char *const *argv) {
         // If the name is not set, read from command line
-        if(name_.empty())
+        if((name_.empty()) || (has_automatic_name_)) {
+            has_automatic_name_ = true;
```
henryiii, line 1282: This only gets called on the main command, it can't be a subcommand. Would it be possible to avoid the `has_automatic_name_`?

henryiii (commented):
Another idea would be to have two lists of subcommands, named and unnamed, and then looping over one or the other as needed.

include/CLI/App.hpp
phlptp: ok, that sounds good.  Should it throw if the pointer if not valid or return a nullptr?

include/CLI/App.hpp
phlptp: It is internal so wouldn't be queried, but if we did a couple more loops over the subcommands it could be removed

include/CLI/App.hpp
phlptp: The issue that is solving is if one of the nameless subcommands were called in an independent context with argc, argv.    In that case it would prohibit it from being used as a nameless subcommand in later contexts unless the automatically generated name can be knowingly and automatically cleared.

henryiii: Okay; I like to limit global state if possible, but I think this one's okay.

include/CLI/App.hpp
henryiii: I think what you have is fine.

phlptp:
Todo yet
- [x] combine all the commits to a single commit
- [x] update the readme

phlptp:
Question.  Should the `parsed_` flag on the nameless subcommands be set.  Right now they are not but it wouldn't be hard to set the nameless subcommands when the `parsed_` flag is set in the main `App`?

phlptp:
Right now the nameless subcommands get printed with help before a list of subcommands.  If a subcommand has a non-default group definition should it do something with that or should the group information on a nameless subcommand be ignored?

phlptp:
@henryiii If the latest commit passes Codecov I will merge everything to a single commit, unless you see some other things that should be changed.

include/CLI/App.hpp
```diff
@@ -1124,17 +1128,29 @@ class App {
...
-                throw OptionAlreadyAdded(subc->name_);
+    App *add_subcommand(std::string subcommand_name = "", std::string description = "") {
+        CLI::App_p subcom = std::shared_ptr<App>(new App(description, subcommand_name, this));
+        return add_subcommand(std::move(subcom));
+    }
+
+    /// Add a previously created app as a subcommand
+    App *add_subcommand(CLI::App_p subcom) {
+        if(!subcom)
+            throw(OptionAlreadyAdded("passed App is not valid"));
```
```suggestion
            throw OptionAlreadyAdded("passed App is not valid");
```

henryiii: Is OptionAlreadyAdded really the best error here? I would at least make the suggested change.

include/CLI/App.hpp
```diff
@@ -1148,9 +1164,41 @@ class App {
                 return subcomptr.get();
         throw OptionNotFound(subcom);
     }
+    /// Get a pointer to subcommand by index
+    App *get_subcommand(int index = 0) const {
```
henryiii, line 1168:
```suggestion
    App *get_subcommand(size_t index = 0) const {
```

This would remove the need for checking <0.

include/CLI/App.hpp
```diff
@@ -1148,9 +1164,41 @@ class App {
...
+    /// Check to see if a subcommand is part of this command (text version)
+    CLI::App_p get_subcommand_ptr(std::string subcom) const {
+        for(const App_p &subcomptr : subcommands_)
+            if(subcomptr->check_name(subcom))
+                return subcomptr;
+        throw OptionNotFound(subcom);
+    }
+
+    /// Get an owning pointer to subcommand by index
+    CLI::App_p get_subcommand_ptr(int index = 0) const {
```
henryiii, line 1193: Same. I *think* the other similar gets use unsigned indexes as well, if not then ignore this suggestion.

henryiii:
Add a few small last minute suggestions you can look at. I'm not sure what an unnamed subcommand should do for "parsed" - it rather seems like setting it to 1 when it is checked (that is, all unnamed subcommands will get a 1) might be an option. Really, a user shouldn't be checking the "count" of an unnamed subcommand, so the only effect would be in the parsing, where it seems like it wasn't needed for tracking unnamed subcommands.

henryiii:
I think we can tweak the formatting if needed on nameless subcommands after the merge, do what you think is best for now.

include/CLI/App.hpp
phlptp: I was poking through the list of errors available, and short of creating a new one I didn't see one that fit very well so I just went with the one that was already being thrown by that other overloads.  Happy to change it if I missed a better one.

phlptp: I am changing it to an IncorrectConstruction Error.    That might make a little more sense

include/CLI/App.hpp
phlptp: I was trying to remember why I put it as an int.  Then I remembered that there is an overload that takes a pointer.  and if you pass a literal 0 you get a ambiguity, since 0 is also a valid pointer (gotta love C++) and if you gave it a NULL it would also be ambiguous. And that would just be annoying to use.  With int it matches the literal precisely and there is no ambiguity.  so you can call get_subcommand(0) and not get a compile error.

henryiii: Great, thanks C++. Okay.

henryiii:
Thanks!

Issue #217, opened 2019-02-05
add support for click-style boolean flags


pwm1234:

pwm1234:
First let me say thanks for your work on CLI11 and making it available. CLI11 is a big help to me.

I am also a user of click for my python programs, and I like how CLI11 supports many of the features of click. One click feature that I would like to see in CLI11 is how to set boolean flags with the "--option/--no-option" syntax (see [Boolean Flags](https://click.palletsprojects.com/en/7.x/options/#boolean-flags)). Is this capability somewhere on the roadmap? If not, is it something you would consider adding? Any suggestions on how to go about implementing this in a pull-request that you would consider for merging?

henryiii:
I guess this has only been talked about in emails and maybe gitter. Yes, it's an oft-requested feature, and I'd be glad to get it in CLI11. I'd like to have flexibility in the "no-" prefix, and stay with the current Option-only design (flags are just options internally). First thing is to cover the API. How would you expect this to be activated when a user adds a flag? Is it settable on the default option? Does it make any sense to do any different for regular options too (I think not)? @phlptp might have ideas and/or suggestions.

henryiii:
PS: I don't think I've looked at the click syntax yet, that might be interesting.

pwm1234:
Click uses an option with a boolean value for what CLI11 calls a flag. The (primary) click api is ``@click.option('--shout/--no-shout', default=False)``. Click also has an ``is_flag`` argument if you want to leave out the ``--no-`` switch. I am not quite sure a good way to map this to the CLI11 flag api. 

The logic seems to be:

1. not require (or not accept) values for bool options (aka flags). 

2. to recognize both the ``--option`` and the negated ``--no-option`` for the same variable, where ``--option`` sets the variable to true and ``--no-option`` sets the variable to false. 

For CLI11 this seems to be neither an option nor a flag. (It is an option without a specified value; it is a flag that needs to set a bool variable.) Perhaps the clearest CLI11 approach would be to add a new method ``CLI::App::add_flag_negatable(std::string flag_name, bool& flag, std::string description="")``.

phlptp:
[...]

PR #219, opened 2019-02-07
click-style boolean flags

phlptp:
Add functionality similar to click style argument that allow specifying a false flag that when false generate a false result on the flag.

See Issue #217 

after the discussion in #217 I started looking at our use cases and realized that I could probably eliminate 30% of the options if this was supported in some fashion.    And since I am on paternity leave and can tinker mostly with what I find interesting, this seemed like an interesting challenge.  

This is by no means complete,  needs quite a few test cases and documentation yet but enough is working to get started discussing it.  

After some tinkering what I ended up with is this. 

The `add_flag` and `add_flag_function` calls can take a syntax   like "--flag|--no_flag, -f|-n"  arguments after a | are considered false_flags.  If they are passed they trigger a false response.  
I went with the '|' instead of '\' like clicked since I didn't want to confuse with windows style options.  I think it would be preferable to generate an error like it does now on that since CLI11 takes only `--` and '-' arguments. and clicked seems to have a bunch of special conditions that deal with that and that seemed like it could cause trouble.  '|' seemed like a nice alternative.  

It is also possible to specify the false_flags through a separate function `false_flags`    This function checks against existing arguments and fills out the fnames_ vector which is automatically filled from `|` notation names.   

What happens is that the flags are checked for presence of  '|' if so they extract the false flags separately and replace the '|' with a comma.  All the arguments are moved into the appropriate lname or sname, so that checking doesn't change.  

Then if the argument takes 0 arguments the actual flag passed is checked against an fnames array.  If it is present the flag is flipped.    

The output vector of strings either contains a "+" or "-"    The counts are essentially number of "+" minus number of "-".   So the signature of add_flag_function was changed so the std::function returned an int as the possibility exists it could be negative.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -538,7 +538,6 @@ class App {
     /// Add option for flag
     Option *add_flag(std::string flag_name, std::string description = "") {
         CLI::callback_t fun = [](CLI::results_t) { return true; };
-
         Option *opt = add_option(flag_name, fun, description, false);
         if(opt->get_positional())
             throw IncorrectConstruction::PositionalFlag(flag_name);
@@ -552,14 +551,21 @@ class App {
     Option *add_flag(std::string flag_name,
                      T &flag_count, ///< A variable holding the count
                      std::string description = "") {
-
         flag_count = 0;
+        Option *opt;
         CLI::callback_t fun = [&flag_count](CLI::results_t res) {
-            flag_count = static_cast<T>(res.size());
+            detail::sum_flag_vector(res, flag_count);
             return true;
         };
+        if(detail::has_false_flags(flag_name)) {
+            std::vector<std::string> neg = detail::get_false_flags(flag_name);
+            detail::remove_false_flag_notation(flag_name);
+            opt = add_option(flag_name, fun, description, false);
+            opt->fnames_ = std::move(neg);
+        } else {
+            opt = add_option(flag_name, fun, description, false);
+        }
 
-        Option *opt = add_option(flag_name, fun, description, false);
         if(opt->get_positional())
             throw IncorrectConstruction::PositionalFlag(flag_name);
         opt->type_size(0);
@@ -570,16 +576,23 @@ class App {
     /// `multi_option_policy(CLI::MultiOptionPolicy::Throw)` is used.
     template <typename T, enable_if_t<is_bool<T>::value, detail::enabler> = detail::dummy>
     Option *add_flag(std::string flag_name,
-                     T &flag_count, ///< A variable holding true if passed
+                     T &flag_result, ///< A variable holding true if passed
                      std::string description = "") {
-
-        flag_count = false;
-        CLI::callback_t fun = [&flag_count](CLI::results_t res) {
-            flag_count = true;
+        flag_result = false;
+        Option *opt;
+        CLI::callback_t fun = [&flag_result](CLI::results_t res) {
+            flag_result = (res[0][0] != '-');
             return res.size() == 1;
         };
+        if(detail::has_false_flags(flag_name)) {
+            std::vector<std::string> neg = detail::get_false_flags(flag_name);
+            detail::remove_false_flag_notation(flag_name);
+            opt = add_option(flag_name, fun, std::move(description), false);
+            opt->fnames_ = std::move(neg);
+        } else {
+            opt = add_option(flag_name, fun, std::move(description), false);
+        }
 
-        Option *opt = add_option(flag_name, fun, description, false);
         if(opt->get_positional())
             throw IncorrectConstruction::PositionalFlag(flag_name);
         opt->type_size(0);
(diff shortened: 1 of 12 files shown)


phlptp:
## Things to do yet
- [x] more tests for coverage
- [x] deal with unsigned types in the flag function as it could wrap yet
- [x] documentation
- [x] figure out how print help for false_flags in some way that makes sense

henryiii:
I don't have time to review this yet, but a quick thought; The | seems to be really hard to tell from I or l, so it does not read well. I don't like click's "mixed" form either, where either / or ; are used depending on the option style. / does really look like an "anti" option, though.

By the way, congratulations!

phlptp:
Some other character options I have seen used are '!' and '~'  and '^'  
regarding '/'  say someone did something like `add_flag("/flag",val)`    on first glance my instinct would say that would be adding a windows style flag option.    Currently this should generate an exceptions since it is an invalid name for an option.  but if we use that for false_flags it would generate a flag that produces false when passed,  which could be really frustrating to figure out what is going on, so I am really hesitant to use '/' in that context.  Click probably gets away with it since they take windows options as valid to begin with.  

other options 
 - `add_flag("--val1|--no-val",val)` , `add_flag("|--no-val",val)`
 - `add_flag("--val1~--no-val",val)` , `add_flag("~--no-val",val)`
 - `add_flag("--val1!--no-val",val)` , `add_flag("!--no-val",val)`
 - `add_flag("--val1^--no-val",val)` , `add_flag("^--no-val",val)`

I suppose it might make sense to do away with the pair of options syntax and just make it such that options that start with `!` trigger to false
so 
 - `add_flag("--val1,!--no-val",val)` , `add_flag("!--no-val",val)`
 - `add_flag("--val1,^--no-val",val)` , `add_flag("^--no-val",val)`

I am not sure '~ would read nicely either

include/CLI/App.hpp
```diff
@@ -527,8 +527,7 @@ class App {
 
     /// Add option for flag
     Option *add_flag(std::string flag_name, std::string description = "") {
-        CLI::callback_t fun = [](CLI::results_t) { return true; };
-
+        CLI::callback_t fun = [](CLI::results_t res) { return true; };
```
henryiii: Why is there a name "res" here now?

henryiii:
I believe I like:

* `add_flag("--val1,!--no-val",val)` , `add_flag("!--no-val",val)`

Best. It's simpler to parse, allows any order `"!--no-val,--no-val"`, and keeps the comma as the option separator.

include/CLI/Option.hpp
```diff
@@ -366,6 +370,32 @@ class Option : public OptionBase<Option> {
             return false;
         }
     }
+    /// set the false flags.  flags in this list trigger a negative response for a flag (false instead of true)
+    void false_names(std::string fnames) {
```
henryiii: I'm not sure if this is really needed as a public function. You currently can't change the names of an option after you create it, either.

henryiii:
So far, really liking where you are going with this.

include/CLI/App.hpp
phlptp: Probably because I started tweaking that function to use then I realized it didn't make much sense for the `add_flag` that doesn't return something or have a user callback.  I will remove it.

include/CLI/Option.hpp
phlptp: My thinking was you could change which flags generated false and which true after the fact,  But it probably is unnecessary and could be added later if there became a stronger need for it.

phlptp:
Playing around with the help syntax.  I wondered about the possibility of this syntax

* `add_flag("--val1,--no-val{false}",val)` , `add_flag("--no-val{false}",val)`  

I think this makes it much more explicit and the help can print a line that can be used in another flag definition.    
I implemented this and {true} also works though that is default so it is ignored but works for symmetry.  

I like this better,  @henryiii  what do you think?

phlptp:
I also did some tinkering with the flag values after noting some inconsistencies.  For example I think passing a flag such that --flag=false  would then trigger that flag to return true,  which would be kind of unexpected behavior.

henryiii:
I would at least like a shortcut notation, `!--val`, since this can be very verbose (and you already have to list the core part of the flag name twice for the most common way of defining the flags). You also have more ways to make a mistake, like `{True}` or `{ON}` I'm fine with the `{false}` and `{true}` also being supported for consistency with the help printout. You might be able to convince me to give up the short cut notation if you have a strong preference.

For `--flag=True` (or false), that's something I think would be nice to support that we don't support currently. Especially because we already have support for the INI reader for similar notation; you could even reuse the true/false matching function it uses to recognize "true"-like and "false"-like values.

phlptp:
I enabled the --flag=XX notation, for command line options otherwise it was somewhat inconsistent and excess code duplication.  I tweaked it a little there was a possibility if a number is passed as the argument it generated that many elements in a vector.  Which if you put in a big number there could easily run out of memory,  so I capped that number at 9 if the number is bigger it just makes a single element(true or false).  I thought about just capping it at 10 but that seemed a bit arbitrary as well, so I don't know if you have a preference, but I think there needs to be some sort of cap and check on that to prevent out of memory errors.

henryiii:
How about putting the number in the vector, and summing them?

```cpp
{"+", "+", "-"}
```

becomes

```cpp
{"1", "1", "-1"}
```

True means 1, False means -1, special case for 0 going to -1 I think so `--flag=0` is false (-1).

?

include/CLI/TypeTools.hpp
```diff
@@ -150,5 +170,27 @@ bool lexical_cast(std::string input, T &output) {
...
+    });
+    output = (count > 0) ? static_cast<T>(count) : T{0};
+}
+
+/// sum a vector of flag representations
+template <typename T,
+          enable_if_t<std::is_integral<T>::value && std::is_signed<T>::value, detail::enabler> = detail::dummy>
+void sum_flag_vector(const std::vector<std::string> &flags, T &output) {
+    int64_t count = std::accumulate(flags.begin(), flags.end(), int64_t{0}, [](int64_t cur, const std::string &flag) {
+        cur += (flag == "1") ? 1 : ((flag == "-1") ? (-1) : std::stoll(flag));
```
henryiii: Why not this:

```suggestion
        cur += std::stoll(flag);
```

Is that for speed, perhaps? Note: This is a question, not a required change.

henryiii:
I think you might be missing an `#include <numeric>` for accumulate.

include/CLI/TypeTools.hpp
phlptp: mostly,  there will be two very common strings with "1" being most common,  didn't see a reason to not take advantage of that fact.

henryiii: I would add a little comment saying that soma reader (like me) won't have to work though the logic just to find it is doing stol.

henryiii:
Remove WIP when ready for a final review and merge.

phlptp:
@henryiii I think it is ready for final review

henryiii:
Other than the touchup I just applied, I think this is ready to go in! It looks fantastic. I'll squash and merge tomorrow after CI (I only changed the README, but this way I get a green checkmark on the PR).

henryiii:
Thanks again!

PR #221, opened 2019-02-08
Change default delimiter to '\0' instead of space

### LINKED ISSUE #218: Quoted arguments with spaces are split up
Given the following program:

```c++
#include <CLI/CLI.hpp>

    CLI::App app("foo");

    std::vector<std::string> bar;
    app.add_option("bar", bar, "bar")->required();

    CLI11_PARSE(app, argc, argv);

    for (const auto& entry : bar) {
        std::cout << entry << std::endl;
    }

    return 0;
}
```

I would expect the output to be like this:

```
$ ./program this is a test
this
is
a
test
$ ./program "this is a" test
this is a
test
```

However, the second command instead prints the words separately just like the first command. This seems like a bug caused by #209.

rafiw:
This maintains CLI11 previous way of work.
This  fixes #218 

Signed-off-by: Rafi Wiener <rafiw@mellanox.com>

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -418,17 +418,22 @@ class App {
     Option *add_option(std::string option_name,
                        std::vector<T> &variable, ///< The variable vector to set
                        std::string description = "",
-                       char delimiter = ' ') {
+                       char delimiter = '\0') {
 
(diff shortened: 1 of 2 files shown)


henryiii:
Thanks!

Issue #124, opened 2018-05-09
Hybrid flag/option


sasq64:
I would like to have support for options that can act as flags if no argument is given.
These options would require the forms `-Xsomething` or `--longer=something` when used as options.

```c++
opts.add_hybrid_option("--sign", stringValue, "Sign archive with specified key or default if none given", "defaultKey");

if(stringValue != "") {
  // Perfom signing
}
```
Should print help text such as;
` --sign[=TEXT]`

henryiii:
Just thinking:

This is almost possible currently if you set the min/max allowed options to 0/1, respectively. The requirement of the special forms could be added as an "allowed forms" type option, with flags. Something like: `->allowed_short_form(CLI::NoSpace)->allowed_long_form(CLI::Equals)`.

henryiii:
This is now possible!

Issue #123, opened 2018-05-09
Flag and/or option fallback function


sasq64:
I want to parse special flags that begins with a dash and contains a single digit to mean compression levels for my zip like compression tool.

Right now I need to handle this with a special check inside a callback for positional arguments.

I would like to have something like:

```c++
opts.add_flag_fallback("-<digit>", [&](const char d) {
    compressionLevel = d - '0';
}, "Set compression level (0-9)");
```

henryiii:
I'm thinking this could be a special type of add:

```cpp
app.add_digit(Optional int value, "description");
```

henryiii:
This is now possible!

PR #222, opened 2019-02-08
CLI::IsMember

henryiii:
This implements `CLI::IsMember`, and deprecates the default `add_set*` options . This fixes the set lifetime and mutability issues in #170 and makes sets easy to modify after adding if you use a (shared) pointer. It also combines case insensitivity and underscore ignoring into validators, without the massive number of methods on App needed now. You can use any iterable that supports `std::begin`, `std::end`, and has `::value_type`, allowing a vector (for example) to be chosen to keep definition order - first item matches.

#### New features:

* Single new validator, no longer a massive list of functions in App
* Shared pointers supported, clear pointer vs. copy semantics
* Combinable custom filter functions
* All types support filter functions
* Other containers, including ordered containers allowed

#### Changes:

* The error for not being found in a set is now `CLI::ValidationError`
* All `add_set*` functions use the new backend, and are deprecated, planned removal of most of them in 1.9, and all of them after that. `mutable` versions may be removed even in 1.8.

#### Ideas:

A `->choices(...)` shortcut could be added, but the syntax is still up for debate, so not in this PR. Feel free to make suggestions.

#### Example of use:

```cpp
// Simple sets
app.add_option("--name", name)->check(CLI::IsMember({"tom", "harry"}));

// Sets can be modified later. `ignore_case` and `ignore_underscores` is addable.
std::set<std::string> names = {"tom", "harry"};
app.add_option("--name", name)->check(CLI::IsMember(&names, CLI::ignore_case));
names->insert("john");

// Shared pointers are supported, and different types. And ordered containers.
auto vals = std::make_shared<std::vector<int>>({1,2});
app.add_option("--val", val)->check(CLI::IsMember(&vals));
names->push_back(3);
```

You can use `std::shared_ptr` (that's internally what's used if you don't give a pointer-like). You can list `CLI::ignore_case` or `CLI::ignore_underscore` (or any other function that will be applied to both sides).

#### Still to do:

- [x] Deprecate old `add_set*` methods (can't be done on all of them because some are templates.)
- [x] Provide new versions of all the tests (coexist for a little while)
- Think about combined name + number style options (might not be part of this PR)
- [x] Add test with non-string function, like abs
- [x] Needs docs and examples
- [x] Fix `ignore_underscore` (no ending s) in the README.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -644,7 +644,7 @@ class App {
     }
 #endif
 
-    /// Add set of options (No default, temp reference, such as an inline set)
+    /// Add set of options (No default, temp reference, such as an inline set) DEPRECATED
     template <typename T>
     Option *add_set(std::string option_name,
                     T &member,           ///< The selected member of the set
@@ -656,7 +656,7 @@ class App {
         return opt;
     }
 
-    /// Add set of options (No default, set can be changed afterwords - do not destroy the set)
+    /// Add set of options (No default, set can be changed afterwords - do not destroy the set) DEPRECATED
     template <typename T>
     Option *add_mutable_set(std::string option_name,
                             T &member,                  ///< The selected member of the set
@@ -668,7 +668,7 @@ class App {
         return opt;
     }
 
-    /// Add set of options (with default, static set, such as an inline set)
+    /// Add set of options (with default, static set, such as an inline set) DEPRECATED
     template <typename T>
     Option *add_set(std::string option_name,
                     T &member,           ///< The selected member of the set
@@ -681,7 +681,7 @@ class App {
         return opt;
     }
 
-    /// Add set of options (with default, set can be changed afterwards - do not destroy the set)
+    /// Add set of options (with default, set can be changed afterwards - do not destroy the set) DEPRECATED
     template <typename T>
     Option *add_mutable_set(std::string option_name,
                             T &member,                  ///< The selected member of the set
@@ -694,7 +694,8 @@ class App {
         return opt;
     }
 
-    /// Add set of options, string only, ignore case (no default, static set)
+    /// Add set of options, string only, ignore case (no default, static set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_case)) instead")
     Option *add_set_ignore_case(std::string option_name,
                                 std::string &member,           ///< The selected member of the set
                                 std::set<std::string> options, ///< The set of possibilities
@@ -706,7 +707,8 @@ class App {
     }
 
     /// Add set of options, string only, ignore case (no default, set can be changed afterwards - do not destroy the
-    /// set)
+    /// set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_case)) with a (shared) pointer instead")
     Option *add_mutable_set_ignore_case(std::string option_name,
                                         std::string &member,                  ///< The selected member of the set
                                         const std::set<std::string> &options, ///< The set of possibilities
@@ -717,7 +719,8 @@ class App {
         return opt;
     }
 
-    /// Add set of options, string only, ignore case (default, static set)
+    /// Add set of options, string only, ignore case (default, static set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_case)) instead")
     Option *add_set_ignore_case(std::string option_name,
                                 std::string &member,           ///< The selected member of the set
                                 std::set<std::string> options, ///< The set of possibilities
@@ -730,6 +733,8 @@ class App {
     }
 
     /// Add set of options, string only, ignore case (default, set can be changed afterwards - do not destroy the set)
+    /// DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(...)) with a (shared) pointer instead")
     Option *add_mutable_set_ignore_case(std::string option_name,
                                         std::string &member,                  ///< The selected member of the set
                                         const std::set<std::string> &options, ///< The set of possibilities
@@ -741,7 +746,8 @@ class App {
         return opt;
     }
 
-    /// Add set of options, string only, ignore underscore (no default, static set)
+    /// Add set of options, string only, ignore underscore (no default, static set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_underscore)) instead")
     Option *add_set_ignore_underscore(std::string option_name,
                                       std::string &member,           ///< The selected member of the set
                                       std::set<std::string> options, ///< The set of possibilities
@@ -753,7 +759,8 @@ class App {
     }
 
     /// Add set of options, string only, ignore underscore (no default, set can be changed afterwards - do not destroy
-    /// the set)
+    /// the set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_underscore)) with a (shared) pointer instead")
     Option *add_mutable_set_ignore_underscore(std::string option_name,
                                               std::string &member,                  ///< The selected member of the set
                                               const std::set<std::string> &options, ///< The set of possibilities
@@ -764,7 +771,8 @@ class App {
         return opt;
     }
 
-    /// Add set of options, string only, ignore underscore (default, static set)
+    /// Add set of options, string only, ignore underscore (default, static set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_underscore)) instead")
     Option *add_set_ignore_underscore(std::string option_name,
                                       std::string &member,           ///< The selected member of the set
                                       std::set<std::string> options, ///< The set of possibilities
@@ -777,7 +785,8 @@ class App {
     }
 
     /// Add set of options, string only, ignore underscore (default, set can be changed afterwards - do not destroy the
-    /// set)
+    /// set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_underscore)) with a (shared) pointer instead")
     Option *add_mutable_set_ignore_underscore(std::string option_name,
                                               std::string &member,                  ///< The selected member of the set
                                               const std::set<std::string> &options, ///< The set of possibilities
@@ -789,7 +798,8 @@ class App {
         return opt;
     }
 
-    /// Add set of options, string only, ignore underscore and case (no default, static set)
+    /// Add set of options, string only, ignore underscore and case (no default, static set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_case, CLI::ignore_underscore)) instead")
     Option *add_set_ignore_case_underscore(std::string option_name,
                                            std::string &member,           ///< The selected member of the set
                                            std::set<std::string> options, ///< The set of possibilities
@@ -801,7 +811,9 @@ class App {
     }
 
     /// Add set of options, string only, ignore underscore and case (no default, set can be changed afterwards - do not
-    /// destroy the set)
+    /// destroy the set) DEPRECATED
+    CLI11_DEPRECATED(
+        "Use ->check(CLI::IsMember(..., CLI::ignore_case, CLI::ignore_underscore)) with a (shared) pointer instead")
     Option *add_mutable_set_ignore_case_underscore(std::string option_name,
                                                    std::string &member, ///< The selected member of the set
                                                    const std::set<std::string> &options, ///< The set of possibilities
@@ -812,7 +824,8 @@ class App {
         return opt;
     }
 
-    /// Add set of options, string only, ignore underscore and case (default, static set)
+    /// Add set of options, string only, ignore underscore and case (default, static set) DEPRECATED
+    CLI11_DEPRECATED("Use ->check(CLI::IsMember(..., CLI::ignore_case, CLI::ignore_underscore)) instead")
     Option *add_set_ignore_case_underscore(std::string option_name,
                                            std::string &member,           ///< The selected member of the set
                                            std::set<std::string> options, ///< The set of possibilities
@@ -825,7 +838,9 @@ class App {
     }
 
     /// Add set of options, string only, ignore underscore and case (default, set can be changed afterwards - do not
-    /// destroy the set)
+    /// destroy the set) DEPRECATED
+    CLI11_DEPRECATED(
+        "Use ->check(CLI::IsMember(..., CLI::ignore_case, CLI::ignore_underscore)) with a (shared) pointer instead")
     Option *add_mutable_set_ignore_case_underscore(std::string option_name,
                                                    std::string &member, ///< The selected member of the set
                                                    const std::set<std::string> &options, ///< The set of possibilities
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -651,18 +651,8 @@ class App {
                     std::set<T> options, ///< The set of possibilities
                     std::string description = "") {
 
-        std::string simple_name = CLI::detail::split(option_name, ',').at(0);
-        CLI::callback_t fun = [&member, options, simple_name](CLI::results_t res) {
-            bool retval = detail::lexical_cast(res[0], member);
-            if(!retval)
-                throw ConversionError(res[0], simple_name);
-            return std::find(std::begin(options), std::end(options), member) != std::end(options);
-        };
-
-        Option *opt = add_option(option_name, std::move(fun), std::move(description), false);
-        std::string typeval = detail::type_name<T>();
-        typeval += " in {" + detail::join(options) + "}";
-        opt->type_name(typeval);
+        Option *opt = add_option(option_name, member, std::move(description));
+        opt->check(IsMember{options});
         return opt;
     }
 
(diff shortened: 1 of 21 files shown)


henryiii:
@phlptp There are also some warnings about unsigned vs. signed comparison fixed - though it looks like you may have already fixed them in your PR, that's fine (that will probably be done before this one by quite a bit). Feedback on the design welcome. :)

include/CLI/App.hpp
```diff
@@ -1176,7 +1198,7 @@ class App {
     }
     /// Get a pointer to subcommand by index
     App *get_subcommand(int index = 0) const {
-        if((index >= 0) && (index < subcommands_.size()))
+        if((index >= 0) && ((size_t)index < subcommands_.size()))
```
phlptp: I suppose really you could use static cast and get rid of the (index>=0) check since a negative number cast into a large unsigned number.  I usually go the other way and cast the sizes to an int but It doesn't make that much difference.

henryiii: In this case, since the index was already checked for `>= 0`, this should be valid. This should be valid for all sizes of ints. Though no one is likely to have something large enough to matter.

include/CLI/Set.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+// CLI Library includes
+#include "CLI/StringTools.hpp"
+
+namespace CLI {
+
+class Set {
+  protected:
+    using storage_type = std::vector<std::string>;
+    std::shared_ptr<storage_type> items_{std::make_shared<storage_type>()};
```
phlptp: The question of how to enable mutable sets is going to be the tricky part here,  especially how to make it not unexpectedly mutable.

henryiii: Sets are always mutable. Not sure why you would add a Set, then change it and not expect it to change. If you change it, it changes in the App. The thing this fixes is you can’t segfault by deleting the set.

phlptp: I think my hesitancy is that when you see something passed by value it is usually copying(or moving) into it and it is then isolated from any changes that might be made at a higher level.    The shared_ptrs take care of lifetime issues with this, but it seems a little non-intuitive that any modifications to the set effect all previous uses of that set, even though it was passed by value.  

For example I have 1 use case where there are set of 5 different modes for one variable,  another has those 5 modes but adds an additional one.    Seeing sets are passed by value I might naively just add an element to the set and give it to the second option.  which would allow a hard to detect invalid set member in the first.  At the very least I am concerned that this will lead to some hard to detect and frustrating bugs on the part of users.  

I am actually not entirely sure what the best approach might be,  maybe take by pointer,  make a copy with the shared state,  then provide a deep copy function? .

include/CLI/Set.hpp
```diff
@@ -0,0 +1,115 @@
...
+#include <memory>
+#include <string>
+#include <vector>
+
+// CLI Library includes
+#include "CLI/StringTools.hpp"
+
+namespace CLI {
+
+class Set {
```
phlptp: I wonder about the possibility of making set a class template, then you could have sets of `<int>`, or other things as long as they had a lexical cast or `>>` operator.  
```
Option *add_cli_set(std::string option_name,
                        T &member,   ///< The selected member of the set
                        Set<T> options, ///< The set of possibilities
                        std::string description = "") {
```

phlptp: Along with a specialization for string that has the extra ignore_case and ignore_underscore functions

henryiii: I thought about that, I just left it off for simplicity for now. I'll probably add it as part of making it work with enum outputs. However, since this is for "possible choices", I think requiring strings would not be too bad if it can't be done.

include/CLI/Set.hpp
henryiii: Since CLI11 takes references as the main way of setting values, I felt that was pretty reasonable. However, since "mutability" is probably not as commonly needed, and should be clear to the user, I'm thinking that simplifying set and then taking two signatures, the set or a `shared_ptr` to a set, might be a good idea. The shared pointer version would clearly be mutable.

phlptp:
Just kind of poking around at things, but I was wondering if it might make sense to have the set member checking run through a Validator instead of in the lambda function.  Then whatever add_set functions in the App class are just wrappers for add_option, and a ->check(CLI::IsMember(CLI::set))  Could even add CLI::IsNotMember(CLI::set) quite easily, then those Validators could have overloads for set or shared_ptr to a set.  or chain set unions or whatever else you can do with Validators.   It is a little bigger API change but it would probably clear out some additional code from App.hpp.  Just a thought.

henryiii:
That could work! It maps what you are doing (validating an option is in a set) much better than a special add function. That might even work around the need to have a mutable set at all, since you can add the set later on after creation. (Changing it would not be easy though, if you really needed it changed for multiple parses)

phlptp:
I think in the case of reparsing, such as a terminal program or script interpreter,  having a mutable set be a little more directly modifiable is probably useful, since you could very well change it between parses or even inside the callbacks themselves.  Then I think supporting a shared_ptr to a set in an overload of the approvpriate Validator would make things straightforward.

henryiii:
For posterity, here's the old plan. PR now completely rewritten based on @phlptp's idea.

This is the start of CLI::Set. This fixes the set lifetime and mutability issues #170 and makes sets easy to modify after adding. It also combines case insensitivity and underscore ignoring into one package, without the massive number of methods on App needed now.

In the draft:

- The name of the method is `add_cli_set` instead of `add_set` while both implementations coexist
- The defaulted method is missing
- No solution yet for enums and numerical matching to sets (maybe provide T param)
- Try using std::shared_ptr as a way to pass in
- Needs lots of tests
- Needs to become the backend for the deprecated methods
- Needs docs and examples

And, of course, it needs feedback.

This should give an idea of some ways to use:

```cpp
app.add_cli_set("--name", name, {"tom", "harry"});

CLI::Set names("tom", "harry");
names->case_insensitive();
app.add_cli_set("--name", name, names);
names.insert("john");
```

phlptp:
@henryiii I was tinkering with a few methods of making flags a little more generalized by extending the default false syntax a little.  I think it will address #123 and #124.  I want to try it out in a real application to try to work out any issues.  I have a couple situations where the combination of the recent features will hopefully be able to collapse a bunch of different options into a single one.  I want to wait until the set is a little more developed to make a PR on it though as I think it will interact nicely with it.

henryiii:
Sounds good. This is nearly ready, just needs more testing and docs. I've been battling with the older compilers to get this to pass under C++11. I've had to give up nicer template errors but I think the basic syntax is intact.

It might be nice to have a shortcut method for creating a simple set inline, so far it's pretty convoluted.

```cpp
app->add_option("--set", value)->set(CLI::case_insensitive, "one", "two", "three");
```

henryiii:
I don't think I can make such a shortcut syntax work on GCC 4.7, it can't tell std::function from anything else.

phlptp:
I have had some luck in the past using std::is_assignable<std::function<...>,Args>  to check if something is a std::function or can be converted to one.  That only works if std::function<... > doesn't participate in the template type deduction though.    I think that works with C++11.  and use that as part of SFINAE to differentiate between function like objects and regular types

henryiii:
Okay, @phlptp, you can see what you think, that worked partially, but ran into the issue that `std::function<std::string(std::string)> = ""` seems to be valid in GCC 4.7 (okay on 4.8+). I've added an extra check for const char *, and eventually may drop GCC 4.7 support. Due to a bug in GCC 4.7, I am not sure the last version or two worked on a proper GCC 4.7 with GCC 4.7's stdlib. I still don't think many users are on 4.7, since there's no major LTS OS that has 4.7 that I know of (4.8, on the other hand, is RHEL/CentOS'7 main compiler so will be supported for many years).

So, current points for review / comments:

* `CLI::IsMember` has three signatures, pointer to container, container, and initializer_list of strings. The pointer can be any copyable pointer, such as a shared pointer. The containers can be any normal type, like a string or an int. `const char *` is not supported. Should I keep the initializer list shortcut?

```cpp
opt->check(CLI::IsMember({"Inline", "set", "only", "supports", "strings"}, CLI::ignore_case));
```

* `Option->choices("choice1", "choice2", CLI::ignore_case)` lets you quickly make a set inline from options. This one only supports strings. Modifier functions must come last. Is this the most useful sort of shortcut? I could also just allow `->choices(...)` to be identical to `->check(CLI::IsMember(...))`. This would allow it to also take sets and things, but would require extra `{}` around inline choices.

```cpp
opt->choices({"Inline", "set", "only", "supports", "strings"}, CLI::ignore_case);
// vs.
opt->choices("Inline", "set", "only", "supports", "strings", CLI::ignore_case); // current version in code
```

* No check for uniqueness is done. You can have a std::vector with duplicate entries, and the first one it finds will be matched. I could add a check, but I don't think it is that important?

Any other suggestions or review from anyone appreciated.

include/CLI/Validators.hpp
```diff
@@ -240,6 +264,116 @@ struct Range : public Validator {
...
+    explicit IsMember(T set, F filter_function) {
+        using item_t = typename T::value_type;
+        std::function<item_t(item_t)> filter_fn = filter_function;
+
+        std::stringstream out;
+        out << detail::type_name<item_t>() << " in {" << detail::join(set, ",") << "}";
+        tname = out.str();
+
+        func = [set, filter_fn](std::string &input) {
+            auto result = std::find_if(std::begin(set), std::end(set), [filter_fn, input](item_t v) {
```
phlptp: This might look cleaner/simpler as a range based for loop instead of the find_if algorithm with a complex lambda, and you could collapse the else condition after result into the loop.  and a boolean found.    You also will only need to update the results if filter_fn actually contains something.

phlptp: or just have the return directly from the loop if found

phlptp:
My experience with gcc 4.7 and 4.8 is limited,  Most of the projects I work on are currently at gcc 4.9+  to enable some of the C++14 stuff and we are starting to think about dropping 4.9 support.    I would say you might consider tagging a 1.8 once things are settled a little then dropping 4.7 support in a 2.0 then you can make some bigger breaking changes if you want to.

phlptp:
I don't think the uniqueness check is that important,  In most cases you are looking a just a small set of options and most users won't have duplicates.

phlptp:
I think I would keep the initializer list overload just so users don't have to create the container ahead of time.  

One thing I do use on occasion is an expression like this
```
constexpr const char *ztypes[] = {"ztype1", "ztype2", "ztype3"}; 
```
 though supporting that might get a little ugly right now.

phlptp:
Overall, I like it.  I think I need to actually try it as part of transitioning to CLI11 and reworking our code to use the new features.    I think it makes sense to have the choices syntax be different from the IsMember Validator syntax, and being able to just directly write in the strings without a {} seems intuitive.

include/CLI/Validators.hpp
henryiii: Done. I could split this into non-function versions too, so that `>>` would not be required on the type, but I think this is probably something that can be added later if anyone ever needs it.

henryiii:
One nice feature of the `{}` syntax: a user doesn't need to know that `->choices(..)` internally is the same as `->check(CLI::IsMember(...))` - you can access everything through choices. If you have an existing set, a shared pointer, etc., you can give them to choices. But, on the flip side, the brackets could be easy to miss and a common mistake. Probably depends on how users need the extra abilities...

henryiii:
> One thing I do use on occasion is an expression like this

This will work:

```cpp
     constexpr const char* ztypes[] = {"one", "two", "three"};
     std::string value;
     auto opt = app.add_option("-s,--set", value)->check(CLI::IsMember{std::vector<std::string>(std::begin(ztypes), std::end(ztypes))});
```

phlptp:
I think it comes down to a stylistic choice.  You could also consider dropping the choice member function completely.  From a code readability ->check(CLI::IsMember(...))  is going to be pretty clear as to what it is doing (but slightly more verbose).    Where as ->choices(...) is not bad but  it is not as clear and adds more code that needs to maintained and managed and documented all to say it does the same thing as check(CLI::IsMember(...)).

henryiii:
I like that idea - I'll drop choices then it can be readded based on usage later if needed.

include/CLI/Validators.hpp
```diff
@@ -240,6 +264,137 @@ struct Range : public Validator {
...
+    using filter_fn_t = std::function<std::string(std::string)>;
+
+    /// This allows in-place construction using an initializer list
+    template <typename... Args>
+    explicit IsMember(std::initializer_list<std::string> values, Args &&... args)
+        : IsMember(std::vector<std::string>(values), std::forward<Args>(args)...) {}
+
+    /// This checks to see if an item is in a set: pointer version. (Empty function)
+    template <typename T, enable_if_t<is_copyable_ptr<T>::value, detail::enabler> = detail::dummy>
+    explicit IsMember(T set)
```
phlptp, line 291: may want to consider std::move(set) for the pointer and list syntax for the inner IsMember call, otherwise you will probably get a few copies of Set made unnecessarily.  Still never sure if the compilers optimize away some of the copies in contexts like this.

include/CLI/Validators.hpp
```diff
@@ -240,6 +264,137 @@ struct Range : public Validator {
...
+        out << detail::type_name<item_t>() << " in {" << detail::join(set, ",") << "}";
+        tname = out.str();
+
+        // This is the function that validates
+        func = [set, filter_fn](std::string &input) {
+            for(const item_t &v : set) {
+                item_t a = v;
+                item_t b;
+                if(!detail::lexical_cast(input, b))
+                    throw ConversionError(input); // name is added later
```
phlptp: Should this be a validationError rather than ConversionError, since it will occur in the validation context, and having one of those throw a conversion Error when someone is checking for a Validation Error could be confusing,  possibly making no difference whatsoever but if you can't convert to the appropriate type of a given set it is a conversionError that generates a validationError.  Which one should be set to the user?

henryiii: The validator always runs first (that gives it the ability to affect the parse). So the following gives a conversion error:

```cpp
int myint;
app.add_option("--value", myint);
// Run with --value one
```

I didn't want the conversion error to change to a validation error here:

```cpp
int myint;
std::set<int> set = {1,2,3};
app.add_option("--value", myint)->check(CLI::IsMember(set));
// Run with --value one
```

But, both are true, and just saying "ValidationErrors" are stronger is probably fine.

You also have the odd mixed types that shouldn't happen but could:

```cpp
std::string mystr;
std::set<int> set = {1,2,3};
app.add_option("--value", mystr)->check(CLI::IsMember(set));
// Run with --value one
```
This fails only when checking the set and a ValidationError would fit better here, so I think I've convinced myself that ValidationError is better.

include/CLI/Validators.hpp
henryiii: I think move is okay here.

henryiii:
@SkyToGround, @kouchy, feel free to take a look too.

CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
     if(MSVC)
         add_definitions("/W4")
     else()
-        add_definitions(-Wall -Wextra -pedantic)
+        add_definitions(-Wall -Wextra -pedantic -Wpessimizing-move)
```
phlptp: This new warning is not available on GCC, at least in MSYS2,  causes the compilation to fail, and isn't -Wpessimizing-move part of -Wall?    The failure is possibly due to some of gtest being C code and that isn't a valid warning in C.  

I had to remove it to get things to compile on MSYS2,  (not CLI11) itself just the test suite and everything.

henryiii: Sorry, didn't mean to commit it! It's added by the other flags anyway, I think.

henryiii:
Other types should now be supported for the initializer list. I've added a helper that corrects `const char *` to `std::string`; a user can further specialize if needed for custom type overrides.

You still need a C++ class for the container for `::value_type`, so a plain array won't work.

SkyToGround:
@henryiii I was about to write a comment about using `enum class` but seeing as you fixed that in your last commit I currently do not have any additional comments. 👍

include/CLI/App.hpp
```diff
@@ -651,18 +651,8 @@ class App {
...
-                throw ConversionError(res[0], simple_name);
-            return std::find(std::begin(options), std::end(options), member) != std::end(options);
-        };
-
-        Option *opt = add_option(option_name, std::move(fun), std::move(description), false);
-        std::string typeval = detail::type_name<T>();
-        typeval += " in {" + detail::join(options) + "}";
-        opt->type_name(typeval);
+        Option *opt = add_option(option_name, member, std::move(description));
+        opt->check(IsMember{options});
```
SkyToGround, line 655: 👍

henryiii:
Will aim for a merge tomorrow if all looks good.

henryiii:
Posting these unused snippets of code for posterity, since I'll squash and merge:

```cpp
/// Allow a set to be quickly created
template <typename... Args> Option *choices(Args &&... args) {
    check(IsMember(std::forward<Args>(args)...));
    return this;
}
	
/// Allow a set to be quickly created (originally string only)
template <typename... Args> Option *choices(std::initializer_list<std::string> tmpset, Args &&... args) {
    std::vector<std::string> myset(tmpset);
    check(IsMember(myset, std::forward<Args>(args)...));
    return this;
}
```

And, the other form of the shortcut:

```cpp
            /// Allow a set to be quickly created, strings followed optionally by functions.
	    ///
	    /// Start: Will always have std::string
	    template <typename... Args> Option *choices(std::string value, Args &&... args) {
	        std::vector<std::string> values = {value};
	        return choices(values, std::forward<Args>(args)...);
	    }
	
	  private:
	    /// Final function called - no function
	    Option *choices(std::vector<std::string> values) {
	        check(IsMember(values));
	        return this;
	    }
	
            /// Final function called - function present
	    template <typename T,
	              enable_if_t<std::is_assignable<std::function<std::string(std::string)>, T>::value &&
	                              !(std::is_same<T, const char *>::value || std::is_same<T, char *>::value),
	                          detail::enabler> = detail::dummy>
	    Option *choices(std::vector<std::string> values, T fn) {
	        check(IsMember(values, fn));
	        return this;
	    }
	
	    /// Append a string
	    template <typename T,
	              typename... Args,
	              enable_if_t<std::is_assignable<std::string, T>::value, detail::enabler> = detail::dummy>
	    Option *choices(std::vector<std::string> values, T value, Args &&... args) {
	        values.push_back(std::move(value));
	        return choices(values, std::forward<Args>(args)...);
	    }
	
	    /// Combine functions
	    template <typename... Args>
	    Option *choices(std::vector<std::string> values,
	                    std::function<std::string(std::string)> fn1,
	                    std::function<std::string(std::string)> fn2,
	                    Args &&... args) {
	        std::function<std::string(std::string)> fn = [fn1, fn2](std::string val) { return fn1(fn2(val)); };
	        return choices(values, fn, std::forward<Args>(args)...);
	    }
	
```

PR #223, opened 2019-02-11
Allow specification that positionals only occur at end

### LINKED ISSUE #220: Is there a way to only parse options before positional arguments?
The SSH command, for example, only allows you to pass options before the host and command parameters. Is there a way to have similar behavior in CLI11 so that using `--` is not necessary?

phlptp:
If merged this PR will add a flag option on Apps that specify that positional arguments only occur at the tail of the argument sequence.

`.positionals_at_end()`

Extra arguments after the first positional that are not captured by other positionals will generate an ExtrasError  regardless of the state of allow_extras.

This closes #220 

It works by triggering the equivalent of a `--` after the first positional argument


henryiii:
Looks great! Thanks!

Issue #88, opened 2018-03-19
Requiring one Option or another, not both


gbitzes:
Hi, is there some way to require that one of { "--option1", "--option2" } is present, but not necessarily both? I looked at the documentation and examples, and couldn't figure out if this is possible. Do I need a custom callback?

By the way, excellent work on this library! Something like this was sorely needed in C++.

henryiii:
Not really, the current way to do that would be to use a set, but that's a different syntax. What would a good api for that be? Maybe passing other options to the `->required()` method?

gbitzes:
I was thinking of either having a variadic `require_one` method in `App`:

```
auto opt1 = app->add_option( ... );
auto opt2 = app->add_option( ... );

app->require_one(opt1, opt2, opt3);
// could also be app->require_at_least(2, opt1, opt2, opt3, opt4);
```

Or the ability to make a Group required, such as `app->require_group("Some group");`

The first seems more flexible (allows more complicated networks of requirements, and overlapping sets), but the second easily allows expressing this in the auto-generated help string for groups.

But then again, we could auto-detect if a group has an "at least N" requirement.. I'm leaning towards the first approach.

I'm not really sure.. not a big fan of a variadic `->required()` though, the intent doesn't seem clear when reading the code, IMHO.

Python's [argparse](https://docs.python.org/2/library/argparse.html#mutual-exclusion) takes the second approach.

OlivierHartmann:
Hi,
I am not completely sure this is the same issue but quite an identical need so I comment this flow.

Here is the simple example:

	#include <CLI/CLI.hpp>

	int main(int argc, char **argv) {

		CLI::App app{"App to demonstrate a paradox."};

		std::vector<int> range;
		auto range_option = app.add_option("--range,-R", range, "A range")->required();

		int min, max, step = 1;
		app.add_option("--min,-m", min, "The minimum")->required()->excludes(range_option);
		app.add_option("--max,-M", max, "The maximum")->required()->excludes(range_option);
		app.add_option("--step,-s", step,  "The step", true)->excludes(range_option);

		CLI11_PARSE(app, argc, argv);
	}

With "--range 1 2 3" argument, I get: "--min is required".
With "--min 2 --max 3 --step 1" arguments, I get: "--range: At least 1 required"
With "--min 2 --max 3 --step 1 --range 1 2 3" arguments, I get: "--range excludes --min"

A solution would be to remove the "required"  flag but then the user may give nonsense. Is there a solution ?

[...]

PR #227, opened 2019-02-19
Option groups

phlptp:
This PR add require_options(min,max)  functionality to App,  the functionality is similar to require_subcommand.  

Also create a subclass for Option_group which is currently a limited wrapper around an App, which constructs a nameless subcommand on an app.

the option_group is created through a standalone function so App doesn't have to be aware of the existence of Option_group.    This should address #88 as @AlexanderAmelkin, as he is correct that while a nameless subcommand with the require_option  will work it does seem kind of a hack.    With the option_group we may be able to add some additional better help and possibly some other features related to clusters of options though I am not sure what that might be at this point.  

more tests would need to be added if we like this approach.  

I think this is an alternative to the radio_id alternative that @AlexanderAmelkin is working on, not sure which is preferred by @henryiii.


henryiii:
I like the addition of the required option count, need to think about the named group App subclass idea, it might be better just to integrate the extra functionality into App. Also, you should be able to move that free function into a member, just either predeclare or make it a template function with a wasted automatic template parameter, I use that trick somewhere else in App or Option. Will revisit. A nice example of usage would help decide if it is a good idea to pursue.

phlptp:
I added a ranges example as shown in #88.  I think the help needs some work and I need more test cases and docs but I think that example can be run with this PR.  The basic idea is to treat option groups the same as options for counting purposes, and then allow options to exclude subcommands, and subcommands to exclude others.

henryiii:
I'd like a readme update, even if just so I can get a good feel for what this affects from a user perspective. Also, could this be able to replace the `->group` argument (at least for options)? This would add a name to unnamed subcommands (so basically, there would be two ways to make a subcommand, and those would put them into two different lists, "subcommands" and "groups". It might make printing the groups easier as well. For backward compatibility, `->group("name")` could add a name group if one does not exist, then move the option there? Just thoughts.

phlptp:
I thought about that a little, but if you were to make the current groups use the new option groups it would alter the parse order, which could be important for positional arguments.  And I think if you had two lists you lose some of the composability aspects of nameless subcommands, and dealing with switching between them could be a tricky issue.    Maybe worth exploring after getting a little experience with how this interacts with everything else.  Your ideas on full group objects might come into play as well.

phlptp:
I will work on the Readme next, but probably won't be able to finish it until this evening or tomorrow morning.

phlptp:
I updated the readMe with the changes, at least a first draft of the update.

README.md
```diff
@@ -53,7 +55,7 @@ CLI11 provides all the features you expect in a powerful command line parser, wi
 It is tested on [Travis][], [AppVeyor][], and [Azure][], and is being included in the [GooFit GPU fitting framework][goofit]. It was inspired by [`plumbum.cli`][plumbum] for Python. CLI11 has a user friendly introduction in this README, a more in-depth tutorial [GitBook][], as well as [API documentation][api-docs] generated by Travis.
 See the [changelog](./CHANGELOG.md) or [GitHub Releases][] for details for current and past releases. Also see the [Version 1.0 post][], [Version 1.3 post][], or [Version 1.6 post][] for more information.
 
-You can be notified when new releases are made by subscribing to <https://github.com/CLIUtils/CLI11/releases.atom> on an RSS reader, like Feedly.
+You can be notified when new releases are made by subscribing to <https://github.com/CLIUtils/CLI11/releases.atom> on an RSS reader, like Feedly, or use the releases mode of the github watching tool.
```
henryiii, line 60: +1 I was very happy when they added that :) (but still like the RSS feed).

henryiii:
Is this ready? I think we can merge, possibly iterating before the next release.

phlptp:
At your discretion,  If you are happy with the readme updates I think it is ready.  I am guessing there will need to be a few tweaks as there is more experience with it, but I don't know what those might be at the moment.

henryiii:
I’m going to consider a few changes/renames in a seperate issue in a few days. Thanks!

Issue #12, opened 2017-06-02
Support for enumerations


Maktm:
I tried the following code to see whether CLI11 supports the use of enumerations in `add_set` and got a series of errors. Is the code wrong is support for enumerations planned/not going to be added?

```cpp
#include <iostream>
#include <string>

#include <CLI/CLI.hpp>

enum Level : std::int32_t
{
  High,
  Medium,
  Low
};

int main(int argc, char* argv[])
{
  CLI::App app{"App description"};
  try
  {
    Level level;
    app.add_set("-l,--level", level, {Level::High, Level::Medium, Level::Low}, "Level settings");
    app.parse(argc, argv);
  }
  catch (CLI::Error const& e)
  {
    app.exit(e);
  }
  return EXIT_SUCCESS;
}
```

Errors:

    1>f:\github\cli11\include\cli\typetools.hpp(108): error C2440: '=': cannot convert from 'std::string' to 'Level'
    1>f:\github\cli11\include\cli\typetools.hpp(108): note: No user-defined-conversion operator available that can perform this conversion, or the operator cannot be called
    1>f:\github\cli11\include\cli\app.hpp(326): note: see reference to function template instantiation 'bool CLI::detail::lexical_cast<T,0>(std::string,T &)' being compiled
    1>        with
    1>        [
    1>            T=Level
    1>        ]
    1>f:\code\visual studio 2017\projects\cmdapp\cmdapp\main.cpp(20): note: see reference to function template instantiation 'CLI::Option *CLI::App::add_set<Level>(std::string,T &,std::set<Level,std::less<_Kty>,std::allocator<_Kty>>,std::string,bool)' being compiled
    1>        with
    1>        [
    1>            T=Level,
    1>            _Kty=Level
    1>        ]
    1>Done building project "cmdapp.vcxproj" -- FAILED.
    ========== Build: 0 succeeded, 1 failed, 0 up-to-date, 0 skipped ==========

henryiii:
I'm curious, what do you expect for the interface? `-l 0` or `-l High`? The second can't work since C++ does not have reflection, so it can't know that `High == "0"`. The first one should have worked, though.

I'll have to see why the SFINAE is selecting the string version for a enum that just wraps an int, though, that should not be happening.

`std::is_enum` needs to be added along with `std::is_int`. Would have thought `is_int` would include enums.

henryiii:
[...]

PR #228, opened 2019-02-19
Adding map support to IsMember

henryiii:
`CLI::IsMember` now supports maps. Example with an enum:

```cpp
std::map<std::string, Level> map = {{"High", Level::High},
                                    {"Medium", Level::Medium},
                                    {"Low", Level::Low}};
Level level;
app.add_option("-l,--level", level, "Level settings")
    ->required()
    ->transform(CLI::IsMember(map, CLI::ignore_case)
              | CLI::IsMember({Level::High, Level::Medium, Level::Low}));
```

And, on the command line:
```bash
$ ./examples/enum -h                                                                                                                                                         Tue Feb 19 23:27:48 2019
Usage: ./examples/enum [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
  -l,--level {High,Medium,Low} REQUIRED
                              Level settings

$ ./examples/enum -l high 
Enum received: 0

$ ./examples/enum -l low
Enum received: 2

$ ./examples/enum -l 1 
Enum received: 1

$ ./examples/enum -l 5
--level: 5 not in {High,Medium,Low} | 5 not in {0,1,2}
```

Changes:
* Containers of pairs convert from first (matched) to second (result) - this includes maps and the like
* Dropped "TEXT in" part of type name - should be clear enough
* Validator combinations are a bit better, and support functional type names
* Validators can be passed to `->transform` now; `->check` validators cannot modify the parse

Todo:

* [x] Flip the order of maps
* [x] Generalize to containers of pairs
* [x] Check to see if type_name setting on option can override functional type names
* [x] Add test with matching types
* [x] Add to docs
* [x] Add a test for transform with tname instead of only tname_function
* See if passing multiple functions to IsMember can avoid requiring `string(string)` form (might be in another PR)

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -299,9 +299,9 @@ class IsMember : public Validator {
     template <typename T, typename F> explicit IsMember(T set, F filter_function) {
 
         // Get the type of the contained item - requires a container have ::value_type
-        // if the type does not have mapped_type an key_type, these are both value_type
-        using element_t = typename detail::element_type<T>::type;                // Removes (smart) pointers if needed
-        using item_t = typename detail::key_map_adaptor<element_t>::mapped_type; // Is value_type if not a map
+        // if the type does not have first_type and second_type, these are both value_type
+        using element_t = typename detail::element_type<T>::type;            // Removes (smart) pointers if needed
+        using item_t = typename detail::pair_adaptor<element_t>::first_type; // Is value_type if not a map
 
         using local_item_t = typename IsMemberType<item_t>::type; // This will convert bad types to good ones
                                                                   // (const char * to std::string)
@@ -315,7 +315,7 @@ class IsMember : public Validator {
             out << "{";
             int i = 0; // I don't like counters like this
             for(const auto &v : detail::smart_deref(set))
-                out << (i++ == 0 ? "" : ",") << detail::key_map_adaptor<element_t>::second(v);
+                out << (i++ == 0 ? "" : ",") << detail::pair_adaptor<element_t>::first(v);
             out << "}";
             return out.str();
         };
@@ -324,7 +324,7 @@ class IsMember : public Validator {
         // It stores a copy of the set pointer-like, so shared_ptr will stay alive
         func = [set, filter_fn](std::string &input) {
             for(const auto &v : detail::smart_deref(set)) {
-                local_item_t a = detail::key_map_adaptor<element_t>::second(v);
+                local_item_t a = detail::pair_adaptor<element_t>::first(v);
                 local_item_t b;
                 if(!detail::lexical_cast(input, b))
                     throw ValidationError(input); // name is added later
@@ -338,10 +338,10 @@ class IsMember : public Validator {
                 if(a == b) {
                     // Make sure the version in the input string is identical to the one in the set
                     // Requires std::stringstream << be supported on T.
-                    // If this is a map, ouptut the map instead.
-                    if(filter_fn || detail::key_map_adaptor<element_t>::value) {
+                    // If this is a map, output the map instead.
+                    if(filter_fn || detail::pair_adaptor<element_t>::value) {
                         std::stringstream out;
-                        out << detail::key_map_adaptor<element_t>::first(v);
+                        out << detail::pair_adaptor<element_t>::second(v);
                         input = out.str();
                     }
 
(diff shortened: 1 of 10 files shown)


henryiii:
Open question: Which direction should a map go? Currently values are matched (->second), and keys are returned (->first), but the other direction, while maybe less natural, makes more sense. You can get non-unique outputs, but non-unique input doesn't really make sense.

So currently:

```python
{1: 'one',
 2: 'two}
```

If flipped:

```python
{'one': 1,
 'two': 2,
 'uno': 1}
```

phlptp:
In #229  I was playing with translate functions.  I was using std::vector<std:pair<std::string, X>> as the basic type, as I didn't want to add maps but it would make sense to if IsMember supports them.  In the context of translate the natural ordering would be {from, to}  especially since there may be many strings that go to the same final element.   

Maps are kind of interesting in that you might want to actually use the fast searching capabilities of a map.  I could easily picture a case where you want to make sure an argument has a value in a hash table that is potentially large.  

I would argue that isMember should not be modifying values.  In the limited context of the functions I think it makes sense to match the member that is actually in the set, but with a map you are talking about much more significant transformations, and you add a requirement that the value part of the map must be convertible,  whereas it is quite reasonable that you might want to search a structure that has a complex class as a value that you don't want translated.    My instinct would be in the case of a map you don't want any requirement on the value portion of the map, since isMember should not be doing any translation other than in the presence of filter_functions making sure the actual value is a Member of original set.  Any translation should be left to something that make that functionality much clearer.

henryiii:
IsMember already has to modify the parsing, since things like `ignore_case` let you match against strings not technically in the set, but a user probably wants to see which item was matched against.

As you can see by the fact almost no code was added here, just minor type adaptors, this is exactly what IsMember was already doing; using a map (I can generalize to any container of pairs) just allows a user to control what happens. I prefer one flexible implementation over large numbers of different tools that all need extensive testing, etc. And this should be better for compile time, etc.

Just a thought I'll look into: how about check and transform both taking Validators, and only transform allowing the modification aspect to work? Then it would be more clear. We can also think about a rename for IsMember, too. Choices, for example?

henryiii:
Both are now supported, and there is a difference between:

```cpp
option->check(CLI::IsMember(...));
option->transform(CLI::IsMember(...));
```

Only the second version will modify the parse result, allowing a user to choose if they want the original string or the matched string. This is very nice for things like `CLI::ignore_case`, because you might want the original string, or you might want the one that was in the set for easy matching. If you use vectors of pairs or maps (both supported), the `transform` one is usually what you want, since you can use it to convert string matches into enums or integers.

I've flipped the order for maps/pairs, as well.

henryiii:
Should the `IsMember` name be changed?

henryiii:
This partially addresses #12, but I think @phlptp has one more improvement as well. @phlptp, can you take a look and see if you think this is ready to be merged? There's still the name issue; IsMember reads well for sets but maybe less so for maps (though since it is written `->transform(CLI::IsMember`, that helps).

include/CLI/Validators.hpp
```diff
@@ -287,34 +292,39 @@ class IsMember : public Validator {
...
 
         // This is the type name for help, it will take the current version of the set contents
         tname_function = [set]() {
             std::stringstream out;
-            out << detail::type_name<item_t>() << " in {" << detail::join(detail::smart_deref(set), ",") << "}";
+            out << "{";
+            int i = 0; // I don't like counters like this
+            for(const auto &v : detail::smart_deref(set))
+                out << (i++ == 0 ? "" : ",") << detail::pair_adaptor<element_t>::first(v);
+            out << "}";
```
phlptp, line 319: In one of the PR's I made a join overload that could take a lambda, that would get rid of the counter.

phlptp:
I like where this is headed.  The idea that check doesn't modify I think fits better, but it does add another layer of indirection, so are you ending up with what could easily be 3 layers of std::function calls.    The naming on translate(IsMember) doesn't read well yet, not sure how to do it differently though right now,  But I do think in a few more iterations this could be nailed down into a pretty solid interface.    The best path forward might be to merge this and a few of the other PR's and see where things are at, as I am pretty sure they interact and could build on eachother.

phlptp (approved):
I think there will be a few more iterations on this idea before it settles down, but this is a good step and basis for further tweaks.

include/CLI/Validators.hpp
henryiii: Noted, will try to switch to this once that is merged.

PR #229, opened 2019-02-19
[WIP] Generalized flag values

phlptp:
This PR will make flag more general in that they can handle any non-vector type and have default values of any type.  

It adds better enum support

adds a transformer and checkedTransformer validators.    

Also adds ["option_name"] syntax for getting options and as<type> for querying results to be translated to any available type.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -291,7 +291,7 @@ class App {
         return this;
     }
 
-    /// specify that the positional arguments are only at the end of the sequence
+    /// specify that the positional arguments are only at the end of the sequence. Subcommands DO NOT inherit this value
     App *positionals_at_end(bool value = true) {
         positionals_at_end_ = value;
         return this;
@@ -353,16 +353,16 @@ class App {
     ///
     Option *add_option(std::string option_name,
                        callback_t option_callback,
-                       std::string description = "",
+                       std::string option_description = "",
                        bool defaulted = false) {
-        Option myopt{option_name, description, option_callback, defaulted, this};
+        Option myopt{option_name, option_description, option_callback, defaulted, this};
 
         if(std::find_if(std::begin(options_), std::end(options_), [&myopt](const Option_p &v) {
                return *v == myopt;
            }) == std::end(options_)) {
             options_.emplace_back();
             Option_p &option = options_.back();
-            option.reset(new Option(option_name, description, option_callback, defaulted, this));
+            option.reset(new Option(option_name, option_description, option_callback, defaulted, this));
             option_defaults_.copy_to(option.get());
             return option.get();
         } else
@@ -373,11 +373,11 @@ class App {
     template <typename T, enable_if_t<!is_vector<T>::value, detail::enabler> = detail::dummy>
     Option *add_option(std::string option_name,
                        T &variable, ///< The variable to set
-                       std::string description = "") {
+                       std::string option_description = "") {
 
         CLI::callback_t fun = [&variable](CLI::results_t res) { return detail::lexical_cast(res[0], variable); };
 
-        Option *opt = add_option(option_name, fun, description, false);
+        Option *opt = add_option(option_name, fun, option_description, false);
         opt->type_name(detail::type_name<T>());
         return opt;
     }
@@ -386,7 +386,7 @@ class App {
     template <typename T, enable_if_t<!is_vector<T>::value, detail::enabler> = detail::dummy>
     Option *add_option_function(std::string option_name,
                                 const std::function<bool(const T &)> &func, ///< the callback to execute
-                                std::string description = "") {
+                                std::string option_description = "") {
 
         CLI::callback_t fun = [func](CLI::results_t res) {
             T variable;
@@ -397,21 +397,34 @@ class App {
             return result;
         };
 
-        Option *opt = add_option(option_name, std::move(fun), description, false);
+        Option *opt = add_option(option_name, std::move(fun), option_description, false);
         opt->type_name(detail::type_name<T>());
         return opt;
     }
+    /// add option with no description or variable assignment
+    Option *add_option(std::string option_name) {
+        return add_option(option_name, CLI::callback_t(), std::string{}, false);
+    }
+
+    /// add option with description but with no variable assignment or callback
+    template <typename T,
+              enable_if_t<std::is_const<T>::value && std::is_constructible<std::string, T>::value, detail::enabler> =
+                  detail::dummy>
+    Option *add_option(std::string option_name, T &option_description) {
+        return add_option(option_name, CLI::callback_t(), option_description, false);
+    }
 
     /// Add option for non-vectors with a default print
-    template <typename T, enable_if_t<!is_vector<T>::value, detail::enabler> = detail::dummy>
+    template <typename T,
+              enable_if_t<!is_vector<T>::value && !std::is_const<T>::value, detail::enabler> = detail::dummy>
     Option *add_option(std::string option_name,
                        T &variable, ///< The variable to set
-                       std::string description,
+                       std::string option_description,
                        bool defaulted) {
 
         CLI::callback_t fun = [&variable](CLI::results_t res) { return detail::lexical_cast(res[0], variable); };
 
-        Option *opt = add_option(option_name, fun, description, defaulted);
+        Option *opt = add_option(option_name, fun, option_description, defaulted);
         opt->type_name(detail::type_name<T>());
         if(defaulted) {
             std::stringstream out;
@@ -425,14 +438,14 @@ class App {
     template <typename T>
     Option *add_option(std::string option_name,
                        std::vector<T> &variable, ///< The variable vector to set
-                       std::string description = "",
+                       std::string option_description = "",
                        char delimiter = '\0') {
 
         CLI::callback_t fun = [&variable, delimiter](CLI::results_t res) {
             bool retval = true;
             variable.clear();
             for(const auto &elem : res) {
-                if(delimiter != '\0') {
+                if((delimiter != '\0') && (elem.find_first_of(delimiter) != std::string::npos)) {
                     for(const auto &var : CLI::detail::split(elem, delimiter)) {
                         if(!var.empty()) {
                             variable.emplace_back();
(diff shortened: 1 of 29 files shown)


phlptp:
@henryiii   This what I have been tinkering with,  Not sure you would want any or all of it.    My target is making a transition from boost::program options in a complex layered setup a little easier.    I also saw you were playing with some map stuff, which is related to some things I was tinkering with related to Transformer.  I suspect you probably don't want both but they may be able to be merged.    

Right now the PR has a number of merges from the recent set PR so that is why there is so many commits.  I might be able to peel off some of them to make a few other PRs of specific components.

phlptp:
This will also address Issues #12 , #123, and #124  I think

henryiii:
Needs a rebase or something equivalent. Sounds exciting. Wonder how it compares to #228. Edit: yes, as you mentioned in your comment.

henryiii:
It would be nice to at least split into two PRs, one for transformer and checkedTransformer validators. I think I'd like to compare with the IsMember validator. The other parts sound useful and could be evaluated and merged better on their own.

include/CLI/App.hpp
```diff
@@ -490,157 +503,191 @@ class App {
...
 
-    /// Add option for flag
-    Option *add_flag(std::string flag_name, std::string description = "") {
-        CLI::callback_t fun = [](CLI::results_t) { return true; };
-        Option *opt = add_option(flag_name, fun, description, false);
-        if(opt->get_positional())
-            throw IncorrectConstruction::PositionalFlag(flag_name);
+  private:
+    /// internal function for adding a flag
+    Option *add_flag_internal(std::string &flag_name, CLI::callback_t &fun, std::string &flag_description) {
```
henryiii, line 573: This should start with an underscore, internal functions all start with underscore. (Python style)

include/CLI/App.hpp
```diff
@@ -490,157 +503,191 @@ class App {
...
                      T &flag_count, ///< A variable holding the count
-                     std::string description = "") {
+                     std::string flag_description = "") {
         flag_count = 0;
-        Option *opt;
         CLI::callback_t fun = [&flag_count](CLI::results_t res) {
-            detail::sum_flag_vector(res, flag_count);
+            try {
+                detail::sum_flag_vector(res, flag_count);
+            } catch(const std::invalid_argument &) {
```
henryiii, line 623: I don't think I've seen `invalid_argument`, nice.

include/CLI/App.hpp
```diff
@@ -649,21 +696,21 @@ class App {
...
+                    std::string option_description = "") {
 
-        Option *opt = add_option(option_name, member, std::move(description));
+        Option *opt = add_option(option_name, member, std::move(option_description));
         opt->check(IsMember{options});
         return opt;
     }
 
-    /// Add set of options (No default, set can be changed afterwords - do not destroy the set) DEPRECATED
+    /// Add set of options (No default, set can be changed afterwards - do not destroy the set)
```
henryiii, line 706: DEPRECATED, please :)

include/CLI/App.hpp
```diff
@@ -1724,10 +1785,8 @@ class App {
             }
         }
 
-        Option *op;
-        try {
-            op = get_option("--" + item.name);
-        } catch(const OptionNotFound &) {
+        Option *op = get_option_no_throw("--" + item.name);
+        if(op == nullptr) {
```
henryiii, line 1789: Nice, I like avoiding exceptions as control flow if possible.

include/CLI/ConfigFwd.hpp
```diff
@@ -69,14 +69,10 @@ class Config {
     /// Convert a configuration into an app
     virtual std::vector<ConfigItem> from_config(std::istream &) const = 0;
 
-    /// Convert a flag to a bool representation
+    /// get a flag value
```
henryiii, line 72: First char caps for docstrings and for comments alone on a line. Comments that follow code on the same line are not capitalized. Okay for now, but slightly easier if I don't have to correct it later.

include/CLI/Option.hpp
```diff
@@ -330,6 +332,25 @@ class Option : public OptionBase<Option> {
         return this;
     }
 
+    /// Adds a transformation/validator with a built in type name
+    Option *transform(const Transformer &tform) {
```
henryiii, line 336: I've applied a very similar idea in my PR, but without the strong typing. I could add typing so that a Validator could not be added to ->transform, but the ability to choose whether a tranformer actually transforms or not is really nice for a user, IMO. It also avoids duplication.

include/CLI/Option.hpp
```diff
@@ -745,6 +791,74 @@ class Option : public OptionBase<Option> {
     /// Get a copy of the results
     std::vector<std::string> results() const { return results_; }
 
+    /// get the results as a particular type
+    template <typename T,
+              enable_if_t<!is_vector<T>::value && !std::is_const<T>::value, detail::enabler> = detail::dummy>
+    void results(T &output) const {
```
henryiii, line 797: Interesting! I've been looking for a way to add types to Option for 2.0, but this partially (not fully) negates the need to do that. The tricky part about the 2.0 idea is that `Option<T>` would need to be stored in the shared/unique pointer array, and I was thinking it would need a reference or copy of T, but this might avoid that (at a cost per access).

Does this pick up the results of a transform? I assume so since they should be transformed in place?

henryiii: Could this be broken out into a standalone function(s including overload)? Then it could be reused by the normal parse also, possibly? That is, this would become something like:
```
template<typename T>
void results(T &output) {
    CLI::detail::results(output, results_, defaultval_, multi_option_policy_);
}
```
And `CLI::detail::results` could be used in the parse function, too. Maybe not, just thinking. Also, it might be possible to make `CLI::detail::results` a struct, and it could have things like default number of arguments attached, and a user could provide specializations to add support for new types. (now well outside the scope of this PR, just thinking)

include/CLI/StringTools.hpp
```diff
@@ -13,6 +13,14 @@
 #include <vector>
 
 namespace CLI {
+
+/// output streaming for enumerations
+template <typename T, typename = typename std::enable_if<std::is_enum<T>::value>::type>
+std::ostream &operator<<(std::ostream &in, const T &level) {
```
henryiii, line 19: Nice addition that I didn't think of. It might be a good idea to make it opt-in in some way though, so a user doesn't clash if they already have something similar. Like `using namespace CLI::enums`?

include/CLI/StringTools.hpp
```diff
@@ -163,39 +186,50 @@ inline std::string find_and_replace(std::string str, std::string from, std::stri
...
-                            const std::vector<std::string> names,
-                            bool ignore_case = false,
-                            bool ignore_underscore = false) {
+inline std::ptrdiff_t find_member(std::string name,
+                                  const std::vector<std::string> names,
+                                  bool ignore_case = false,
+                                  bool ignore_underscore = false) {
+    auto it = std::end(names);
     if(ignore_case) {
         if(ignore_underscore) {
```
henryiii, line 212: Would it make sense to have a vector of transform functions, and allow a user to add any transforms they want; something like `opt->filter(CLI::ignore_underscore)->filter(CLI::ignore_case)`? Vector would be part of default option, etc. Just a thought for later.

include/CLI/TypeTools.hpp
```diff
@@ -101,20 +102,79 @@ template <typename T, enable_if_t<is_vector<T>::value, detail::enabler> = detail
 constexpr const char *type_name() {
     return "VECTOR";
 }
+/// print name for enumeration types
+template <typename T, enable_if_t<std::is_enum<T>::value, detail::enabler> = detail::dummy>
+constexpr const char *type_name() {
+    return "ENUM";
```
henryiii, line 108: Nice.

include/CLI/Validators.hpp
```diff
@@ -96,6 +96,25 @@ class Validator {
         };
         return newval;
     }
+
+    /// create a validator that fails when a given validator succeeds
+    Validator operator!() const {
+        Validator newval;
+        newval.tname = "NOT " + tname;
```
henryiii, line 103: You might want to support tname_function, too.

tests/app_helper.hpp
```diff
@@ -41,15 +41,15 @@ class TempFile {
 };
 
 inline void put_env(std::string name, std::string value) {
-#ifdef _MSC_VER
+#ifdef _WIN32
```
henryiii, line 44: I am not very familiar with non-MSVC compilers on Windows anymore, so never know what is best here. I'll assume that non-MSVC compilers are still able to access the Windows calls, rather than GNU counterparts? It would be nice to add a test to AppVeyor (or Azure) someday (not as part of this PR, though).

include/CLI/StringTools.hpp
henryiii: Fantastic, inside an exported namespace it works great, I'm adding this to my PR if you don't mind...

phlptp:
This might to much to adequately review, I will try to split some things up into different PR's,  they were all interconnected but I think there are some distinct pieces that are are valid on their own.

tests/app_helper.hpp
phlptp: The issue was MSYS, which is mingw,  deep in some of the libraries is uses some of the windows subsystems, so has most of the posix calls,  but when you get down into environmental variables you use the windows calls,  which I think was the case here.

include/CLI/StringTools.hpp
phlptp: Feel free,  It took a little while to figure out where to actually put this in the code so it was actually visible to all the functions that might need it.  I pulled it from the enum.cpp example, so it would work with the transformer Validator.

include/CLI/StringTools.hpp
phlptp: One thing I ran into was that translate functions should run before the actual validators so I had to make sure they were inserted at the front of the validator vector.    
So I think it is kind of an interesting usability discussion of whether check() should be allowed to modify or not.    On one had that differentiates check and translate but it disallows a CheckedTranslate since that would translate and make sure it was translated as a single unit.  You would need a translate and a separate check.

phlptp:
I added a number of PR #230, #232, #233, #235.  That try to parse out specific components of this PR.   I haven't added one for the transform piece yet since I think #228 will modify it heavily.

phlptp:
I am not going to fix this PR so it will be closed eventually, but am leaving in place for the moments since it contains a useful discussion at least until the PR's are merged/closed out.

phlptp:
This will be closed when #239 is merged or closed

phlptp:
All parts of this have now been merged

PR #232, opened 2019-02-20
clear up some additional Wshadow warnings on older compilers

phlptp:
This PR adds Wshadow flag and clears up some warnings


henryiii:
So that's why you've been renaming those. Okay.

phlptp:
I kept getting a large number of warnings when trying to use CLI11 in some projects.  I think most of the warnings on older gcc are mostly false positives but Wshadow has saved me a few times with the newer ones, so I like to leave it on.

henryiii:
Looks good! Thanks!

PR #233, opened 2019-02-20
Add better enum support in the library

phlptp:
This PR is part of addressing #12.   It adds better translation to/from enumerations.


henryiii:
I like it, thanks! I think we can finally close #12! :)

PR #235, opened 2019-02-20
Allow general flag types

phlptp:
If merged this pull request will allow general flag values to be used, along with customized default values.  It also introduces a shortcut syntax for retrieving flag and option values.  Which is useful for allowing options with no storage variables.  

```
app["--flag"].as<std::string>() 
```
this is similar notation to some other Command line libraries and could make transitioning to CLI easier

It also allows numbers to be used as short flags, so `-7` and similar is now allowed

This could be specified to have a default value like `-7{7}`  in which case if the value -7 were passed it would produce a result of `7` in the output.  

This PR addresses #123 and #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,13 +177,15 @@ The initialization is just one line, adding options is just two each. The parse 
 While all options internally are the same type, there are several ways to add an option depending on what you need. The supported values are:
 
 ```cpp
+app.add_option(option_name, help_str="")
+
 app.add_option(option_name,
-               variable_to_bind_to, // bool, int, float, vector, enum, or string-like
+               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string
                help_string="",
                default=false)
 
 app.add_option_function<type>(option_name,
-               function <void(const type &value)>, // int, float, enum, vector, or string-like
+               function <void(const type &value)>, // int, bool, float, enum, vector, or string-like, or anything with a defined conversion from a string
                help_string="")
 
 app.add_complex(... // Special case: support for complex numbers
@@ -192,17 +194,19 @@ app.add_flag(option_name,
              help_string="")
 
 app.add_flag(option_name,
-             int_or_bool,
+             variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string
              help_string="")
 
 app.add_flag_function(option_name,
-             function <void(int count)>,
+             function <void(int64_t count)>,
              help_string="")
+
+app.add_flag_callback(option_name,function<void(void)>,help_string="")
 
 App* subcom = app.add_subcommand(name, description);
 ```
 
-An option name must start with a alphabetic character or underscore. For long options, anything but an equals sign or a comma is valid after that. Names are given as a comma separated string, with the dash or dashes. An option or flag can have as many names as you want, and afterward, using `count`, you can use any of the names, with dashes as needed, to count the options. One of the names is allowed to be given without proceeding dash(es); if present the option is a positional option, and that name will be used on help line for its positional form. If you want the default value to print in the help description, pass in `true` for the final parameter for `add_option`.
+An option name must start with a alphabetic character, underscore, or a number. For long options, anything but an equals sign or a comma is valid after that, though for the `add_flag*` functions '{' has special meaning. Names are given as a comma separated string, with the dash or dashes. An option or flag can have as many names as you want, and afterward, using `count`, you can use any of the names, with dashes as needed, to count the options. One of the names is allowed to be given without proceeding dash(es); if present the option is a positional option, and that name will be used on help line for its positional form. If you want the default value to print in the help description, pass in `true` for the final parameter for `add_option`.
 
 The `add_option_function<type>(...` function will typically require the template parameter be given unless a `std::function` object with an exact match is passed.  The type can be any type supported by the `add_option` function.
 
@@ -210,27 +214,37 @@ Flag options specified through the functions
 
 ```cpp
 app.add_flag(option_name,
-             int_or_bool,
+             help_string="")
+
+app.add_flag(option_name,
+             variable_to_bind_to,
              help_string="")
 
 app.add_flag_function(option_name,
-             function <void(int count)>,
+             function <void(int64_t count)>,
              help_string="")
+
+app.add_flag_callback(option_name,function<void(void)>,help_string="")
 ```
 
-which allow a syntax for the option names to default particular options to a false value if some flags are passed.  For example:
+which allow a syntax for the option names to default particular options to a false value or any other value if some flags are passed.  For example:
 
 ```cpp
 app.add_flag("--flag,!--no-flag,result,"help for flag");`
 ``````
 
 specifies that if `--flag` is passed on the command line result will be true or contain a value of 1. If `--no-flag` is
-passed result will contain false or -1 if result is a signed integer type, or 0 if it is an unsigned type.  An
+passed `result` will contain false or -1 if `result` is a signed integer type, or 0 if it is an unsigned type.  An
 alternative form of the syntax is more explicit: `"--flag,--no-flag{false}"`; this is equivalent to the previous
-example.  This also works for short form options `"-f,!-n"` or `"-f,-n{false}"` If `int_or_bool` is a boolean value the
-default behavior is to take the last value given, while if `int_or_bool` is an integer type the behavior will be to sum
-all the given arguments and return the result.  This can be modified if needed by changing the `multi_option_policy` on
-each flag (this is not inherited).
+example.  This also works for short form options `"-f,!-n"` or `"-f,-n{false}"` If `variable_to_bind_to` is anything but an integer value the
+default behavior is to take the last value given, while if `variable_to_bind_to` is an integer type the behavior will be to sum
+all the given arguments and return the result.  This can be modified if needed by changing the `multi_option_policy` on each flag (this is not inherited).
+The default value can be any value For example if you wished to define a numerical flag
+```cpp
+app.add_flag("-1{1},-2{2},-3{3}",result,"numerical flag")
+```
+using any of those flags on the command line will result in the specified number in the output.  Similar things can be done for string values, and enumerations, as long as the default value can be converted to the given type.  
+
 
 On a C++14 compiler, you can pass a callback function directly to `.add_flag`, while in C++11 mode you'll need to use `.add_flag_function` if you want a callback function. The function will be given the number of times the flag was passed. You can throw a relevant `CLI::ParseError` to signal a failure.
 
@@ -258,6 +272,7 @@ Before parsing, you can set the following options:
 -   `->group(name)`: The help group to put the option in. No effect for positional options. Defaults to `"Options"`. `""` will not show up in the help print (hidden).
 -   `->ignore_case()`: Ignore the case on the command line (also works on subcommands, does not affect arguments).
 -   `->ignore_underscore()`: Ignore any underscores in the options names (also works on subcommands, does not affect arguments). For example "option_one" will match with "optionone".  This does not apply to short form options since they only have one character
+-   `->disable_flag_override()`:  from the command line long form flag option can be assigned a value on the command line using the `=` notation `--flag=value`. If this behavior is not desired, the `disable_flag_override()` disables it and will generate an exception if it is done on the command line.  The `=` does not work with short form flag options.
 -   `->description(str)`: Set/change the description.
 -   `->multi_option_policy(CLI::MultiOptionPolicy::Throw)`: Set the multi-option policy. Shortcuts available: `->take_last()`, `->take_first()`, and `->join()`. This will only affect options expecting 1 argument or bool flags (which do not inherit their default but always start with a specific policy).
 -   `->check(CLI::IsMember(...))`: Require an option be a member of a given set. See below for options.
@@ -296,7 +311,7 @@ On the command line, options can be given as:
 -   `-ffilename` (no space required)
 -   `-abcf filename` (flags and option can be combined)
 -   `--long` (long flag)
--   `--long_flag=true` (long flag with equals)
+-   `--long_flag=true` (long flag with equals to override default value)
 -   `--file filename` (space)
 -   `--file=filename` (equals)
 
@@ -306,9 +321,10 @@ If `allow_windows_style_options()` is specified in the application or subcommand
 -   `/long` (long flag)
 -   `/file filename` (space)
 -   `/file:filename` (colon)
+-   `/long_flag:false (long flag with : to override the default value)
 =  Windows style options do not allow combining short options or values not separated from the short option like with `-` options
 
-Long flag options may be given with and `=<value>` to allow specifying a false value See [config files](#configuration-file) for details on the values supported.  NOTE: only the `=` or `:` for windows-style options may be used for this, using a space will result in the argument being interpreted as a positional argument.  This syntax can override the default (true or false) values.
+Long flag options may be given with an `=<value>` to allow specifying a false value, or some other value to the flag. See [config files](#configuration-file) for details on the values supported.  NOTE: only the `=` or `:` for windows-style options may be used for this, using a space will result in the argument being interpreted as a positional argument.  This syntax can override the default values, and can be disabled by using `disable_flag_override()`.
 
 Extra positional arguments will cause the program to exit, so at least one positional option with a vector is recommended if you want to allow extraneous arguments.
 If you set `.allow_extras()` on the main `App`, you will not get an error. You can access the missing options using `remaining` (if you have subcommands, `app.remaining(true)` will get all remaining options, subcommands included).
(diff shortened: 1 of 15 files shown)


phlptp:
One question is whether there should be a method on options to disable flag value overrides.  for example you are able to specify 
```add_flag("--count{345}");```

but on the command line `--count=45`  
The question is should there be a way to disallow the =45 as an override?  and if that were disabled what should the behavior be (ignore, or error)?

**Comment by henryiii** [comment]
Not sure about the proper way to disable it, but I think an error is better than ignoring an issue.

PS:  I'll be traveling on vacation from tomorrow till Tuesday, so might not be able to do very complicated reviews for a while.

File: examples/CMakeLists.txt [hunk 1]
```diff
@@ -130,7 +130,12 @@ add_cli_exe(enum enum.cpp)
 add_test(NAME enum_pass COMMAND enum -l 1)
 add_test(NAME enum_fail COMMAND enum -l 4)
 set_property(TEST enum_fail PROPERTY PASS_REGULAR_EXPRESSION
-    "--level: 4 not in {High,Medium,Low} | 4 not in {0,1,2}")
+    "--level: 4 not in {")
```
henryiii: Is this really different now?

include/CLI/App.hpp
```diff
@@ -402,9 +402,22 @@ class App {
...
         return opt;
     }
+    /// Add option with no description or variable assignment
+    Option *add_option(std::string option_name) {
+        return add_option(option_name, CLI::callback_t(), std::string{}, false);
+    }
+
+    /// Add option with description but with no variable assignment or callback
+    template <typename T,
+              enable_if_t<std::is_const<T>::value && std::is_constructible<std::string, T>::value, detail::enabler> =
```
henryiii, line 412: I've tried to add this before but didn't have the `is_constructible` trick. Adding a description is not quite as important as it used to be now that it can be set afterward, but it is still ideal for consistency.

include/CLI/App.hpp
```diff
@@ -523,133 +536,157 @@ class App {
...
+        // take flag_description by const reference otherwise add_flag tries to assign to help_description
         if(help_ptr_ != nullptr) {
             remove_option(help_ptr_);
             help_ptr_ = nullptr;
         }
 
         // Empty name will simply remove the help flag
         if(!flag_name.empty()) {
-            help_ptr_ = add_flag(flag_name, std::move(help_description));
+            help_ptr_ = add_flag(flag_name, help_description);
```
henryiii, line 548: JC: Why are you removing the move?

include/CLI/App.hpp
```diff
@@ -523,133 +536,157 @@ class App {
...
 
-    /// Add option for flag
-    Option *add_flag(std::string flag_name, std::string flag_description = "") {
-        CLI::callback_t fun = [](CLI::results_t) { return true; };
-        Option *opt = add_option(flag_name, fun, flag_description, false);
-        if(opt->get_positional())
-            throw IncorrectConstruction::PositionalFlag(flag_name);
+  private:
+    /// internal function for adding a flag
+    Option *_add_flag_internal(std::string &flag_name, CLI::callback_t &fun, std::string &flag_description) {
```
henryiii: A few comments in this function describing what it does would be nice. Or at least a few empty lines to break it up visually. I like the unification here.

henryiii: Why non-const `std::string &` for `flag_description`?

include/CLI/App.hpp
```diff
@@ -1322,6 +1359,10 @@ class App {
         throw OptionNotFound(option_name);
     }
 
+    /// Shortcut bracket operator for getting results
+    const Option &operator[](std::string option_name) const { return *get_option(option_name); }
+    /// Shortcut bracket operator for getting results overload for const char *
+    const Option &operator[](const char *option_name) const { return *get_option(option_name); }
```
henryiii: Must have both, I guess? Disappointing that only `std::string` doesn't work.

examples/CMakeLists.txt
phlptp: Not in this PR, it was copied over from #229 in which is was different,  I will change it back in the next commit.

include/CLI/App.hpp
phlptp: it uses non-const reference to be able to move the name, description, and function and not copy them first.  We can revisit that but I wanted to avoid an extra copy if possible.

include/CLI/App.hpp
phlptp: for some reason it was routing to a standard operator when you used app["--option"]  I didn't quite follow what was going on with the warning.  But it didn't seem to work unless I directly overloaded to const char *.  I might try a few more things and see if const std::string & works or something like that.

include/CLI/App.hpp
henryiii: If this is something that a user types one by one, a computer is fast enough to make a copy. If it happens inside a loop, then avoiding copies is (more) important. Also, for things like names, I expect those are mostly short strings, and the short string optimization means moves are just as expensive as copies. Keeping the code simple and maintainable is also important.

include/CLI/App.hpp
phlptp: it uses a const std::string & now, so it makes sure to trigger the correct overload with the description and description has to be const otherwise it will try to write to it.    So you can't move a const std::string &

include/CLI/App.hpp
henryiii: I'm not shocked that it would require both, just saddened.

include/CLI/App.hpp
henryiii: Okay, makes sense.

phlptp:
TODOs
- [x] update documentation/ReadME
- [x] some code cleanup with in code documentation and better consistency with rest of code

phlptp:
For @henryiii 
regarding the operator[],  do you think it should return a const Option & or a const Option *?

And I made the operator[] function a template so there is only one in the code, but it may have the potential to generate a bunch of symbols if different string constants are used, so I may prefer the two overloads without the template, but I think either might be valid and have advantages/disadvantages, so i think it is your call?

henryiii:
I generally avoid returning references, because this is not safe:

```cpp
int& value(int& x) {
    return x;
}

int main() {
    int x = 3;
    auto y = value(x);
    y = 4;
    
    std::cout << x << " and " << y << std::endl;
}
```

You have to write `auto&` to get the correct behavior.

Now, disallowing copies helps somewhat, but it still might be more surprising to a novice programmer. Pointers are a bit more common and it's harder to make a mistake with auto.

README.md
```diff
@@ -192,45 +194,57 @@ app.add_flag(option_name,
...
-which allow a syntax for the option names to default particular options to a false value if some flags are passed.  For example:
+which allow a syntax for the option names to default particular options to a false value or any other value if some flags are passed.  For example:
 
 ```cpp
 app.add_flag("--flag,!--no-flag,result,"help for flag");`
 ``````
 
 specifies that if `--flag` is passed on the command line result will be true or contain a value of 1. If `--no-flag` is
-passed result will contain false or -1 if result is a signed integer type, or 0 if it is an unsigned type.  An
+passed the result will contain false or -1 if the result is a signed integer type, or 0 if it is an unsigned type.  An
```
```suggestion
passed `result` will contain false or -1 if `result` is a signed integer type, or 0 if it is an unsigned type.  An
```

README.md
```diff
@@ -258,6 +272,7 @@ Before parsing, you can set the following options:
 -   `->group(name)`: The help group to put the option in. No effect for positional options. Defaults to `"Options"`. `""` will not show up in the help print (hidden).
 -   `->ignore_case()`: Ignore the case on the command line (also works on subcommands, does not affect arguments).
 -   `->ignore_underscore()`: Ignore any underscores in the options names (also works on subcommands, does not affect arguments). For example "option_one" will match with "optionone".  This does not apply to short form options since they only have one character
+-   `->disable_flag_override()`:  from the command line long form flag option can be assigned a value on the command line using the `=` notation `--flag=value` If this behavior is not desired the `disable_flag_override()` disables it and will generate an exception if it is done on the command line.  The `=` does not work with short form flag options
```
```suggestion
-   `->disable_flag_override()`:  from the command line long form flag option can be assigned a value on the command line using the `=` notation `--flag=value`. If this behavior is not desired, the `disable_flag_override()` disables it and will generate an exception if it is done on the command line.  The `=` does not work with short form flag options.
```

README.md
```diff
@@ -317,10 +333,18 @@ You can access a vector of pointers to the parsed options in the original order
 If `--` is present in the command line that does not end an unlimited option, then
 everything after that is positional only.
 
+#### Getting results
+In most cases the easiest way is to return the results through a callback or variable specified in one of the `add_*` functions.  But there are situations where this is not possible or desired.  For these cases the results may be obtained through one of the following functions
```
```suggestion
In most cases the fastest and easiest way is to return the results through a callback or variable specified in one of the `add_*` functions.  But there are situations where this is not possible or desired.  For these cases the results may be obtained through one of the following functions:
```

README.md
```diff
@@ -317,10 +333,18 @@ You can access a vector of pointers to the parsed options in the original order
...
+
+- `results()`: retrieves a vector of strings with all the results in the order they were given.
+- `results(variable_to_bind_to)`: gets the results according to the MultiOptionPolicy and converts them just like the `add_option_function` with a variable.
```
henryiii, line 340: This and the ones below it will "reconvert" the variable from a string; this should not be done in performance critical code. A note of some sort might be useful here or nearby.

README.md
```diff
@@ -317,10 +333,18 @@ You can access a vector of pointers to the parsed options in the original order
...
+
 ### Subcommands
 
-Subcommands are supported, and can be nested infinitely. To add a subcommand, call the `add_subcommand` method with a name and an optional description. This gives a pointer to an `App` that behaves just like the main app, and can take options or further subcommands. Add `->ignore_case()` to a subcommand to allow any variation of caps to also be accepted. `->ignore_underscore()` is similar, but for underscores. Children inherit the current setting from the parent. You cannot add multiple matching subcommand names at the same level (including ignore
-case).
+Subcommands are supported, and can be nested infinitely. To add a subcommand, call the `add_subcommand` method with a name and an optional description. This gives a pointer to an `App` that behaves just like the main app, and can take options or further subcommands. Add `->ignore_case()` to a subcommand to allow any variation of caps to also be accepted. `->ignore_underscore()` is similar, but for underscores. Children inherit the current setting from the parent. You cannot add multiple matching subcommand names at the same level (including ignore_case and ignore_underscore).
```
```suggestion
Subcommands are supported, and can be nested infinitely. To add a subcommand, call the `add_subcommand` method with a name and an optional description. This gives a pointer to an `App` that behaves just like the main app, and can take options or further subcommands. Add `->ignore_case()` to a subcommand to allow any variation of caps to also be accepted. `->ignore_underscore()` is similar, but for underscores. Children inherit the current setting from the parent. You cannot add multiple matching subcommand names at the same level (including `ignore_case` and `ignore_underscore`).
```

README.md
```diff
@@ -353,7 +377,8 @@ There are several options that are supported on the main app and subcommands. Th
 -   `.got_subcommand(App_or_name)`: Check to see if a subcommand was received on the command line.
 -   `.get_subcommands(filter)`: The list of subcommands given on the command line.
 -   `.get_parent()`: Get the parent App or nullptr if called on master App.
--   `.get_option(name)`: Get an option pointer by option name
+-   `.get_option(name)`: Get an option pointer by option name will throw if the specified option is not available,  nameless subcommands are also searched
+-   `.get_option_no_throw(name)`: Gen an option pointer by option name, this function will return a `nullptr` instead of throwing if the option is not available.
```
```suggestion
-   `.get_option_no_throw(name)`: Get an option pointer by option name. This function will return a `nullptr` instead of throwing if the option is not available.
```

README.md
```diff
@@ -370,6 +395,7 @@ There are several options that are supported on the main app and subcommands. Th
 -   `.set_help_all_flag(name, message)`: Set the help all flag name and message, returns a pointer to the created option. Expands subcommands.
 -   `.failure_message(func)`: Set the failure message function. Two provided: `CLI::FailureMessage::help` and `CLI::FailureMessage::simple` (the default).
 -   `.group(name)`: Set a group name, defaults to `"Subcommands"`. Setting `""` will be hide the subcommand.
+- `[option_name]`: retrieve a const reference to an option given by `option_name` for Example `app["--flag1"]` will get an option reference for the "--flag1" value,  `app["--flag1"].as<bool>() will get the results of the command line for a flag
```
henryiii: If we make sure options cannot be copied (copy would need to be a protected function with some friends, I believe), this might be okay since it is intended to be chained. This should at least fail to compile:

```cpp
auto flag1 = app["--flag1"];
```

include/CLI/Error.hpp
```diff
@@ -231,6 +231,9 @@ class ArgumentMismatch : public ParseError {
     static ArgumentMismatch TypedAtLeast(std::string name, int num, std::string type) {
         return ArgumentMismatch(name + ": " + std::to_string(num) + " required " + type + " missing");
     }
+    static ArgumentMismatch FlagOverride(std::string name) {
+        return ArgumentMismatch(name + "was given a disallowed flag override");
```
henryiii:
```suggestion
        return ArgumentMismatch(name + " was given a disallowed flag override");
```

henryiii:
> The question is should there be a way to disallow the =45 as an override?

Just a thought; should flags that have defaults automatically get this disabled? That is,

```cpp
->add_flag("--flag,--no-flag{false}", ...);
```

Maybe should allow `--flag=false` but probably not `--no-flag=true`, which doesn't even do what you might read for it anyway.

Edit: note this is two suggestions, the one below is an extension of this idea (possibly not a good one).

---

In fact, it seems like that almost negates the need for the explicit setting; if a user did this:

```cpp
->add_flag("--flag{true},--no-flag{false}", ...);
```

That would mean no `=` syntax allowed. That would also indicate that all simple classic style options would support the = syntax by default, and the verbose syntax would be required instead of an inheritable setting like you have now, I'm not sure what I think of that.

README.md
phlptp: I moved it to a pointer,  copying an option would probably not be the intended result.

henryiii:
FYI, in the "files changed" tab you can queue up a list of accepted suggestions and commit them in one go. ;)

README.md
henryiii: Okay, I like that, and it is consistent. Probably should verify options can't be accidentally copied (not in this PR).

phlptp:
I will change [] to return by pointer.  

> > The question is should there be a way to disallow the =45 as an override?
> 
> Just a thought; should flags that have defaults automatically get this disabled? That is,
> 
> ```c++
> ->add_flag("--flag,--no-flag{false}", ...);
> ```
> 
> Maybe should allow `--flag=false` but probably not `--no-flag=true`, which doesn't even do what you might read for it anyway.
> 
> In fact, it seems like that almost negates the need for the explicit setting; if a user did this:
> 
> ```c++
> ->add_flag("--flag{true},--no-flag{false}", ...);
> ```
> 
> That would mean no `=` syntax allowed. That would also indicate that all simple classic style options would support the = syntax by default, and the verbose syntax would be required instead of an inheritable setting like you have now, I'm not sure what I think of that.

My hesitancy with that would be 
 ```c++
->add_flag("--flag,--no-flag{false}", ...);
```

you might still want to allow --flag=false, so I would a little hesitant to enable that by default with mixed usage,  though it might be advisable if all flags list defaults.  

Also even if the disable_flag_override is selected  if you specify ={} or =false  then there is no error.  

right now if you were to do `--no-flag=false` it would be the equivalent of `--flag=true`  you get the double negative thing going on.

**Comment by henryiii** [comment]
I haven't checked the code in detail, so not sure if it's even possible, but in the first half of that I was suggesting that

```
--flag,--no-flag{false}
```

Would produce a half-way state, where `--flag=false` would work, and `--no-flag=true` would fail. Setting the `disable_flag_override` can't affect an individual flag name, so that would put you into an all or nothing state.

**Comment by henryiii** [comment]
We can merge, play with it a while, then tweak if needed.

**Comment by phlptp** [comment]
> FYI, in the "files changed" tab you can queue up a list of accepted suggestions and commit them in one go. ;)

Did not know that,  Our code reviews haven't used suggestions that much,  maybe should start using that feature a little more

**Comment by phlptp** [comment]
do you have an opinion on the operator[] template,  whether it should remain a template or split into 2 non-template overloads?

**Comment by henryiii** [comment]
Since it's a simple function, I'd prefer avoiding the template (mildly). Simpler compile, doesn't create extra symbols for different string types. Slightly nicer error message too. Not a strong preference though. And on the other side, you only have to have one unit test for the template instead of two. If was any longer, I'd lean toward the template instead (or having one call the other; rather makes sense since you are working around a compiler "feature" for `operator[]`.

**Comment by henryiii** [comment]
> Did not know that

It's pretty new, it might still even list "experimental" when using it. It's quite handy though.

**Comment by phlptp** [comment]
I moved to two functions and collapsed all the commits.

I think I would lean away from ever putting on disable_flag_override by default.  One of the drivers was a use case to have a default output for flags for example
```
std::string file;
app.add_flag("--output{out.txt}",file,"enable file out and specify file");
```

In this case it is expected overrides would be fairly common and you wouldn't want to disable the overrides.

phlptp:
I think this is ready.

PR #239, opened 2019-02-24
add transformer and checkedTransformer

phlptp:
This is the last of the PR's originating from #229.  
This builds on the IsMember function and adds Transformer and CheckedTransformer validators  

Modifications include disabling the IsMember validator from using the second element of a map, so maps with non-convertible types can be used.  IsMember will modify the input to match an element of the set.

Validators added through the transform function are inserted at the beginning of the validators list, this allows transformations to occur before a check regardless of the order added.  

Some tweaks and updates to the IsMember function to clean up some loops and remove some unnecessary processing from inside the loop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,12 +31,17 @@ CLI11 is a command line parser for C++11 and beyond that provides a rich feature
 -   [Usage](#usage)
     -   [Adding options](#adding-options)
         -   [Option types](#option-types)
+        -   [Example](#example)
         -   [Option options](#option-options)
         -   [Validators](#validators)  🚧
+            -   [Transforming Validators](#transforming-validators)🚧
+            -   [Validator operations](#validator-operations)🚧
+            -   [Custom Validators](#custom-validators)🚧
+            -   [Querying Validators](#querying-validators)🚧
         -   [Getting Results](#getting-results) 🚧
     -   [Subcommands](#subcommands)
         -   [Subcommand options](#subcommand-options)
-        -   [Option Groups](#option-groups) 🚧
+        -   [Option groups](#option-groups) 🚧
     -   [Configuration file](#configuration-file)
     -   [Inheriting defaults](#inheriting-defaults)
     -   [Formatting](#formatting)
@@ -385,7 +390,7 @@ of `Transformer`:
 
 NOTES:  If the container used in `IsMember`, `Transformer`, or `CheckedTransformer` has a specialized search function like `std::unordered_map`  or `std::map` then that function is used to do the searching. If it does not have a find function a linear search is performed.  If there are filters present, the fast search is performed first, and if that fails a linear search with the filters on the key values is performed.
 
-##### Validator Operations🚧
+##### Validator operations🚧
 Validators are copyable and have a few operations that can be performed on them to alter settings.  Most of the built in Validators have a default description that is displayed in the help.  This can be altered via `.description(validator_description)`.
 the name of a Validator, which is useful for later reference from the `get_validator(name)` method of an `Option` can be set via `.name(validator_name)`
 The operation function of a Validator can be set via
@@ -500,7 +505,7 @@ There are several options that are supported on the main app and subcommands and
 
 > Note: if you have a fixed number of required positional options, that will match before subcommand names. `{}` is an empty filter function.
 
-#### Option Groups 🚧
+#### Option groups 🚧
 
 The method 
 ```cpp
--- a/README.md
+++ b/README.md
@@ -32,10 +32,11 @@ CLI11 is a command line parser for C++11 and beyond that provides a rich feature
     -   [Adding options](#adding-options)
         -   [Option types](#option-types)
         -   [Option options](#option-options)
+        -   [Validators](#validators)  🚧
         -   [Getting Results](#getting-results) 🚧
     -   [Subcommands](#subcommands)
         -   [Subcommand options](#subcommand-options)
-    -   [Option Groups](#option-groups) 🚧
+        -   [Option Groups](#option-groups) 🚧
     -   [Configuration file](#configuration-file)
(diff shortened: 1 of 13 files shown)


phlptp:
TODO yet:
- [x] A few additional tests
- [x] ReadMe and documentation update
- [x] fix the help printing with multiple Validators
- [x] some more refactoring and extracting functions

tests/TransformTest.cpp
```diff
@@ -0,0 +1,245 @@
...
+    args = {"-s", "val2"};
+    run();
+    EXPECT_EQ(value, test::val2);
+
+    args = {"-s", "val3"};
+    run();
+    EXPECT_EQ(value, test::val3);
+
+    args = {"-s", "val4"};
+    EXPECT_THROW(run(), CLI::ConversionError);
```
henryiii, line 56: Add `args = {"-s", "5"};` here too

henryiii:
Having three copies of nearly identical code bothers me, three times the maintenance and extra compile cost. Is there a way to combine and/or share bits? Having a way to "uncheck" a Validator would be useful (causing failures to be ignored). I don't think there's a good operator for it, but a method could be used maybe?

phlptp:
I was going to take a look at the code duplication once I got it working completely with all the tests.  I am pretty sure there are some things that can be done.

tests/TransformTest.cpp
phlptp: That won't produce any errors, since Transformer explicitly doesn't do any checking, and 5 is still a valid member of the underlying type.  Is the proposed check just to verify that?

henryiii: Yes. Tests for things that pass are good too, plus it's an example of what's different.

phlptp:
I tried extracting out a search function, I think It is working now.  I need to run through them some more to make sure they are actually working like I expect.  Also tried to remove some of the copying that was going on with the smart_deref and pair_adaptor.  Someone should really make sure those templates actually do what they are supposed to do.  This deep of template nesting is not something I do everyday.  An interesting exercise though.

include/CLI/Validators.hpp
```diff
@@ -32,37 +32,75 @@ class Option;
...
     std::function<std::string(std::string &)> func;
+    /// The name for search purposes of the Validator
+    std::string name_;
+    /// Enable for Validator to allow it to be disabled if need be
+    bool active_{true};
 
   public:
+    Validator() = default;
+    // Construct Validator from basic information
+    explicit Validator(std::function<std::string(std::string &)> op, std::string name = "", std::string desc = "") {
```
```suggestion
    explicit Validator(std::function<std::string(std::string &)> op, std::string name = "", std::string desc = "")
        : func(std::move(op)), name_(std::move(name)), description_(std::move(desc)): {}
```

(Note: needs to be formatted correctly) (this will make Codacity happy)

include/CLI/Validators.hpp
```diff
@@ -32,37 +32,75 @@ class Option;
...
   protected:
-    /// This is the type name, if empty the type name will not be changed
-    std::string tname;
+    /// This is the validator description
+    std::string description_;
 
-    /// This is the type function, if empty the tname will be used
-    std::function<std::string()> tname_function;
+    /// This is the description function, if empty the description_ will be used
+    std::function<std::string()> desc_function;
```
henryiii: Please use the convention that member variables end in underscore for `desc_function` and `func`.

include/CLI/Validators.hpp
```diff
@@ -32,37 +32,75 @@ class Option;
 
 ///
 class Validator {
-    friend Option;
-
   protected:
-    /// This is the type name, if empty the type name will not be changed
-    std::string tname;
+    /// This is the validator description
+    std::string description_;
```
henryiii: Could this be removed and all descriptions come from a function? (At one point I did different things based on whether it was a function or not, but I'm not sure you do)

include/CLI/Validators.hpp
phlptp: fixed

README.md
```diff
@@ -334,7 +316,110 @@ You can access a vector of pointers to the parsed options in the original order
 If `--` is present in the command line that does not end an unlimited option, then
 everything after that is positional only.
 
-#### Getting results 🚧
+#### Validators
```
henryiii, line 331: TOC needs change too?

README.md
```diff
@@ -334,7 +316,110 @@ You can access a vector of pointers to the parsed options in the original order
...
+Validators are structures to check or modify inputs, they can be used to verify that an input meets certain criteria or transform it into another value.  They are added through an options `check` or `transform` functions.  They differences between those two function are that checks do not modify the input whereas transforms can and are executed before any Validators added through `check`.
+
+CLI11 has several Validators built in that perform some common checks
+
+-   `CLI::IsMember(...)`: 🚧 Require an option be a member of a given set. See below for options.
+-   `CLI::Transformer(...)`: 🚧 Modify the input using a map. See below for options.
+-   `CLI::CheckedTransformer(...)`: 🚧 Modify the input using a map, and verify that the input is either in the set or already one of the outputs of the set. See below for options.
```
henryiii: What about the `->unchecked` idea?

README.md
phlptp: I think the Readme needs another pass especially with the option_group merge, so I can double check to make sure the TOC matches the headers,  Would you want all the subtitles in the TOC?

README.md
phlptp: There is an active(false) function on validators now.  That is the same concept,  It could be renamed as checked(), unchecked(), is_checked(), if you think that would read better and be more intuitive.

README.md
henryiii: Unless it makes it too long, I think it would be fine.

PR #240, opened 2019-02-26
Option delimiter

phlptp:
This PR addresses an issue I came across working on the Validators.  The problem was with vectors and Validators if you wanted a vector with a range of say [0, 20]  There are options for Validating individual elements but not after the conversion if you also wanted to use a custom delimiter.  Furthermore if you wanted to require 3 elements,  that had to be 3 elements before the delimiter was applied.  

What is comes down to is the delimiter was being applied too late, and had to be applied in a bunch of places if you wanted to use it.  

The solution was to move it as part of the Option class.  Any option can now have a delimiter, and the delimiter is applied when the results are loaded into an option.  This can allow options that require multiple arguments to accept them as a comma separated list or something along those lines and have that meet the requirement.  

One twist on the add_complex is you could add ->delimiter('+') and it would accept argument such as "3+2j"  as a single command line argument since the + on +2j is redundant.   a negative number would still need to be written like 4+-2j  for that to work.

My use case I wanted to able to specify a bunch of flags  like --flags=flag1,flag2,flag3,..  and have them checked against a hash table.  That would only work if the delimiter was applied before the Validator.

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,7 @@ Before parsing, you can set the following options:
 -   `->ignore_case()`: Ignore the case on the command line (also works on subcommands, does not affect arguments).
 -   `->ignore_underscore()`: Ignore any underscores in the options names (also works on subcommands, does not affect arguments). For example "option_one" will match with "optionone".  This does not apply to short form options since they only have one character
 -   `->disable_flag_override()`:  from the command line long form flag option can be assigned a value on the command line using the `=` notation `--flag=value`. If this behavior is not desired, the `disable_flag_override()` disables it and will generate an exception if it is done on the command line.  The `=` does not work with short form flag options.
+-    `->delimiter(char)`: allows specification of a custom delimiter for separating single arguments into vector arguments, for example specifying `->delimiter(',')` on an option would result in `--opt=1,2,3` producing 3 elements of a vector and the equivalent of --opt 1 2 3 assuming opt is a vector value
 -   `->description(str)`: Set/change the description.
 -   `->multi_option_policy(CLI::MultiOptionPolicy::Throw)`: Set the multi-option policy. Shortcuts available: `->take_last()`, `->take_first()`, and `->join()`. This will only affect options expecting 1 argument or bool flags (which do not inherit their default but always start with a specific policy).
 -   `->check(CLI::IsMember(...))`: Require an option be a member of a given set. See below for options.
@@ -338,9 +339,7 @@ In most cases the fastest and easiest way is to return the results through a cal
 
 - `results()`: retrieves a vector of strings with all the results in the order they were given.
 - `results(variable_to_bind_to)`: gets the results according to the MultiOptionPolicy and converts them just like the `add_option_function` with a variable.
-- `results(vector_type_variable,delimiter)`: gets the results to a vector type and uses a delimiter to further split the values
-- `Value=as<type>()`: returns the result or default value directly as the specified type if possible.
-- `Vector_value=as<type>(delimiter): same the results function with the delimiter but returns the value directly.  
+- `Value=as<type>()`: returns the result or default value directly as the specified type if possible, can be vector to return all results, and a non-vector to get the result according to the MultiOptionPolicy in place.
 
 ### Subcommands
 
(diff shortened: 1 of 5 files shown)


phlptp:
@henryiii  if you are ok with a few breaking changes regarding the delimiter argument with some of the options, I think you could eliminate a couple of the overloads for options that have the delimiter argument in them.  Since there is now a delimiter(char) function on the Option.

henryiii:
> @henryiii if you are ok with a few breaking changes regarding the delimiter argument with some of the options, I think you could eliminate a couple of the overloads for options that have the delimiter argument in them. Since there is now a delimiter(char) function on the Option.

I think that's better. And API that is not in a released version of CLI11 is fine to change! I strongly prefer Option methods, since they are easy to read, rather than positional arguments that do not describe what they do.

henryiii:
Let me know when you are ready for final review & merge. Do you want to drop the extra argument on `add_option` in this PR or should we do it later?

henryiii:
Please update the readme as well (optionally the changelog too, but that's completely optional, since it can cause conflicts when merging multiple PRs, and is pretty easy to do later).

include/CLI/App.hpp
```diff
@@ -435,66 +435,28 @@ class App {
...
     }
 
-    /// Add option for vectors (no default)
+    /// Add option for vectors
     template <typename T>
     Option *add_option(std::string option_name,
                        std::vector<T> &variable, ///< The variable vector to set
                        std::string option_description = "",
-                       char delimiter = '\0') {
+                       bool defaulted = false) {
```
henryiii: The reason for having two copies of the defaulted method was to allow types that could not be converted to strings (or should not be converted to strings, such as an uninitialized int) - if you don't provide the defaulted argument, CLI11 does not try to make a default at all. However, for vectors, it might be combinable (since the uninitialized part is not a problem), but you'd need some template trickery to only convert from a string if it is supported.

README.md
```diff
@@ -434,7 +433,7 @@ arguments, use `.config_to_str(default_also=false, prefix="", write_description=
 
 Many of the defaults for subcommands and even options are inherited from their creators. The inherited default values for subcommands are `allow_extras`, `prefix_command`, `ignore_case`, `ignore_underscore`, `fallthrough`, `group`, `footer`, and maximum number of required subcommands. The help flag existence, name, and description are inherited, as well.
 
-Options have defaults for `group`, `required`, `disable_flag_override`,`multi_option_policy`, `ignore_underscore`, and `ignore_case`. To set these defaults, you should set the `option_defaults()` object, for example:
+Options have defaults for `group`, `required`, `disable_flag_override`,`multi_option_policy`, `ignore_underscore`,`delimiter`, and `ignore_case`. To set these defaults, you should set the `option_defaults()` object, for example:
```
henryiii, line 436: Having the delimiter in the inheritable defaults is a great addition :)

include/CLI/App.hpp
phlptp: Should I add the overload back in?   I would argue if you are turning on defaulted and using an an uninitialized int that is a user issue.    But I could buy having a separate function that doesn't even have that code as you could have a variable type that doesn't have a << operator but does have a conversion from a string.

phlptp:
I think I will let you do the changelog, that is probably worth making sure the style is consistent.  I updated the readMe.    Let me know if I need to add the defaulted version as a separate overload for vectors.  If so it might be desirable to have a test that actually needs that and wouldn't compile without the overload.

include/CLI/App.hpp
henryiii: We should add it back in for now. I am thinking it might be possible to have a template function that checks for the ability to convert and does the right thing in both cases. Yes, turning on defaulted for an uninitialized value is a user problem, though you have to be careful not to capture it as a string regardless, unfortunately (otherwise we could have a `->default()` option turn it on later)

henryiii:
> If so it might be desirable to have a test that actually needs that and wouldn't compile without the overload.

That would be great, especially since I want to try adding a function to make the overload unneeded. But that's after this PR gets merged.

Edit: Added Issue #241.

include/CLI/App.hpp
henryiii: Is it valid to have a lambda function take an uninitialized value by capture? I think so, since we already do that to capture the normal value... That might actually mean `->default()` is possible...

phlptp: I added as as_string helper function in the transformer PR,  could convert to that for the default when/if that gets merged.  Might be possible to use some more template trickery to deal with some types that don't have a stream overload.

phlptp: I think capturing an uninitialized value would be undefined behavior if it is used before storing something to it.  Though for things like int and other trivial classes I think the effect is just a random non-nonsensical number being used, as long as it is not being used to index something else.

henryiii: > if it is used before

Yes, so:

```cpp
int x;
app.add_option("-x", x)->defaulted();
```

would be invalid (and hopefully might even be caught by the compiler), since you run the lambda function capturing the value of x when you run `defaulted()`. But this is a user issue and I think it would be okay. You could set the value of `x` in between the `add_option` and the `defaulted()`, and that would be fine. Suggestions for a better name (at least if this idea works) would be accepted...

henryiii:
Thanks!

rafiw:
Great work

henryiii:
What about `3 + 2j` and `"3 + 2j"`? Along with `-` not working by itself, I think this wouldn't work very well. (this still is fantastic, I expect using `;` and `:` as separators might be popular)

PR #246, opened 2019-03-01
Nameless sub subcommands

### LINKED ISSUE #245: Nested subcommands in nameless subcommands
Subcommands created within nameless subcommands are not found or processed.  They should ideally be treated the same as if they were in the main app as far as search and execution goes.  

This will also apply to option groups.

phlptp:
This PR will close  #245  
It makes the search for subcommands take into account subcommands that may be in nameless subcommands.  It applies it in `get_subcommand`, and `_valid_subcommands`

It does not apply it to get_subcommand_ptr,  I am not sure whether it should or not?

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1045,10 +1045,10 @@ class App {
 
     /// Check to see if a subcommand is part of this command (text version)
     App *get_subcommand(std::string subcom) const {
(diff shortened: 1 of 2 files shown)


henryiii:
I'm going to start working through and merging the PRs (mine needs some more work), then we can evaluate how everything fits together and see if we can fix some corner cases and simplify a bit.

PR #247, opened 2019-03-03
make duplicate subcommands work

phlptp:
This PR is a first cut at making duplicate subcommands work reasonably without adding any extra methods.  I added one test case with duplicates.  There probably should be a few more tests added about the interactions of duplicate subcommands and subcommands with duplicate names in different option groups.


henryiii:
This looks pretty good. I'd like to add the ability to run app callbacks multiple times in order, with the correct options run each time, but that's a more complex, trickier change.

henryiii:
I'll merge if you think it's ready.

phlptp:
There are some more issues with fallthrough and nested option groups that this doesn't deal with but that is somewhat more complex  and I am working on another change that will fix that.  But for the moment I think this is fine, and seems to be working.

PR #248, opened 2019-03-03
Fix Error when compiling with gcc with -Wparentheses-equality

rafiw:
Signed-off-by: Rafi Wiener <rafiw@mellanox.com>


henryiii:
Thanks! That was bugging me on Clang too.

PR #250, opened 2019-03-04
add read and write file validators #249

### LINKED ISSUE #249: ExistingFileValidator should also check for read permission
Hi @henryiii ,
ExistingFileValidator doesn't check if file is readable, e.g no permission.
I think it should be added as a default, instead of adding a new validator for this check.
Another option is adding a parameter to the ExistingFileValidator to check read write and not only if exists.
what do you think?

rafiw:
Will close #249.

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,6 +377,8 @@ CLI11 has several Validators built-in that perform some common checks
 -   `CLI::AsNumberWithUnit(...)`:🆕 Modify the `<NUMBER> <UNIT>` pair by matching the unit and multiplying the number by the corresponding factor. It can be used as a base for transformers, that accept things like size values (`1 KB`) or durations (`0.33 ms`).
 -   `CLI::AsSizeValue(...)`: 🆕 Convert inputs like `100b`, `42 KB`, `101 Mb`, `11 Mib` to absolute values. `KB` can be configured to be interpreted as 10^3 or 2^10.
 -   `CLI::ExistingFile`: Requires that the file exists if given.
+-   `CLI::ExistingReadFile`: Requires that the file given exists and have permission to read it.
+-   `CLI::ExistingWriteFile`: Requires that the file given exists and have permission to write to it.
 -   `CLI::ExistingDirectory`: Requires that the directory exists.
 -   `CLI::ExistingPath`: Requires that the path (file or directory) exists.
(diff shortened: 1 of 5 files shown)


rafiw:
@henryiii if you agree with this approach i will add tests

henryiii:
CLI::ExistingReadableFile And ...Writable... look like better names I think, but otherwise I like the way you’ve added it! +1

henryiii:
Will close #249.

rafiw:
@henryiii the permissions in windows is different then the linux once. that why the test fails.
How do you want to approach this?

cbachhuber:
What is the plan with this PR? Continue working on it or close it? I can also try to pick up the work

rafiw:
the issue is with windows since the permissions there is more complicated.
i can put the windows test under ifdef so it will pass.
is this ok?

phlptp:
This would need to be made to work with the changes from #341.  For C++17 that enables the use of filesystem so it might be a little easier in C++17 mode on windows now.

cbachhuber:
Right, but we would still need to enable file handling (and therefore, this feature) for windows under C++11, correct? Consequently, I would prefer to implement it for Windows & C++11 and not put the test under ifdef. What do you think?

rafiw:
i will take another look on this and add C++17 support as well

phlptp:
@henryiii this and a few other issues about validators indicate that some additional validators are of interest.  The balance here is that the Validators can take time to compile so for simple applications that don't use them that is a lot of non useful code that needs to compile that must be balanced with the high utility of the additional Validators for those that use them.  

What I am wondering if it makes sense to split the Validators into two pieces.  The basic definition is required for CLI11,  but many of the specific validators and all the machinery that goes with them could be split into a separate file that possibly is even not included in the single file header.  Then a separate Validators.hpp would define a library of  additional validators, that if someone wanted it they could include that header as well.    With the exception of the underlying filesystem operations none of the validators are required by the rest of CLI11.    So apart from the basic definition and maybe a couple of the most common validators, the library could be split without issue for 2.0.  Then we could get these validators in and some of the other issues without an increasing impact on the compile time.

phlptp:
Changes included and updated in #1203

Issue #166, opened 2018-09-27
select order of positional arguments based using arity as a mode switch?


KantarBruceAdams:
Hi,
    New user experimenting with CLI11 here.
I have an application which I want to parse arguments matching the following pattern:

 _app ?number1 number2? file1 ?file2?_

i.e. legal argument orders are:

**file1 
file1 file2
number1 number2 file1
number1 number2 file1 file2**

Basically I want the interpretation of arguments (or rather the mapping of variables) to vary with the number of arguments given (i.e. the arity) .
Is it possible to do this with CLI11?

I suspect the best we can do is with a mode switch like:

 _app --mode1 file1 ?file2?
 app --mode2 number1 number2 file1 ?file2?_

**mode1 file1 
mode1 file1 file2
mode2 number1 number2 file1
mode2 number1 number2 file1 file2**

or by making the numbers options instead of positional arguments. ie.

_?--numbers number1 number2? file1 ?file2?_

Where _--numbers_ is an option with a _type_size()_ of 2.

A related question. What happens with two or more optional positional arguments?
For example if I reordered the parameters to:

_file1 ?file2? ?number1 number2?_

There is an ambiguous parse as you can't distinguish between file2 and number1 except on type.

In a roll your own parser (which I would like to replace with something like CLI11) I would solve
the first case by selecting the mode (subcommand) based on the number of (positional) arguments.
I might resolve the second case by testing whether the argument was a number or not (though this has the big caveat of preventing numeric filenames).

Would there be any interest in supporting these behaviours (if they aren't already supported) or would they be considered of dubious value?

Before looking at CLI11 I was thinking along the lines of a parser with methods:
  
  _add_option()_ - like CLI11
  _add_variable(name,variable)_
  _set_positional_arguments(arity,args)_   
     - override positional arguments for a given arity - no args are optional - all are required.
 _set_positional_arguments(args)_ 
    - default positional argument order, optional args can be omitted

where args is a vector of argument names each mapping to the variable (& type) of the same name.

Would that fit in with the design of CLI11?

Related to this, I am curious about a couple of design choices. 
Typically with hand rolled parsers I have a loop to parse the options first and parse the positional arguments after the loop (using -- to mean end of options in the unix style). There is no such separation in the parse() method. Is this for efficiency? 
[...]

PR #251, opened 2019-03-04
Pre parse callback

phlptp:
This PR adds a pre parse callback and clarifies the callback timing of subcommands and option groups.  

It will also address issue #166.   The basic concept is to allow each app/subcommand/option_group to have two callbacks,  one would execute after the first argument  of the subcommand/option group is processed.  This allows an ability to effect the parsing dynamically, based on the number of arguments passed like Issue #166, or to dynamically enable or disable groups of options based on which options or subcommands were passed, or whatever other program state is appropriate to an application.    The other callback executes after parsing is complete.    With `immediate_callback` this is done immediately following the parse.    So callbacks from multiple subcommands can be executed in order and executed multiple times.    If that flag is not set all callbacks are executed on completion of all parsing in order of creation.  

To facilitate adaptive enabling or disabling option_groups some flags were created `enabled_by_default` and `disabled_by_default` that reset the disabled_ flag to a defined value on start of parsing.  

In the process of this a few issues with subcommand termination were fixed, and the readme updated with a new section on callbacks.  Also a subcommand_terminator was added (`++`)  to enable explicitly specifying that a subcommand is finished.   

Also added two new examples `shapes` with repeated subcommands and callbacks
and positional_arity which is Issue #166 coded up.
TODO: 
- [x] add missing tests for code coverage
- [x] add some helper functions to simplify triggered option groups

I am not sure I like all the names of the flags, so suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,7 @@ Option_group *app.add_option_group(name,description); // 🚧
 -app.add_mutable_set_ignore_case_underscore(... // 🆕 String only
 ```
 
-An option name must start with a alphabetic character, underscore, or a number🚧. For long options, after the first character '.', and '-' are also valid. For the `add_flag*` functions '{' has special meaning. Names are given as a comma separated string, with the dash or dashes. An option or flag can have as many names as you want, and afterward, using `count`, you can use any of the names, with dashes as needed, to count the options. One of the names is allowed to be given without proceeding dash(es); if present the option is a positional option, and that name will be used on help line for its positional form. If you want the default value to print in the help description, pass in `true` for the final parameter for `add_option`.
+An option name must start with a alphabetic character, underscore, a number 🚧,  '?'🚧, or '@'🚧. For long options, after the first character '.', and '-' are also valid characters. For the `add_flag*` functions '{' has special meaning. Names are given as a comma separated string, with the dash or dashes. An option or flag can have as many names as you want, and afterward, using `count`, you can use any of the names, with dashes as needed, to count the options. One of the names is allowed to be given without proceeding dash(es); if present the option is a positional option, and that name will be used on help line for its positional form. If you want the default value to print in the help description, pass in `true` for the final parameter for `add_option`.
 
 The `add_option_function<type>(...` function will typically require the template parameter be given unless a `std::function` object with an exact match is passed.  The type can be any type supported by the `add_option` function.
 
@@ -511,9 +511,10 @@ There are several options that are supported on the main app and subcommands and
 
 #### Callbacks 
 A subcommand has two optional callbacks that are executed at different stages of processing.  The `preparse_callback` 🚧 is executed once after the first argument of a subcommand or application is processed and gives an argument for the number of remaining arguments to process.  For the main app the first argument is considered the program name,  for subcommands the first argument is the subcommand name.  For Option groups and nameless subcommands the first argument is after the first argument or subcommand is processed from that group.
-The second callback is executed after parsing.  Depending on the status of the `immediate_callback` flag 🚧.  This is either immediately after the parsing of the subcommand.  Or if the flag is false, once after parsing of all arguments.  If the immediate_callback is set then the callback can be executed multiple times.   If the main app or subcommand has a config file, no data from the config file will be reflected in immediate_callback.
+The second callback is executed after parsing.  The behavior depends on the status of the `immediate_callback` flag 🚧. If true, this runs immediately after the parsing of the subcommand.  Or if the flag is false, once after parsing of all arguments.  If the `immediate_callback` is set then the callback can be executed multiple times if the subcommand list given multiple times.  If the main app or subcommand has a config file, no data from the config file will be reflected in immediate_callback. `immediate_callback()` has no effect on the main app, though it can be inherited.  For option_groups `immediate_callback` causes the callback to be run prior to other option groups and options in the main app, effectively giving the options in the group priority.  
 
 For example say an application was set up like
+
 ```cpp
 app.callback(ac);
 sub1=app.add_subcommand("sub1")->callback(c1)->preparse_callback(pc1)->immediate_callback();
@@ -523,6 +524,7 @@ app.preparse_callback( pa1);
 ... A bunch of other options
 
 ```
+
 Then the command line is given as
 
 ```
@@ -534,8 +536,8 @@ program --opt1 opt1_val  sub1 --sub1opt --sub1optb val sub2 --sub2opt sub1 --sub
 * c1 will be called when the `sub2` command is encountered
 * pc2 will be called with value of 6 after the sub2 command is encountered.
 * c1 will be called again after the second sub2 command is encountered
-* ac will be called after completing the parse
 * c2 will be called once after processing all arguments 
+* ac will be called after completing the parse and all lower level callbacks have been executed
 
 A subcommand is considered terminated when one of the following conditions are met.
 1. There are no more arguments to process
@@ -549,20 +551,26 @@ If the `immediate_callback` flag is set then all contained options are processed
 
 #### Option groups 🚧
 
-The subcommand method 
+The subcommand method
+
 ```cpp
 .add_option_group(name,description)
 ```
+
 Will create an option Group, and return a pointer to it.  An option group allows creation of a collection of options, similar to the groups function on options, but with additional controls and requirements.  They allow specific sets of options to be composed and controlled as a collective.  For an example see [range test](./tests/ranges.cpp).  Option groups are a specialization of an App so all [functions](#subcommand-options) that work with an App also work on option groups.  Options can be created as part of an option group using the add functions just like a subcommand, or previously created options can be added through
+
 ```cpp
 ogroup->add_option(option_pointer)
 ```
+
 ```cpp
 ogroup->add_options(option_pointer)
 ```
+
 ```cpp
 ogroup->add_options(option1,option2,option3,...)
 ```
+
 The option pointers used in this function must be options defined in the parent application of the option group otherwise an error will be generated.  
 Options in an option group are searched for a command line match after any options in the main app, so any positionals in the main app would be matched first.  So care must be taken to make sure of the order when using positional arguments and option groups.
 Option groups work well with `excludes` and `require_options` methods, as an Application will treat an option group as a single option for the purpose of counting and requirements.  Option groups allow specifying requirements such as requiring 1 of 3 options in one group and 1 of 3 options in a different group. Option groups can contain other groups as well.   Disabling an option group will turn off all options within the group.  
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -235,7 +236,7 @@ Option_group *app.add_option_group(name,description); // 🚧
...
+An option name must start with a alphabetic character, underscore, a number🚧, or one 🚧 of '$', '?', '#', '@', '%'. For long options, after the first character '.', and '-' are also valid characters. For the `add_flag*` functions '{' has special meaning. Names are given as a comma separated string, with the dash or dashes. An option or flag can have as many names as you want, and afterward, using `count`, you can use any of the names, with dashes as needed, to count the options. One of the names is allowed to be given without proceeding dash(es); if present the option is a positional option, and that name will be used on help line for its positional form. If you want the default value to print in the help description, pass in `true` for the final parameter for `add_option`.
```
henryiii: Why $? That's hard to add in most bash-like shells.

README.md
```diff
@@ -505,9 +508,48 @@ There are several options that are supported on the main app and subcommands and
...
+The second callback is executed after parsing.  Depending on the status of the `immediate_callback` flag 🚧.  This is either immediately after the parsing of the subcommand.  Or if the flag is false, once after parsing of all arguments.  If the immediate_callback is set then the callback can be executed multiple times.   If the main app or subcommand has a config file, no data from the config file will be reflected in immediate_callback. `immediate_callback()` has no effect on the main app.  For option_groups `immediate_callback` causes the callback to be run prior to other option groups and options in the main app, effectively giving the options in the group priority.
```
```suggestion
The second callback is executed after parsing.  The behavior depends on the status of the `immediate_callback` flag 🚧. If true, this runs immediately after the parsing of the subcommand.  Or if the flag is false, once after parsing of all arguments.  If the `immediate_callback` is set then the callback can be executed multiple times if the subcommand list given multiple times.  If the main app or subcommand has a config file, no data from the config file will be reflected in immediate_callback. `immediate_callback()` has no effect on the main app, though it can be inherited.  For option_groups `immediate_callback` causes the callback to be run prior to other option groups and options in the main app, effectively giving the options in the group priority.  
```

README.md
```diff
@@ -505,9 +508,48 @@ There are several options that are supported on the main app and subcommands and
...
+For example say an application was set up like
+```cpp
+app.callback(ac);
+sub1=app.add_subcommand("sub1")->callback(c1)->preparse_callback(pc1)->immediate_callback();
+sub2=app.add_subcommand("sub2")->callback(c2)->preparse_callback(pc2);
+app.preparse_callback( pa1);
+
+... A bunch of other options
+
+```
```
henryiii, line 529: Remove ending blank line inside the code fence, add blank lines around all code fences (some markdown parsers will not parse correctly if a code fence is not surrounded by whitespace).

README.md
```diff
@@ -505,9 +508,48 @@ There are several options that are supported on the main app and subcommands and
...
+
+* pa will be called prior to parsing any values with an argument of 14.
+* pc1 will be called immediately after processing the sub1 command with a value of 10.
+* c1 will be called when the `sub2` command is encountered
+* pc2 will be called with value of 6 after the sub2 command is encountered.
+* c1 will be called again after the second sub2 command is encountered
+* c2 will be called once after processing all arguments 
+* ac will be called after completing the parse and all lower level callbacks have been executed
+
+A subcommand is considered terminated when one of the following conditions are met.
```
henryiii, line 545: Add a blank line before starting (or ending) a list. Some markdown parsers need space.

include/CLI/App.hpp
```diff
@@ -90,6 +90,16 @@ class App {
     /// If set to true the subcommand is disabled and cannot be used, ignored for main app
     bool disabled_{false};
 
+    /// Flag indicating that the pre_parse_callback has been triggered
+    bool pre_parse_called_{false};
+
+    /// Flag indicating that the callback for the subcommand should be executed immediately on parse completion which is
+    /// before help or ini files are processed.
```
henryiii: This should be inheritable.

include/CLI/App.hpp
```diff
@@ -292,6 +314,25 @@ class App {
         return this;
     }
 
+    /// Set the subcommand to be disabled by default, so on clear(), at the start of each parse it is disabled
+    App *disabled_by_default(bool disable = true) {
+        disabled_by_default_ = disable;
+        return this;
+    }
+
+    /// Set the subcommand to be enabled by default, so on clear(), at the start of each parse it is enabled(not
```
```suggestion
    /// Set the subcommand to be enabled by default, so on clear(), at the start of each parse it is enabled (not
```

include/CLI/App.hpp
```diff
@@ -1720,44 +1778,58 @@ class App {
...
-        if(_valid_subcommand(current))
+        if(_valid_subcommand(current, ignore_used_subcommands))
             return detail::Classifier::SUBCOMMAND;
         if(detail::split_long(current, dummy1, dummy2))
             return detail::Classifier::LONG;
         if(detail::split_short(current, dummy1, dummy2))
             return detail::Classifier::SHORT;
         if((allow_windows_style_options_) && (detail::split_windows_style(current, dummy1, dummy2)))
             return detail::Classifier::WINDOWS;
+        if((current == "++") && (!name_.empty()))
```
```suggestion
        if((current == "++") && !name_.empty())
```

(Make it slightly easier to parse by eye)

include/CLI/App.hpp
```diff
@@ -1811,20 +1883,34 @@ class App {
         }
 
         for(App_p &sub : subcommands_) {
-            sub->_process_env();
+            if((sub->get_name().empty()) || (!sub->immediate_callback_))
```
```suggestion
            if(sub->get_name().empty() || !sub->immediate_callback_)
```

include/CLI/App.hpp
```diff
@@ -1811,20 +1883,34 @@ class App {
...
+                sub->_process_env();
         }
     }
 
     /// Process callbacks. Runs on *all* subcommands.
     void _process_callbacks() {
+
+        for(App_p &sub : subcommands_) {
+            // process the priority option_groups first
+            if((sub->get_name().empty()) && (sub->immediate_callback_)) {
```
```suggestion
            if(sub->get_name().empty() && sub->immediate_callback_) {
```

include/CLI/App.hpp
```diff
@@ -1952,7 +2038,10 @@ class App {
                     }
                 }
             }
-            sub->_process_requirements();
+            if((sub->count() > 0) || (sub->name_.empty())) {
+                sub->_process_requirements();
+            }
+
             if((sub->required_) && (sub->count_all() == 0)) {
```
henryiii: (I think it's better not to add () if they are not needed, but it's done on this way on the line below that was not changed by this PR, so it's okay to leave them for now if you prefer. I might try to clean them up one day.

include/CLI/App.hpp
```diff
@@ -2116,52 +2234,59 @@ class App {
...
+                if(subc->_parse_positional(args)) {
+                    if(!subc->pre_parse_called_) {
+                        subc->_trigger_pre_parse(args.size());
+                    }
+                    return true;
                 }
             }
         }
-
+        /// let the parent deal with it if possible
```
henryiii: Stylistic note: Comments inline are two `//`, three slashes indicated that Doxygen picks up the comment for the API docs, and it can't pick up comments in code.

include/CLI/App.hpp
```diff
@@ -2408,6 +2573,47 @@ class Option_group : public App {
         add_options(args...);
     }
 };
+/// Helper function to enable one option group/subcommand when another is used
```
henryiii, line 2638: It's hard to tell here, but these should be in namespace detail, correct? (Or a nameless namespace)

README.md
henryiii: It might be okay to support $, it just seems a bit odd. I know that `-?` can be useful.

henryiii:
Names of the flags, etc. seem fine, at least to me.

README.md
phlptp: /? is pretty standard for help in windows applications, and as I am converting other applications most of the time I just support help with -? and that failed the first time I tried it. So that is the most useful one.  The others I used -@ once as a shortcut for --address.  I guess I can see `$` being potentially problematic in bash.  If you wanted to remove all but '?'  that is probably fine.

henryiii: I'd like to drop `$` to avoid issues with bash shells (adding a `--$this` option would not behave as expected), similarly `#` and `%` (comments in bash, Variables in windows) `?` is safe, as is `@` I think. However, the code could support `$#%` still; I just don't think they should be advertised. The only issue is if you want to reserve these as shortcuts, like the `!{}` chars.

include/CLI/App.hpp
phlptp: I wasn't entirely sure where to put these.  They are intended to used by users similarly to the Validators.  They could be member functions of app.  They could be classes on their own.    I wanted the functionality but I am not sure of the name or location.  Basically a way to turn on/off option groups in a sequence during a parse.

henryiii: Okay, they have a slightly different use than I thought.

include/CLI/App.hpp
```diff
@@ -1720,44 +1778,58 @@ class App {
             app->_configure();
         }
     }
-    /// Internal function to run (App) callback, top down
+    /// Internal function to run (App) callback, bottom up
```
phlptp, line 1820: This is a change in behavior,  the order of callbacks between the main app and child subcommands.  Previously it was a top down approach.  Now it is a bottom up approach.  There wasn't any tests that relied on this behavior but it is a change.   I would argue it makes more sense as a bottom up approach, to know that the main app callback only executes after everything else has completed.  But being there was a comment on the effect I wasn't entirely sure if there was a reason behind it or something that depended on that behavior.    I would say in most applications it would make no difference, but this is a change in behavior.

include/CLI/App.hpp
phlptp: I cleaned up a few of them as I was looking through other suggestions, I am sure there are more.

phlptp:
As a note: a number of the issues with callback ordering came up as I am trying to integrate CLI11 into one of our bigger applications.  A lot of the capabilities I have worked on over the last few weeks have been driven by the requirements and use cases from those applications, so I think I am nearing the end of integrating it, but we will find out over the next few days as I continue working on that.

include/CLI/App.hpp
henryiii: Would a user be able to do something like:

```bash
git -C dir add -u 
```

That is, add would need to change to the `dir` directory. It's not too important, for this example or the existing code, since you don't have to run this in callbacks; you can just do a complete parse then inspect the results, which I assume is a bit more common.

henryiii:
We'll wait to do a 1.8 release at least until you have finished integration and verified everything works together. :)

include/CLI/App.hpp
henryiii: PS: I'm okay with the change, if I didn't say that clearly at the end (I didn't)

phlptp:
I updated based on #252

henryiii:
Is this ready for review/merge?

phlptp:
I have been slowly integrating CLI11 into our bigger applications, and am probably about half way through that effort.  I think with what I did yesterday, I need to add a remove_subcommand function to app and add_subcommand(sub *) to the option_group class.  But I haven't done that yet or written test cases for them.  

So I think you can merge/review as is and I will add the rest and what else I might uncover in another PR or you can wait a few days while I complete the integration.

phlptp:
Will close #256

phlptp:
I think this is ready to go, for another look.

henryiii:
Okay, will look at it as soon as I can. Traveling back from a conference in Switzerland tomorrow, then to JLab this weekend for another one.

henryiii:
Looks pretty good, will merge. Standard feedback that I might suggest adjustments later before release.

Issue #194, opened 2019-01-17
Improve compile time


ncihnegn:
For the same options, CLI11 triples my compile time of boost::program_options.

henryiii:
`boost::program_options` is not a header-only library - it is pre-compiled. The benefit of not having to pre-compile the library comes at a compile time cost, as with all header only libraries. However, you can mitigate that by making sure the CLI11 code only is in one cpp file, and keeping your components separate. There's no need to include CLI11 in any other file - you can just use plain C++ types! That's part of the design. PyBind11 has some excellent examples of how to keep headers leaking into other files, for maximum compile-time performance.

However, CLI11 still only takes a couple of seconds to compile, so do you have a very simple application? It should not grow as your application becomes more complex.

henryiii:
Technically, CLI11's current design would allow it to have an optional precompiled mode - I think a (very) few other libraries support a dual usage mode. However, it may become more heavily templated in 2.0, with `Option<T>` instead of the current `Option`. If that happens, it will need to remain header only (since templates cannot be precompiled).

henryiii:
To support precompilation, all headers would need to be split into two, a declaration and an implementation, then there would be a way to precompile the implementation files optionally.

Personally, though, if I have to do it, I'll wait till both the redesign, and until C++ Modules become finalized, probably C++20, hopefully not C++23.

ncihnegn:
I agree. But it will be nice if there is a macro like FAST_COMPILE to exclude some codes.

henryiii:
I wonder if something like this would help: <https://quuxplusone.github.io/blog/2019/01/06/hyper-function/> - CLI11 uses `std::function` everywhere. Would have to be re-implemented or permission to use it in a BSD software would be needed, but would be interesting to test.

There really aren't many optional codes in CLI11. The only one is Timer, and you don't get it unless you include it. In general, you should only includes CLI11 is a limited set of files, probably only your main file.

henryiii:
I'm planning to move the extra transforms and any other less heavily used optional portions to a separate file, that will close this issue when done.

jlisee:
[...]

PR #253, opened 2019-03-11
Support for size values (640 KB) and numbers with unit in general

metopa:
This PR adds infrastructure for accepting inputs in form `<number> <unit>`. This covers things like size inputs, durations, and many more. `<suffix>` may or may not be separated from the number with space, must contain only letters and optionally can be matched case-insensitively.

`CLI::AsNumberWithUnit` takes a mapping from unit to factor. Transformation for durations may look like this:
```
map<string, double> m = {{"ns", 1e-9}, {"us", 1e-6}, {"ms", 1e-3}, {"s", 1}, {"m", 60}};
app.add_option("-d", duration)->transform(CLI::AsNumberWithUnit(mapping));
```
Transformation for size is already provided:
```
app.add_option("-s", size)->transform(CLI::AsSizeValue(true));
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ The `add_option_function<type>(...` function will typically require the template
 🚧 Flag options specified through the `add_flag*` functions allow a syntax for the option names to default particular options to a false value or any other value if some flags are passed.  For example:
 
 ```cpp
-app.add_flag("--flag,!--no-flag,result,"help for flag"); // 🚧
+app.add_flag("--flag,!--no-flag",result,"help for flag"); // 🚧
 ```
 
 specifies that if `--flag` is passed on the command line result will be true or contain a value of 1. If `--no-flag` is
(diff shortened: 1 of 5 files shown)


henryiii:
I like the general version and adding an explicit one for common uses. I will probably break out the more specialized transforms (at least the size one here, IP4, etc.) into a separate utilities file so that single header file users won't have a growing file each release. It would be available for either normal or single header file use, but a separate file. Just a heads up for my plans, does not affect this PR. Will hopefully help with #194.

metopa:
@henryiii Could you please take a look at the building error on GCC 4.7? I don't have it on my machine and the error is a bit of WTF

README.md
```diff
@@ -242,7 +242,7 @@ The `add_option_function<type>(...` function will typically require the template
 🚧 Flag options specified through the `add_flag*` functions allow a syntax for the option names to default particular options to a false value or any other value if some flags are passed.  For example:
 
 ```cpp
-app.add_flag("--flag,!--no-flag,result,"help for flag"); // 🚧
+app.add_flag("--flag,!--no-flag",result,"help for flag"); // 🚧
```
henryiii, line 246: 😳

henryiii:
I'll try it out in a 4.7 docker container later today and see if I can see what it's complaining about. GCC 4.7 is a pain, I'll probably be dropping it in a release or two. This looks fixable without too much work, I think.

README.md
metopa: It happens :)

henryiii:
Note: This is on my todo list but I'm traveling a bit unexpectedly to cover for someone at a conference, so eventually will get to this!

metopa:
@henryiii Sure thing, thank you a lot!

henryiii:
I believe you are seeing this bug in GCC: https://stackoverflow.com/a/12086873/2402816. I believe you are only putting these methods in the class for organization? There are no members. We can either move them to put this in 1.8 or wait till 1.9 and drop GCC 4.7 support.

metopa:
Tricky! Thank you for figuring this out.

metopa:
After returning back to this after some time I got an idea. Should I rename `SuffixedNumber` to smth like `NumberWithUnit` and generally use "unit" term in docs? Right now it makes more sense to me.

henryiii:
@phlptp , what do you think on the naming question? I like both. My physics backround leans toward Unit, but Suffix is a little more general

phlptp:
I think I would have  a preference toward "Unit".  Given a significant chunk of my time recently has been on a runtime units library I may be bit biased though.

henryiii:
@metopa Do you want to wait and target 1.9 instead, or try to get this in to 1.8?

metopa:
@henryiii I've just updated it

henryiii (approved):

henryiii:
Thanks!

PR #254, opened 2019-03-12
Make ConfigVersion arch independent

pleroux0:
CLI11 is header only and the architecture shouldn't matter. The arch
dependence has the potential to cause find_package to fail incorrectly.


henryiii:
Thanks!

PR #255, opened 2019-03-14
Resolve two switch related warnings on GCC 8

andrew-hardin:
These two commits resolve one `switch-default` warning and three `switch-enum` warnings raised under GCC 8.


henryiii:
We will need to turn off coverage for these lines.

henryiii:
Thanks!

PR #262, opened 2019-03-30
Positional argument checks

phlptp:
If merged this PR will add an option to require positional arguments to pass validation before matching.    This would allow creating positionals for a file and number and allowing them to be entered in any order on the command line.  

Modifications include adding `validate_positionals` function on App
moving the `_validate` code into a separate function in options.
removing the `set_results` function in options which was being used by config file functions but it was bypassing some of the result processing, which could cause some issues if the delimiter option was also being used.  

Also tweaked the template for add_option to separate out the lexical_cast type from the assignment type if desired.  (Probably need a few more tests of this yet).


henryiii:
@phlptp, can you fix the conflict? I’ll be back from yet another conference on Tuesday.

Issue #264, opened 2019-04-17
app.remaining() can't be passed directly into parse


phlptp:
When using CLI11 I have sometimes found it necessary to cascade the options using allow_extras() or prefix_command().  I then want to pass through the options into a different CLI app.   

The first one would usually be an argc, argv input, but the subsequent ones would use the vector form of parse.    But when used the arguments returned via remaining are not in the order expected by a further parse, resulting in incorrect parse when used if any of the arguments had values, without first reversing.  which is somewhat annoying to remember each time it is needed.  

I am proposing an additional function to get the remaining arguments in a form expected by parse.    I will submit a PR with a draft of this change.

phlptp:
Issue closed with #265

PR #265, opened 2019-04-17
add a function to get the remaining arguments in a valid order

phlptp:
See issue #264 

This adds function remaining_for_passthrough()  --> different name suggestions are welcome!

It also adds some overloads with rvalue references so parse can take the function directly when used in a cascade fashion, and not do extra work or generate warnings when used.    The parse from string and (argc, argv) also use this overload since the args can't be used any further.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,7 @@ On the command line, options can be given as:
 
 Extra positional arguments will cause the program to exit, so at least one positional option with a vector is recommended if you want to allow extraneous arguments.
 If you set `.allow_extras()` on the main `App`, you will not get an error. You can access the missing options using `remaining` (if you have subcommands, `app.remaining(true)` will get all remaining options, subcommands included).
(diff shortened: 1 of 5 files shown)


phlptp:
TODO:
- [x] Documentation
- [x] check coverage
- [x] add Example

phlptp:
This also fixed a potential bug if the main app had a name and a subcommand terminator was used.

henryiii:
Ready for review?

phlptp:
yes

henryiii (approved):
I think the name is okay, it's at least easy to find if one looks for it. I would be tempted to suggest some sort of templating on the parse function, but this is better for something like the worked planned for #194 .

phlptp:
I thought about making it a template, I also thought about making the parse take a const ref to a vector but then forcing the one that doesn't modify becomes a little more awkward.    So I wasn't entirely clear what the best method of dealing with this was.    In any case the main reason was so you could call parse(app.remaining_for_passthough()) without the compiler getting mad at you.

PR #266, opened 2019-04-18
fix tests on aarch64

ryan4729:
This fixes `HelpersTest` on aarch64. The test assumed that the char type is always signed, but on aarch64 it's unsigned by default.


henryiii:
Great, thanks!

PR #269, opened 2019-04-25
Alignment in multiline description

elszon:
Hi, 
I want to propose formatting for multiline description in help.
Next line of description should be alignment to first.

Regards,
elszon


henryiii:
This looks fine to me.

henryiii:
Thanks!

Issue #273, opened 2019-05-07
How should we disable building tests for CLI11 when added as subproject?


kcsaul:
Hi,

Having found/added CLI11 to my project previously (very pleased to find it!), I've recently upgraded to the latest version.

Rather than copying files or using more complicated commands for including external projects, I prefer to add external projects as a git submodule with cmake add_subdirectory to include them in the overall project build.

After upgrading to the latest version, I'm now getting build failures because the googletest library has already been defined when googletest was included by CLI11 (before being included directly for my project tests).

It would seem this has come about due to #183 forcing the CLI11_TESTING option based on the global BUILD_TESTING option which is defined when including CTest (something I do before including external projects so option can be used to suppress including googletest in the build).

Based on comments in a [similar issue for a different project](https://github.com/google/re2/issues/138), it would seem the only way of disabling tests for a subproject is it uses a specific option (like CLI11_TESTING), as there's no way of overriding/presetting the global BUILD_TESTING option for a subproject, without disabling tests for main project.

I note there's also comments in #183 which indicate CLI11_TESTING will be deprecated/removed in a future release.

Unless there's another way of achieving it(?), I'd prefer the use of CLI11_TESTING be kept, perhaps defaulting based on BUILD_TESTING (if not already preset before including CLI11).

Thanks,
Kevin

henryiii:
I've been working on ideas related to this. Because `BUILD_TESTING` is a global value, it clashes easily. I think I had a solution: only add testing if this is the main project *and* BUILD_TESTING is defined, or if CLI11_TESTING is defined. You can see the discussion here: https://gitlab.com/CLIUtils/modern-cmake/issues/7 . 

This need to be addressed before the 1.8 release.

phlptp:
I have found it is handy in subprojects and libraries that are used as components to have some ability to turn off testing even if they are not directly a subproject.  So I would recommend having CLI11_TESTING be checked as well if it is defined.  So setting CLI11_TESTING to OFF would turn off the build of all the testing regardless of other variables and project scope.

kcsaul:
Here's a variation on the initial solution to consider that may be a little more flexible for those who want to explicitly enable/disable CLI11 testing.

- Default CLI11_TESTING based on whether it's the main project, provided it's not been defined/cached already.

- Before enabling/including tests, check both CLI11_TESTING and BUILD_TESTING (if defined).

henryiii:
[...]

PR #277, opened 2019-05-11
Build testing

henryiii:
The original implementation (and many implementations of `BUILD_TESTING`, I think) has an issue; if you set `BUILD_TESTING` in a parent project, it gets picked up unconditionally in a child project. This makes `BUILD_TESTING` be ignored if CLI11 is a child project; `CLI11_TESTING` will still control it if a parent project really wants a child project to build its tests. If you set `BUILD_TESTING` with CLI11 as the parent project, this will override `CLI11_TESTING`, as it does now.

Based on plans mentioned in #273, follows up on #183. This *may* be the last change for 1.8.

Also adds clang-format testing on Azure, along with GCC 4.7, 9, Clang 3.4, 8. You can now force libc++ if this is the master project.

Drops experimental optional auto-support, it can be fragile.

|Main Project | CLI11_TESTING unset | CLI11_TESTING OFF | CLI11_TESTING ON |
|-------------|----------------------|----------------------|--------------------|
|BUILD_TESTING default |       ✔️        |                  ❌              |                ✔️             |
|BUILD_TESTING OFF |          ❌       |                 ❌                |              ✔️              |
|BUILD_TESTING ON |            ✔️       |                  ❌               |             ✔️                |

|Sub Project  | CLI11_TESTING unset | CLI11_TESTING OFF | CLI11_TESTING ON |
|-------------|----------------------|----------------------|--------------------|
|BUILD_TESTING unset |       ❌        |                  ❌              |                ✔️             |
|BUILD_TESTING OFF |          ❌       |                 ❌                |              ✔️              |
|BUILD_TESTING ON |           ❌       |                  ❌              |              ✔️                |


CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ Set handling has been completely replaced by a new backend that works as a Valid
 * Cleanup for shadow warnings [#232]
 * Better alignment on multiline descriptions [#269]
 * Better support for aarch64 [#266]
+* Respect `BUILD_TESTING` only if CLI11 is the main project; otherwise, `CLI11_TESTING` must be used [#277]
+* Drop auto-detection of experimental optional; must be enabled explicitly (too fragile) [#277]
```
phlptp, line 25: We actually ran into some difficulty with this so this is good!

CMakeLists.txt
```diff
@@ -149,14 +158,25 @@ if(CLI11_SINGLE_FILE)
...
+cmake_dependent_option(CLI11_TESTING
+    "Build the tests and add them (BUILD_TESTING also works if this is the main project)"
+    ON
+    "CUR_PROJ"
+    OFF)
+
+if(DEFINED BUILD_TESTING AND CUR_PROJ)
+    set(CLI11_TESTING_INTERNAL ON)
+else()
+    set(CLI11_TESTING_INTERNAL "${CLI11_TESTING}")
```
phlptp, line 170: These conditions don't seem quite right.  
If BUILD_TESTING is defined to OFF in a main project then testing is turned ON regardless, and I don't necessarily see a way to turn if off, and if BUILD_TESTING is defined and it is a main project then CLI11_TESTING is ignored completely.

azure-pipelines.yml
```diff
@@ -6,21 +6,68 @@
...
+
 jobs:
-- job: Linux
+
+- job: ClangFormat
   pool:
     vmImage: 'ubuntu-16.04'
   steps:
-    - template: .ci/azure-steps.yml
+    - script: scripts/check_style_docker.sh
```
phlptp: Will the other format check be dropped.  Seems a little redundant to have both, or are they checking different things?

CMakeLists.txt
henryiii: Thanks, I think that was a copy-and-paste mistake. Maybe defining both CLI11_TESTING and BUILD_TESTING in a main project should be an error? (At least if they differ)

azure-pipelines.yml
henryiii: I'll be slowly reducing our usage of non-Azure CI, so yes. I could probably go ahead and drop that in this PR.

CMakeLists.txt
phlptp: I think my preference would be an AND condition at least for the case of a main project.

henryiii: New version: `CLI11_TESTING` will be respected if it is defined in all cases, otherwise `BUILD_TESTING` is defined and set to ON if this is a main project. I think that's probably the best way to go, especially since I'm probably going to recommend `BUILD_TESTING` + main project only for Modern CMake.

azure-pipelines.yml
henryiii: I'm actually still only checking "clang-tidy" on Travis. I'll need to move that in here in the future.

phlptp (approved):

PR #279, opened 2019-05-16
Boost optional is no longer automatic

henryiii:
Checking to see if this faster. Drops the "automatic detection" of Boost for Boost::Optional.


henryiii:
I think the main fix here is avoiding an unlimited number of parallel build jobs - for some reason Linux native really cared. But including boost::optional headers is a bit heavy and probably should only be explicit. std::optional is in the stdlib, so is something I feel okay including automatically if found.

henryiii:
Build times are now down from 10-25 minutes to something like 5 for Linux native build.

PR #280, opened 2019-05-18
Version 1.8

### LINKED ISSUE #270: Plan for CLI11 1.8
Here's my plan for 1.8, and a rough idea for the future.

The last major feature for 1.8 (#242) was merged. In the next week or so, I'd like to do a little polishing (feel free to pitch in and test/suggest/help). I'm hoping to look at #250 and #253 in that time. Possibly also add a move constructor to simplify a common use case #268 (if it does not appear to be safe, I'll be happy to wait till 1.9 to decide, though). I'll probably try to make an official beta release (pretty much consider it in beta now if you'd like to test it!), and a full release at some point not long after.

For 1.9, I'd like to work on a method to allow part of the library to be precompiled to address #194. I'll also be removing and adding some warnings for the deprecated bits of the library, preparing for a slightly bigger removal in 2.0. I'm also highly considering dropping support for GCC 4.7 in one of those versions; please speak up if you need it!

Other general tasks include moving more of the CI over to Azure (I plan to release a post on the process on [iscinumpy.gitlab.io](https://iscinumpy.gitlab.io), too), moving the tutorial so it shares the same repository (I love gitlab.io, but it will be easier to keep in sync with the code and additions - and maybe CLI11 will be synced there too!)

henryiii:
1.8 will be released when this is merged. Closes #270.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,15 +33,15 @@ CLI11 is a command line parser for C++11 and beyond that provides a rich feature
         -   [Option types](#option-types)
         -   [Example](#example)
         -   [Option options](#option-options)
-        -   [Validators](#validators) 🚧
-            -   [Transforming Validators](#transforming-validators) 🚧
-            -   [Validator operations](#validator-operations) 🚧
-            -   [Custom Validators](#custom-validators) 🚧
-            -   [Querying Validators](#querying-validators) 🚧
-        -   [Getting Results](#getting-results) 🚧
+        -   [Validators](#validators) 🆕
+            -   [Transforming Validators](#transforming-validators) 🆕
+            -   [Validator operations](#validator-operations) 🆕
+            -   [Custom Validators](#custom-validators) 🆕
+            -   [Querying Validators](#querying-validators) 🆕
+        -   [Getting Results](#getting-results) 🆕
     -   [Subcommands](#subcommands)
         -   [Subcommand options](#subcommand-options)
-        -   [Option groups](#option-groups) 🚧
+        -   [Option groups](#option-groups) 🆕
         -   [Callbacks](#callbacks)
     -   [Configuration file](#configuration-file)
     -   [Inheriting defaults](#inheriting-defaults)
@@ -55,7 +55,7 @@ CLI11 is a command line parser for C++11 and beyond that provides a rich feature
 -   [Contribute](#contribute)
 -   [License](#license)
 
-Features that were added in the last released major version are marked with "🆕". Features only available in master are marked with "🚧".
+Features that were added in the last released major version are marked with "🆕". Features only available in master are marked with "🆕".
 
 ## Background
 
@@ -191,14 +191,14 @@ While all options internally are the same type, there are several ways to add an
 
 ```cpp
 // Add options
-app.add_option(option_name, help_str="") // 🚧
+app.add_option(option_name, help_str="") // 🆕
 
 app.add_option(option_name,
-               variable_to_bind_to, // bool, int, float, vector, 🚧 enum, or string-like, or anything with a defined conversion from a string
+               variable_to_bind_to, // bool, int, float, vector, 🆕 enum, or string-like, or anything with a defined conversion from a string
                help_string="")
 
 app.add_option_function<type>(option_name,
-               function <void(const type &value)>, // 🚧 int, bool, float, enum, vector, or string-like, or anything with a defined conversion from a string
+               function <void(const type &value)>, // 🆕 int, bool, float, enum, vector, or string-like, or anything with a defined conversion from a string
                help_string="")
 
 app.add_complex(... // Special case: support for complex numbers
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -208,42 +208,42 @@ app.add_flag(option_name,
...
+app.add_flag_callback(option_name,function<void(void)>,help_string="") // 🆕
 
 // Add subcommands
 App* subcom = app.add_subcommand(name, description);
 
-Option_group *app.add_option_group(name,description); // 🚧
+Option_group *app.add_option_group(name,description); // 🆕
 
-// 🚧 All add_*set* methods deprecated in CLI11 1.8 - use ->transform(CLI::IsMember) instead
+// 🆕 All add_*set* methods deprecated in CLI11 1.8 - use ->transform(CLI::IsMember) instead
```
phlptp: For clarity sake maybe consider some sort of mark in the readme next to each deprecated function,
maybe ⚠️ or something similar

README.md
```diff
@@ -465,66 +465,66 @@ even exit the program through the callback. The main `App` has a callback slot,
 You are allowed to throw `CLI::Success` in the callbacks.
 Multiple subcommands are allowed, to allow [`Click`][click] like series of commands (order is preserved).  The same subcommand can be triggered multiple times but all positional arguments will take precedence over the second and future calls of the subcommand.  `->count()` on the subcommand will return the number of times the subcommand was called.  The subcommand callback will only be triggered once unless the `.immediate_callback()` flag is set.  In which case the callback executes on completion of the subcommand arguments but after the arguments for that subcommand have been parsed, and can be triggered multiple times.
```
phlptp:
```suggestion
Multiple subcommands are allowed, to allow [`Click`][click] like series of commands (order is preserved).  🆕 The same subcommand can be triggered multiple times but all positional arguments will take precedence over the second and future calls of the subcommand.  `->count()` on the subcommand will return the number of times the subcommand was called.  The subcommand callback will only be triggered once unless the `.immediate_callback()`  🆕 flag is set.  In which case the callback executes on completion of the subcommand arguments but after the arguments for that subcommand have been parsed, and can be triggered multiple times.
```

README.md
```diff
@@ -465,66 +465,66 @@ even exit the program through the callback. The main `App` has a callback slot,
...
 -   `.footer(message)`: Set text to appear at the bottom of the help string.
 -   `.set_help_flag(name, message)`: Set the help flag name and message, returns a pointer to the created option.
 -   `.set_help_all_flag(name, message)`: Set the help all flag name and message, returns a pointer to the created option. Expands subcommands.
 -   `.failure_message(func)`: Set the failure message function. Two provided: `CLI::FailureMessage::help` and `CLI::FailureMessage::simple` (the default).
 -   `.group(name)`: Set a group name, defaults to `"Subcommands"`. Setting `""` will be hide the subcommand.
-- `[option_name]`: 🚧 retrieve a const pointer to an option given by `option_name` for Example `app["--flag1"]` will get a pointer to the option for the "--flag1" value,  `app["--flag1"]->as<bool>()` will get the results of the command line for a flag.
+- `[option_name]`: 🆕 retrieve a const pointer to an option given by `option_name` for Example `app["--flag1"]` will get a pointer to the option for the "--flag1" value,  `app["--flag1"]->as<bool>()` will get the results of the command line for a flag.
```
phlptp:
```suggestion
- `[option_name]`: 🆕 retrieve a const pointer to an option given by `option_name` for Example `app["--flag1"]` will get a pointer to the option for the "--flag1" value,  `app["--flag1"]->as<bool>()` will get the results of the command line for a flag. The operation will throw an exception if the option name is not valid.
```

README.md
```diff
@@ -658,7 +658,7 @@ The App class was designed allow toolkits to subclass it, to provide preset defa
...
+The most important parse function is `parse(std::vector<std::string>)`, which takes a reversed list of arguments (so that `pop_back` processes the args in the correct order). `get_help_ptr` and `get_config_ptr` give you access to the help/config option pointers. The standard `parse` manually sets the name from the first argument, so it should not be in this vector. You can also use `parse(string, bool)` to split up and parse a string; the optional bool should be set to true if you are
 including the program name in the string, and false otherwise.
 
 Also, in a related note, the `App` you get a pointer to is stored in the parent `App` in a `unique_ptr`s (like `Option`s) and are deleted when the main `App` goes out of scope.
```
phlptp:
```suggestion
Also, in a related note, the `App` you get a pointer to is stored in the parent `App` in a `shared_ptr`s (similar to `Option`s) and are deleted when the main `App` goes out of scope unless the object has another owner.
```

README.md
henryiii: That looks like a good idea. 👍

henryiii: Added, at least to the two places I could think of putting it.

PR #285, opened 2019-05-21
modified option template

phlptp:
This PR modifies one of the add_option templates allowing two template parameters to allow direct specification of the conversion type if it is not desired to be the same as the assignment type. 

For example if you were assigning to variant,  you could specify the assignment as a string, int, double, enum, etc.  

Also add overloads of lexical cast for types which can take assignment from double or int.  

What this would allow is the complete removal of any mention or inclusion of optional from CLI11, and still maintain support for optional types.   `Optional<string>`  `optional<double>` or `optional<int>`  all work with no additional code, and other optional types can be made to work using the two parameter template as long as the underlying type has a lexical cast conversion.  

@henryiii Not sure if this is a direction you want to take, but it could completely get rid of a file and add some additional capabilities that we have used in our applications.  

If you think it a good direction I will add some more tests to make sure everything works as advertised with the two parameter template.


phlptp:
looks like some issues with the single_file generation due to commenting out some code in Optional.hpp.  I will fix it later if this is a useful direction.

henryiii:
I think this is a good direction to go. I was beginning to think about ways to remove the explicit inclusion of the optional.

phlptp:
I think this is working mostly how I intended now.    Got into some strange corners with different compilers.  
https://stackoverflow.com/questions/56296149/narrowing-cast-warnings-in-sfinae-evaluation.  

As near as i can figure Gcc allows narrowing conversions using {} syntax in some circumstances, other compilers do not.    Which isn't necessarily a big problem since it wasn't supposed to in template code, but for certain compilers gcc <5.0 and gcc 7 in C++17 it issued a warning in unevaluated context.  I haven't traced this yet but since it was fixed in later compilers I am assuming it was a bug of some sort.    Now also ran into some issues with visual studio 2015 and the type_trait is_assignable, so that led to some "interesting" issues with some peculiarly designed classes to them which explicitly deleted all constructors and assignment operators but some particular ones.    So in VS2015 it was saying things were assignable but then when you actually tried the assignment it would generate a compile errors.    Which kind of made the trait not so useful.   

 I tried making some different forms of the assignable trait but it must be something with the compiler cause whatever I tried did the same thing.    So in the end I just disabled the tests for VS 2015 and earlier.    I don't actually think this affects the actual usage much since it is pretty strange classes that trigger it, that were designed to target very specific overloads.  In actual code it would be pretty easy to work around.

phlptp:
@henryiii I think this is ready for review.

include/CLI/TypeTools.hpp
```diff
@@ -124,19 +124,62 @@ struct pair_adaptor<
     }
 };
 
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wnarrowing"
```
henryiii, line 135: Could you add a little comment as to why this warning needs to be suppressed?

include/CLI/TypeTools.hpp
```diff
@@ -124,19 +124,62 @@ struct pair_adaptor<
...
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif
+
 // Check for streamability
 // Based on https://stackoverflow.com/questions/22758291/how-can-i-detect-if-a-type-can-be-streamed-to-an-stdostream
 
-template <typename S, typename T> class is_streamable {
-    template <typename SS, typename TT>
+template <typename T, typename S = std::ostringstream> class is_streamable {
```
henryiii: Should we rename this to `is_ostreamable`?

include/CLI/TypeTools.hpp
```diff
@@ -124,19 +124,62 @@ struct pair_adaptor<
...
     static auto test(int) -> decltype(std::declval<SS &>() << std::declval<TT>(), std::true_type());
 
     template <typename, typename> static auto test(...) -> std::false_type;
 
   public:
-    static const bool value = decltype(test<S, T>(0))::value;
+    static const bool value = decltype(test<T, S>(0))::value;
+};
+
+// check for input streamability
```
henryiii: Three slashes here, please.
```suggestion
/// Check for input streamability
```

include/CLI/TypeTools.hpp
```diff
@@ -124,19 +124,62 @@ struct pair_adaptor<
...
+    template <typename TT, typename SS>
+    static auto test(int) -> decltype(std::declval<SS &>() >> std::declval<TT &>(), std::true_type());
+
+    template <typename, typename> static auto test(...) -> std::false_type;
+
+  public:
+    static const bool value = decltype(test<T, S>(0))::value;
 };
 
+// templated operation to get a value from a stream
```
henryiii:
```suggestion
/// Templated operation to get a value from a stream
```

include/CLI/TypeTools.hpp
```diff
@@ -343,21 +414,112 @@ bool lexical_cast(std::string input, T &output) {
     return true;
 }
 
-/// Non-string parsable
+/// assignable from double or int
```
henryiii:
```suggestion
/// Assignable from double or int
```

include/CLI/TypeTools.hpp
```diff
@@ -343,21 +414,112 @@ bool lexical_cast(std::string input, T &output) {
...
 template <typename T,
           enable_if_t<!std::is_floating_point<T>::value && !std::is_integral<T>::value &&
-                          !std::is_assignable<T &, std::string>::value && !std::is_enum<T>::value,
+                          !std::is_assignable<T &, std::string>::value &&
+                          !std::is_constructible<T, std::string>::value && !std::is_enum<T>::value &&
+                          is_direct_constructible<T, double>::value && is_direct_constructible<T, int>::value,
```
henryiii, line 428: TODO (not necessarily in this PR): I think we can make combinations of these as traits and use those; this is getting very long and hard to parse by eye.

phlptp: I made an attempt at this in PR #286

henryiii:
Rebased. It looks like the `string_view` addition in #300 broke this.

phlptp:
I guess there are a few oddities about appending string_view to a std::string,  It compiles now on visual studio with C++17 and on gcc 9 at least in MSYS  now.

henryiii:
This *almost* works if you remove the special string_view patch. Except on GCC 9. Any ideas?

phlptp:
I think I fixed it.  had to convert lexical cast to take a const std::string & instead so it wasn't converting a temporary variable to a string_view.  

Out of curiosity I tried making the value a string_view for add_option.  It compiled fine but doesn't work, since the input vector of strings to the callback is constructed on the fly as is deleted soon after, so that will never work.

though I do believe the `results` function or `as` will work using string_view since that just does the conversion directly on the results_ variable unless you are using the JOIN as a multioption policy and have more than one input.

Issue #289, opened 2019-06-08
Integer overflow is undefined behavior


jamesjer:
While building CL11 1.8.0 on an x86_64 machine with gcc 9.1.1 (Fedora Rawhide), this test in HelpersTest fails:

[----------] 4 tests from CheckedMultiply
[ RUN      ] CheckedMultiply.Int
/builddir/build/BUILD/CLI11-1.8.0/tests/HelpersTest.cpp:398: Failure
Value of: CLI::detail::checked_multiply(a, b)
  Actual: true
Expected: false
[  FAILED  ] CheckedMultiply.Int (0 ms)

The issue is that checked_multiply relies on integer overflow wrapping around, but that is undefined behavior according to the standard.  When compiling with gcc, -fwrapv must be passed to the compiler so that it doesn't optimize away such code.  I have verified that the test passes after adding that option to the build flags.

PR #290, opened 2019-06-11
remove undefined bahavior

phlptp:
change the checked_multiply function to not use undefined behavior to check for potential undefined behavior and wrapping.

See Issue #289

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -1,5 +1,4 @@
 #pragma once
-
 // Distributed under the 3-Clause BSD License.  See accompanying
 // file LICENSE or https://github.com/CLIUtils/CLI11 for details.
 
@@ -9,6 +8,7 @@
 #include <cmath>
 #include <functional>
 #include <iostream>
+#include <limits>
 #include <map>
 #include <memory>
 #include <string>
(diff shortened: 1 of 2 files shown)


himikof:
I would like to point out that while the intent of this change is correct, the implementation fails in an edge case:
```c++
assert(checked_multiply(2, std::numeric_limits<int>::min() / 2));
```
I would suggest the following change, which should work for both signed and unsigned types (the second branch is never called for them):
```c++
    if (a > 0 == b > 0) {
        if ((std::numeric_limits<T>::max)() / absval(a) < absval(b))
          return false;
    }
    else {
        if ((std::numeric_limits<T>::min)() / absval(a) > -absval(b))
            return false;
    }
```

henryiii:
@phlptp, will you add the test and fix from @himikof, or should I?

phlptp:
I won't be able to get to it until tomorrow morning, if you want to wait I can do it then, otherwise go ahead.  My one concern was the possibility of unused code warnings for the unsigned types.

henryiii:
That's sooner than I could do it; I'm edging out a little time during the CoDaS-HEP workshop to catch up a little. Won't start having larger chunks of time till next week or so.

phlptp:
I think this is ready to go now.  with the additions recommended by @himikof

PR #299, opened 2019-07-20
meson: Basic meson support

mensinda:
With this patch, CLI11 can be used as a meson
subproject: http://mesonbuild.com/Subprojects.html

However, CMake is still required for testing and
installation. The current meson.build is not a
complete replacement.


mensinda:
The Azure CI errors are unrelated.

henryiii:
I plan to fix the CI then merge this, probably in the next few days...

henryiii:
Could you rebase and force push (or just merge) with current master? Thanks!

mensinda:
Done. Let's hope the CI gets green :)

henryiii:
How hard would it be to add a new job to Azure to run meson with a little example project that uses this? (Not very familiar with Meson)

mensinda:
Depends. That project would have to include the CLI11 repository in its's subproject directory. So, copying the CLI11 repo in the CI or a seperate repository would be required. Also python (>= 3.5), meson and ninja would be required on all images.

The actual project would be fairly simple and easy to provide, however, implementing this in the CI will probably the most difficult part.

henryiii:
Could you provide the example project? I can probably set up the ci pretty easily. At the minimum, I could do a `brew install meson` on a mac image. We can symlink CLI11 into the directory as part of the test before running the meson build. I can merge this PR and we work in a new PR, or we can work in this PR, your call.

mensinda:
Adding it to this PR should work. Do you have any preferences where to put the test?

henryiii:
In a folder in the test folder. If it looks really useful as an example it might eventually be moved into the examples folder.

mensinda:
Added the test project. Also a minimal README.md, in case this will be a meson example.

henryiii:
This does not add the C++ standard setting, so it fails if you build on LLVM (or probably GCC < 6).

```
ccache c++ -Imain@exe -I. -I.. -I../subprojects/CLI11/include -Xclang -fcolor-diagnostics -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -g -MD -MQ 'main@exe/main.cpp.o' -MF 'main@exe/main.cpp.o.d' -o 'main@exe/main.cpp.o' -c ../main.cpp
```

(macOS)

mensinda:
Should be fixed now (defaults to `c++11`)

henryiii:
I believe it does not pick up the default (should be minimum...) C++ standard from a target. Not sure why this is supposed to be better than CMake... But, okay.

henryiii:
It also does not regenerate correctly and has to have the build directory removed before it picks up the change...

mensinda:
I mean, the C++ std is now `c++11` for the project by default with my changes. The default options are only applied when generating a new build directory. Changing options of an already existing directory has to be done with `meson configure`

henryiii:
Yes, but I would have thought it would have made it C++11 because the sub-project was C++11, but that's not too bad. When I ran meson build again, it said that ninja would regenerate if necessary, but it did not because the build file changed. Just seems like usability issues.

henryiii:
Should be ready to merge. Would you like it in examples instead of tests? I'd be okay with that.

mensinda:
I would like to keep it in tests for now. It is a bit minimal for an example in the current state. Maybe I do a new PR in the future with a bit more content

henryiii:
Okay. Thanks!

PR #300, opened 2019-07-25
Fix std::string_view support in transforming validators

peterazmanov:
Transforming validators convert ```const char *``` to ```std::string```, but don't do the same for ```std::string_view```. This causes UB in ```details::lexical_cast``` that returns ```std::string_view``` holding dangling pointer.

### DIFF
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -9,9 +9,12 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+
+// [CLI11:verbatim]
 #if defined(CLI11_CPP17)
 #include <string_view>
 #endif
+// [CLI11:verbatim]
 
 namespace CLI {
 
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -9,6 +9,9 @@
 #include <string>
 #include <type_traits>
 #include <vector>
+#if defined(CLI11_CPP17)
+#include <string_view>
+#endif
 
 namespace CLI {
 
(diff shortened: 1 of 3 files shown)


henryiii:
I'm thinking of moving the `string_view` part to the test, and using CLI11 + StringView would require a user add the overload. I'm not fond of adding a new include, it adds a dependency that can sometimes fail (as seen in the tests). I was hoping I could add a declaration instead of an include, but I'm not sure how to do that since it's really `std::__1::basic_string_view` and I don't think I want to rely on the `__1` being there.

The other change is fine, adding a test is fine, and it might be something improved by #285.

peterazmanov:
> I'm not fond of adding a new include, it adds a dependency that can sometimes fail (as seen in the tests).

```std::string``` actually includes ```<string_view>``` for ```__cplusplus >= 201703L```. So there is no additional overhead.

> ... and using CLI11 + StringView would require a user add the overload.

I don't know how to prevent usage of ```std::string_view``` without adding the overload.

> I was hoping I could add a declaration instead of an include, but I'm not sure how to do that since it's really `std::__1::basic_string_view` and I don't think I want to rely on the `__1` being there.

AFAIK adding forward declarations for STL classes is not allowed by the standard.

henryiii:
> `std::string` actually includes `<string_view>` for `__cplusplus >= 201703L`. So there is no additional overhead.

I don't think that's required, but it's good enough for me.

> AFAIK adding forward declarations for STL classes is not allowed by the standard.

Yes, I believe so, which is why they are fine to put them in inline namespaces...

> > it adds a dependency that can sometimes fail (as seen in the tests)

Okay, this was my mistake, now it looks fine.

PR #304, opened 2019-07-27
Moving tutorial docs

henryiii:
Moving the tutorial docs to GH.


henryiii:
This may not build properly yet, but that will be easiest to test/adjust after merging.

Issue #295, opened 2019-06-22
Multiple positional arguments are greedy (order is not maintained)


kelvinhammond:
Positional arguments 

```
std::vector<std::string> sources;
std::string dest;
sub->add_option("src", sources, "Source files");
sub->add_option("dest", dest, "Destination folder or directory");
```
This is the output from `./cmd 1 2 3
```
src: 1 2 3
dest: 
```

The first `src` argument consumes the last argument which should be `dest`.

WenbinHou:
I'm facing the same issue. Any ideas please?

Set `src` and `dest` as required() does not work.

phlptp:
I added a solution in PR #306.   The way it would work is 
```cpp
std::vector<std::string> sources;
std::string dest;
sub->add_option("src", sources, "Source files");
sub->add_option("dest", dest, "Destination folder or directory")->required();
sub->positionals_at_end();
```

The other option that I think would work on the existing code is to apply validators on the positionals  and the 'validate_positionals()'  flag,  then if source files have to be existing files and dest is a folder they can have different validators and the positional validation will not load them if they don't validate and move on to the next one.    That isn't a general solution but if you are actually looking at files that should work.

henryiii:
Fixed in #306 and will be released in 1.9.

Issue #297, opened 2019-07-16
Reading values passed after command option as list of strings


lkammath:
Hi,

I have to write an application that takes raw hex bytes as input

./app -raw 0x05 0x04 0x03 .. some thing like this

I will interpret the data got and parse it to see I have received in the proper hex format

I know that add_option() will take a vector as argument which we can access like arg[0], arg[1] which will be commands and it is number of subcommands

    std::vector<std::string> args{};
    app.add_option("-r, —raw", args, "Send a RAW data and print response");

Is there a way I can read all the data passed after -raw option as list of strings? and treat arg[0] as the main command instead of treating them as subcommands?

Thanks in advance for the help

Thanks,
Lakshmi

henryiii:
Currently, your example `./app -raw 0x05 0x04 0x03` would produce the vector: `["0x05", "0x04", "0x03"]`. Is that not what you want? If you want `["-raw", "0x05", "0x04", "0x03"]` instead, that could be done by making the argument a positional argument, that is, give it a name without a dash. `app.add_option("raw", args, "Send a RAW data and print response");`. I may not understand your question, though.

lkammath:
so, the second case you explained by making argument as positional argument

arg[0] would point to -raw or raw? and arg[1] onwards i can treat as the input raw bytes and put them to list?

lkammath:
I need some way to parse a normal command vs a raw command passed

For e.g i have same application accepting another command like 

./app -command [some command name] [sub command1]

The same app will also have option like i told above

./app -raw/raw 0x05 0x04 0x03, etc.

I hope i am clear now

Thanks,
Lakshmi

henryiii:
In CLI11, long options must start with a double dash. Subcommands should have no dash. So you can do what you want to rather easily as long as you can drop the dash, that is:
    
    ./app command [some command name] [sub command1]
    ./app raw/raw 0x05 0x04 0x03

Here's an example: <https://wandbox.org/permlink/rMGeZwjyGxYtY9Im>

Change the "raw" to "command" and you'll see that it's filling in a different "remaining" vector.

lkammath:
Sorry .. I messed up again

I meant , I want to have two options in my app

./app --command [main command][subcommand1] - This one i have already implemented
./app --raw 0x05 0x04 0x03 - This is the one i am asking suggestion for.. Should I change it to a positional argument like what you answered in the first comment? Then i will lose -- before raw

Thanks,
Lakshmi

[...]

PR #306, opened 2019-07-30
required positional arguments and a vector positionals

phlptp:
This PR adds a loop on the positional parsing to check for remaining required positionals and prioritizing them if positionals_at_end is activated.  

it addresses issue #295

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -2326,11 +2326,15 @@ class App {
     /// Count the required remaining positional arguments
     size_t _count_remaining_positionals(bool required_only = false) const {
         size_t retval = 0;
-        for(const Option_p &opt : options_)
-            if(opt->get_positional() && (!required_only || opt->get_required()) && opt->get_items_expected() > 0 &&
-               static_cast<int>(opt->count()) < opt->get_items_expected())
-                retval = static_cast<size_t>(opt->get_items_expected()) - opt->count();
-
+        for(const Option_p &opt : options_) {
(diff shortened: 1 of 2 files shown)


henryiii (approved):
This looks good to me - I was going to suggest adding a test for two ending positionals but you already have one. 👍 @kelvinhammond, @WenbinHou, does this look good to you? The `positionals_at_end` setting is already there, this just modifies the way it works with an ending positional.

phlptp:
I will add that it will ONLY modify current operations if there are required positionals and the positionals_at_end flag is activated.

henryiii:
Do we need the `positionals_at_end` and required only behavior? Why would you ever want an unlimited positional followed by limited positionals to not match the limited positionals? Otherwise, they would never match.

phlptp:
I think the biggest reason was to ensure the behavior worked and reduce potential ambiguities.  If you didn't require positionals_at_end you could have regular arguments at the end of the args vector and the option would not trigger. 

I think if you removed that, you would add a bit of overhead for regular positionals and might be confusing sometimes why it wouldn't work, but otherwise in the cases tested it should work equivalently.  Right now there is a specialized case in which the check occurs, and only a bool check overhead in other cases.   

It might be reasonable to expand it further but it might be worth considering clarifying and expanding desired rules for positional argument priority first.  This topic also has some relationship with #297.

WenbinHou:
So why this PR is not merged till now?

henryiii:
I think we can go ahead and merge now, then address the "worth considering clarifying and expanding desired rules for positional argument priority first" as listed above.

Issue #296, opened 2019-06-27
Options that take an index + value argument.


mathiaswagner:
While I think it is possible to hack things with the function passed to add_options I was wondering whether there is a cleaner solution to handle options that take a pair of values.
There is #128 but what I need are options like

`--something 1 "double"  --something 2 "single" --other 1 16.0 --other 3 0.0`.

Basically you can think of the `--something` option being a vector but instead of specifying 
`--something "double" "single"` the 1 and 2 as the first argument to the option specify the index in that vector. For the 2nd argument I still want to be able to use the transform / check options.

The most fancy use case is when you have a vector of vectors, like:
`--vectorsomething 1 12 8 6 4 --vectorsomething 2 4 4 4 4`
Here the first number again gives the position in a vector but the components are itself a vector (here: `12 8 6 4` and `2 2 2 2`).

The vector of vector thing is optional, but I am still looking for clean way to achieve the first thing.
Current hack looks like:
```
std::vector<int> mgpre(3);
  auto entry_option{
      app.add_option("--entry",
                     [&mgpre](CLI::results_t vals) {size_t l; int j;// results_t is just a vector of strings
                      bool worked = true;
                      
                      std::map<std::string,int> prec_map({{"double", 8}, {"single", 4}, {"half", 2}});
                      CLI::Range validlevel(0,2);
                      CLI::CheckedTransformer prec_transform(prec_map);
                      for (size_t i{0}; i < vals.size() / 2; ++i){ // will always be a multiple of 2
                        validlevel(vals.at(2*i));
                        prec_transform(vals.at(2*i+1));
                        worked = worked and CLI::detail::lexical_cast(vals.at(2*i),l);
                        worked = worked and CLI::detail::lexical_cast(vals.at(2*i+1),j);

                      
                        if (worked) mgpre[l] = j;
                      }
                       return worked;
                     },
                     "level & value")};
  entry_option->type_name("level VALUE")->type_size(-2);
  entry_option->expected(-1);
```

but that is not very portable and errors also are not handled properly. If you have some pointer to where I could find useful components I am happy to work on that.

phlptp:
I was playing around a bit with support for a pair(possibly generalized to a tuple) but will start with pair,   i think that would allow your use case.  Then for the first case at least it would look like
```cpp
std::pair<size_t, int> index;
app.add_option("--entry",index," pair option");  
```
just like any other single variable.  

Might also be able to support complex directly without the extra function for that currently.

[...]

PR #307, opened 2019-07-30
Support tuple

phlptp:
If merged this pull request will enable full support for tuples up to 5 elements in size.   This includes support for std::pair and std::array(up to 5 elements).


phlptp:
@henryiii   this is something I have been tinkering with and I wanted to see how close I am on coverage and platform support.  There is a bit more work to be done on testing, and possible support for things like vector<pair> 

I looked at the complex,  but there is a special support for "j" notation on complex numbers so can't get rid of that yet. 

you can take a look see if this is of interest

phlptp:
ok, obviously have a little ways to go to get broad compiler support

phlptp:
This should address #296 since you can now do
```cpp
std::pair<size_t, int> index;
app.add_option("--entry",index,"pair option");  
```

include/CLI/TypeTools.hpp
```diff
@@ -390,6 +448,60 @@ constexpr const char *type_name() {
...
+    typename T,
+    enable_if_t<classify_object<T>::value == tuple_value && type_count<T>::value == 4, detail::enabler> = detail::dummy>
+std::string type_name() {
+    return std::string("[") + type_name<typename std::tuple_element<0, T>::type>() + "," +
+           type_name<typename std::tuple_element<1, T>::type>() + "," +
+           type_name<typename std::tuple_element<2, T>::type>() + "," +
+           type_name<typename std::tuple_element<3, T>::type>() + "]";
+}
+
+/// Print type name for 5 element tuples
```
henryiii, line 494: Is the limit of 5 here just the printout for the typename?

include/CLI/TypeTools.hpp
```diff
@@ -607,6 +719,160 @@ bool lexical_assign(const std::string &input, T &output) {
...
+        retval &= lexical_cast(strings[2], std::get<2>(output));
+    }
+    if(strings.size() > 3) {
+        retval &= lexical_cast(strings[3], std::get<3>(output));
+    }
+    return retval;
+}
+
+/// conversion for five element tuple
+template <class T, class XC, enable_if_t<type_count<T>::value == 5, detail::enabler> = detail::dummy>
```
henryiii, line 856: It seems like there should be a recursive way to do this.

include/CLI/TypeTools.hpp
phlptp: It is all the places tuples are supported.  I tried to figure out how to do that with some template magic, but the initial solutions I tried got out of hand quickly so I didn't really like that solution.  On my other projects I had access to Boost so using the MPL  library or one of the other TMP libraries this gets all pretty easy, but I am a little scared to try to look under the hood for that stuff, and it didn't seem appropriate to start including Boost now.  So 5 is one greater than the biggest  tuple case I could conceivably find a use case for, which is where that came from.

include/CLI/TypeTools.hpp
phlptp: For some parts I think it is doable with a compile time recursion, but it wasn't clear to me how to transmit the runtime state along with the compile time recursion and keep everything sane.  Would love it if you or anyone knew how to do that but I was not seeing any clear solutions to get around the limit of 5 in a nice way.  Not sure how much value there is in supporting 6 element tuples either.

henryiii:
mp11 or (probably) hanna can do it, so it should be possible. For now, this is probably fine and is a nice feature to have.

phlptp:
I think my real question is not whether it can be done, but whether it can be done a straightforward and somewhat readable way.

phlptp:
From #296 what I don't know quite yet is how to handle the validators that might be distinct for each part of a tuple.   Not sure the best way to handle that yet.

henryiii:
A future improvement could be looking at how to do this without a limit and 5x repeated code.

PR #309, opened 2019-08-14
Footer callback

phlptp:
If merged this PR will add a callback for footer generation, to allow extra information into the help screen that can only be generated at run time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,6 +554,7 @@ There are several options that are supported on the main app and subcommands and
 -   `.positionals_at_end()`: 🆕 Specify that positional arguments occur as the last arguments and throw an error if an unexpected positional is encountered.
 -   `.prefix_command()`: Like `allow_extras`, but stop immediately on the first unrecognized item. It is ideal for allowing your app or subcommand to be a "prefix" to calling another app.
 -   `.footer(message)`: Set text to appear at the bottom of the help string.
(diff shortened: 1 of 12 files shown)


phlptp:
This is based on the tuple PR so once that is merged this will go out of draft, or it can be extracted if that direction is not desired.

henryiii:
Should there be a footer and a footer callback? Can't the footer sting function just make a callback that returns the string given? Other than that it looks good.

phlptp:
My initial thought was to do that as well, but the issue with that is footer as a string is inheritable, and I don't think you want a callback to be inheritable,  So there really isn't a good resolution between them other than leaving both in place.

see the [gitter question](https://gitter.im/CLI11gitter/Lobby?at=5ca0cc05016a930a453c312c) a few months ago.  I thought about making the callback inheritable but that has issues as well.  The use case I needed it for was a cascading app,  so I do some processing, then possibly depending on options called another processing app (embedded in another class) is called to process some more command line arguments.    I wanted the help to actually generate help for all of them.  So the solution was the footer callback (which we have been using for a couple months now), which would call the second app inside the footer callback to generate the help for the additional options inside the footer.  Eventually I hope there is a cleaner way that could be done but it works pretty well for the moment when some of the help is dynamic.  And I don't think that would work or even make sense if that callback were passed to the subcommands.

PR #310, opened 2019-08-17
Indexed tuple validation

phlptp:
If merged this PR will remove the restrictions on tuple size, and add an index to validators to allow them to apply to specific components of a tuple type or vector type.  

it should address #296

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1859,8 +1859,14 @@ class App {
             return detail::Classifier::SUBCOMMAND;
         if(detail::split_long(current, dummy1, dummy2))
             return detail::Classifier::LONG;
-        if(detail::split_short(current, dummy1, dummy2))
+        if(detail::split_short(current, dummy1, dummy2)) {
+            if(dummy1[0] >= '0' && dummy1[0] <= '9') {
+                if(get_option_no_throw(std::string{'-', dummy1[0]}) == nullptr) {
+                    return detail::Classifier::NONE;
+                }
+            }
             return detail::Classifier::SHORT;
+        }
         if((allow_windows_style_options_) && (detail::split_windows_style(current, dummy1, dummy2)))
             return detail::Classifier::WINDOWS;
         if((current == "++") && !name_.empty() && parent_ != nullptr)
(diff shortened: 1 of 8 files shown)


include/CLI/App.hpp
```diff
@@ -1859,8 +1859,14 @@ class App {
             return detail::Classifier::SUBCOMMAND;
         if(detail::split_long(current, dummy1, dummy2))
             return detail::Classifier::LONG;
-        if(detail::split_short(current, dummy1, dummy2))
+        if(detail::split_short(current, dummy1, dummy2)) {
+            if(dummy1[0] >= '0' && dummy1[0] <= '9') {
+                if(get_option_no_throw(std::string{'-', dummy1[0]}) == nullptr) {
+                    return detail::Classifier::NONE;
```
henryiii, line 1865: This fixes a bug mentioned earlier that -N can match when it should be a number, correct?

include/CLI/Validators.hpp
```diff
@@ -43,6 +43,8 @@ class Validator {
     std::function<std::string(std::string &)> func_{[](std::string &) { return std::string{}; }};
     /// The name for search purposes of the Validator
     std::string name_;
+    /// A validate will only apply to an indexed value (-1 is all elements)
```
```suggestion
    /// A validator will only apply to an indexed value (-1 is all elements)
```

include/CLI/Validators.hpp
```diff
@@ -113,7 +115,13 @@ class Validator {
         non_modifying_ = no_modify;
         return *this;
     }
-
+    /// Specify the application index of a validator
+    Validator &application_index(int app_index) {
```
henryiii, line 137: Could a const copying version work? This currently cannot be applied inline without changing the validator.

henryiii (approved):

include/CLI/App.hpp
phlptp: yes it should fix the bug mentioned [here](https://gitter.im/CLI11gitter/Lobby?at=5d5090521dadc42a11535752)

include/CLI/Validators.hpp
phlptp: I suppose the same is true for any of the validator modification functions.  `active`, `name`,`description`,`non_modifying`.  I think we would want them all to have the same paradigm.  

It would add a number of overloads but might make the calling a little easier for many of the const objects, which require an extra CLI::Validator( ) around them.  I will make that change tomorrow morning.

phlptp:
not sure why codacy isn't analyzing things?

PR #313, opened 2019-08-22
Third stage callback

phlptp:
If merged this PR will add a third callback to apps.

Based on the discussion from [gitter](https://gitter.im/CLI11gitter/Lobby?at=5d00f8c3c5386926f6d7da29)

It adds two new function `final_callback` and `parse_complete_callback`  The `callback function will populate one of these depending on the status of immediate_callaback.    immediate_callback will also toggle between them if one is empty.    The default is routing to final_callback which has the same behavior as prior to the PR.  

Some thoughts about deprecating the immediate_callback but losing it would remove capabilities at present,  so I don't think it should be deprecated yet at least until the callback function itself is deprecated.


phlptp:
I want to see where additional test coverage is needed for this PR, so I think it ready for review, but I will probably need to add some more tests yet.

phlptp:
I think everything is ready to go here.  I left `immediate_callback` operation in place to avoid any breaking changes.  So `callback` should behave as before even though the operation is technically redundant.  `immediate_callback` still functions from a user perspective as before even though the actual operations it does are very different.

README.md
```diff
@@ -408,36 +408,41 @@ will produce a check for a number less than 0.
...
 Validators are copyable and have a few operations that can be performed on them to alter settings.  Most of the built in Validators have a default description that is displayed in the help.  This can be altered via `.description(validator_description)`.
 The name of a Validator, which is useful for later reference from the `get_validator(name)` method of an `Option` can be set via `.name(validator_name)`
 The operation function of a Validator can be set via
-`.operation(std::function<std::string(std::string &>)`.  The `.active()` function can activate or deactivate a Validator from the operation.
-All the functions return a Validator reference allowing them to be chained.  For example
+`.operation(std::function<std::string(std::string &>)`.  The `.active()` function can activate or deactivate a Validator from the operation.  A validator can be set to apply only to a specific element of the output.  For example in a pair option `std::pair<int, std::string>` the first element may need to be a positive integer while the second may need to be a valid file.  The `.application_index(int)` 🚧function can specify this.  It is zero based and negative indices apply to all values.  
+```
```
```suggestion
```cpp
```

README.md
```diff
@@ -408,36 +408,41 @@ will produce a check for a number less than 0.
 
 ##### Transforming Validators
 There are a few built in Validators that let you transform values if used with the `transform` function.  If they also do some checks then they can be used `check` but some may do nothing in that case.
- * 🆕 `CLI::Bounded(min,max)` will bound values between min and max and values outside of that range are limited to min or max,  it will fail if the value cannot be converted and produce a `ValidationError`
```
henryiii, line 411: Why are you changing stars to dashes? I like stars *slightly* better than dashes, since they look more like the final result. If there's a reason to change, I'd be happy to change.

README.md
henryiii: (also should have a new line before, as well)

henryiii (commented):
I think this would really benefit from having the tutorial book updated. I'll try to merge that into CLI11 in the next week or two.

README.md
phlptp: A couple reasons, neither of which I consider definitive,  the first is consistency, we have used dashes in some spots and `*` in others,  I don't have a strong preference but I use `-` in everything else I edit,  I am fine either way.  The other is I by default run a markdown linter in my editor, and that always shows some warning, so I have slowly been resolving them with the `-`.   

Like I said a few non-definitive reasons if you want them changed back I would be happy to do so,  though some clarity on style might be useful.

phlptp:
Keeping the book up to date will probably be a lot easier once it is merged. I know I would be more inclined to update that if it was in a single repository.

README.md
henryiii: We can go to all dashes, I'd be okay with that; a linter liking dashes is possibly enough of a reason to change.

Issue #291, opened 2019-06-12
Add needs(option_or_subcommand) option for subcommand


Bouska:
Currently .excludes() exists for options, options groups and subcommands, but .needs() only exists for option.

In my case, some of my subcommands require a printer and therefore the parameters for that printer. I have a option group at the top app that are used for this, but I can't currently enforce they are present when I use the printer related subcommands. Is there a workaround the lack of .needs() or could it be implemented in the next releases?

phlptp:
I thought about this earlier, but there are a few corner-cases that were not obvious how to deal with and it wasn't required for the work we were doing at the time, so it never got implemented.   

If I understand what you want is a `.needs(...)` method  on an option_group/subcommand that requires another option_group/subcommand to be used as well?  

As far as what to do at present you could use the TriggerOn(printer_app,   vector of apps_requiring_printer) function to only enable those apps requiring a printer after the printer was specified.    

Or in the callbacks of the different apps do a check if the count_all() in the printer group>0  and throw a `RequiresError` if it doesn't.

Bouska:
In my case, I would have used `.needs(...)` method with a list of options pointers (or names) because I'm still figuring out how to use option groups without messing the help output, but the idea is the same.

I'm going to try both of your ideas to see if that works for me.

phlptp:
a needs function has been added for apps and subcommands with support for options and other subcommands.

Issue #288, opened 2019-06-03
Is there a way to alias a subcommand?


comfortablynick:
I would like to have subcommands with multiple names (long and short), such as `a`|`add` or `i`|`install`. Is there a way to do this without duplicating all of the command definitions?

Thanks!

phlptp:
Not that I know of directly.  The closest right now would be to create a shared_ptr to an app(do NOT give it a name) with all the options for a particular subcommand.  

Then in the main app create a subcommand for each of the names you want to be the same, and call add_subcommand(shared_app)  with the shared_ptr to the option_group that contains all the actual options.  Fallthrough will always go up to the last subcommand you added it to, but otherwise should work as you expect unless you are doing interesting things with the callbacks on the apps and option groups.

phlptp:
The current master branch has an `alias` function that will do this now.

Issue #284, opened 2019-05-21
The reason for exception OptionAlreadyAdded could be clarified


tobgu:
If I add an option with short_name="o", name="option", and then another option with short_name="o" and name="another-option" an OptionAlreadyAdded will be thrown since the short name "o" is already in use.

The error message will state that: "another-option is already added", however.

It would be nice if the message could also include the reason for the collision (eg. that the short name is already used).

jiggersplat:
I just spent an hour figuring out that i can't add an option with short_name="-h" because it is already used by the default help option. OptionAlreadyAdded was a total red herring.

zaphodikus:
A single comment in the header file to this effect would probably fix it for some people, but this will tend to only happen once.

Issue #272, opened 2019-05-01
name validation on subcommands


phlptp:
As we were integrating CLI11 into our applications, there was one location where we were using add_subcommand and had transposed the name and the description.  Everything ran normally until we go to a some tests where the subcommand would be called and nothing happened.  Due to the nature of the code it took a while to find the issue.    

What I am wondering about is whether add_subcommand should have a check on the name.  For example a simple check on whether the name contained any spaces would immediately distinguish it from most descriptions and catch that error pretty quickly.  

I suppose in theory you could have a subcommand name with spaces  and use quotes on the command line to trigger it, but I don't imagine that is a common case, and if a name restriction is in add_subcommand there is a workaround even for that situation.  Subcommands with a '-' or '--' in front could be problematic as well.    

So I am wondering if there should be a check in add_subcommand on the name and what the nature of that check should be,  a runtime check and throw,  an assert, or something else?

henryiii:
Just curious, would `app.add_subcommand("name")->description("This is a discription");` be better for avoiding mistakes?

phlptp:
definitely, but that would require deprecating the two string version to be effective.

PR #317, opened 2019-09-06
Add needs to subcommand

phlptp:
This PR will address #291 and #288 

it adds a needs function for subcommands and options to the app as well as a remove needs.  it also adds some alias operations to app so multiple subcommand names can represent a single subcommand.  

This will also work for nameless subcommands(option_groups) so the same group can be a subcommand or act just like regular options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,6 +535,7 @@ There are several options that are supported on the main app and subcommands and
 -   `.enabled_by_default()`: 🆕 Specify that at the start of each parse the subcommand/option_group should be enabled.  This is useful for allowing some Subcommands to disable others.
 -   `.validate_positionals()`: 🆕 Specify that positionals should pass validation before matching.  Validation is specified through `transform`, `check`, and `each` for an option.  If an argument fails validation it is not an error and matching proceeds to the next available positional or extra arguments.
 -   `.excludes(option_or_subcommand)`: 🆕 If given an option pointer or pointer to another subcommand, these subcommands cannot be given together.  In the case of options, if the option is passed the subcommand cannot be used and will generate an error.
+-   `.needs(option_or_subcommand)`: 🚧 If given an option pointer or pointer to another subcommand, the subcommands will require the given option to have been given before this subcommand is validated.
 -   `.require_option()`: 🆕 Require 1 or more options or option groups be used.
 -   `.require_option(N)`: 🆕 Require `N` options or option groups, if `N>0`, or up to `N` if `N<0`. `N=0` resets to the default to 0 or more.
 -   `.require_option(min, max)`: 🆕 Explicitly set min and max allowed options or option groups. Setting `max` to 0 implies unlimited options.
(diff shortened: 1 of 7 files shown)


phlptp:
- [x] write up readme documentation changes
- [x] a few tests of different callback ordering and the needs operation in context of argument ordering
- [x] a little more exploration of the implications of the aliases
- [x] get full code coverage

phlptp:
I have run into a few issues with the interaction of needs and the callback operations and the fallthrough flag.  Still working on how to resolve these, but it will probably be a few days before things come together.

phlptp:
This PR should also address #284 when finished

phlptp:
#272 is also addressed, since the aliases needed to be checked as well

phlptp:
In considering Aliases,  there is a possibility of conflict with existing subcommands and ignore_case, ignore underscore variants of such.   This required somewhat more complicated checking of the subcommand names.  But also in exploring it the ignore_case, ignore_underscore, name functions also did not currently check this, which means it was possible to generate conflicts.  So these checks were added and those functions were made exception safe.    more testing is required

phlptp:
@henryiii I have been adding checks to make sure subcommands don't overlap in name or alias.  
Previously we were not checking overlap on the option_groups, so multiple option_groups could have the same subcommand name See [SameSubcommand](https://github.com/CLIUtils/CLI11/blob/b029b325b50389a2e11d6361c8b3112206bff56d/tests/OptionGroupTest.cpp#L523)  The question is do we want to allow this explicitly or not?  Basically should we allow the potential for using option_groups to have the same subcommand mean multiple different things if called multiple times.    

Options
1.  Disallow the same name to potentially reference multiple subcommands
2.  Allow the same name to reference multiple subcommands through option_groups(this is what is done currently)
3.  Add a flag that explicitly allows 2 if set to true.

phlptp:
@henryiii   We would need a decision on 1,2 or 3 from the previous question to be able to move forward.  I don't really have a strong preference either way.   I was inclined towards option 1, but that would break an existing test case.

phlptp:
So what I ended up doing with regard to subcommand names, is disable the check if the subcommand is disabled.  So if the existing behavior is needed there is a means of allowing it, but in general it would trigger an error.

phlptp:
@henryiii I think something is wrong with the travis gcc 4.8 build and the azure clang 3.4 builds

henryiii:
What's the current status? Do you think it's ready for a final review and merge?

phlptp:
I think so, I haven't found anything I have wanted to change in a while.  I think there is a possibility of more work needed on the help output in relation to the changes from this PR, but I really don't know the best way to present that information yet and I think it will take some experience using it to know for sure.

henryiii (approved):

henryiii:
I think it looks good enough to merge. We can adjust things like help output in the future.

PR #319, opened 2019-09-18
refine the implementation of has_find type_traits

fpeng1985:
sfinae_true and test_find are not used elsewhere, so a single has_find is enough to implement the function of the type trait.


henryiii:
Thanks!

PR #320, opened 2019-09-19
Option check method error

fpeng1985:
the validator is added using "push_back", but the name is set to the front of validators_, which  i think is a mistake~


phlptp (approved):

henryiii:
Thanks!

PR #323, opened 2019-09-25
Require semicolon after CLI11_PARSE()

elazarg:
Placates some analysis tools that warn about empty statements.


henryiii:
This adds a rather ugly construct, just to break people's code that does not have a simicolon. A better solution might be just to leave of the simicolon if a tool complains. Or to avoid the macro entirely.

henryiii:
This part of the code will change quite a bit anyway in 2.0, so will close for now.

Issue #322, opened 2019-09-22
 A bug with take_first take_last and join 


fpeng1985:
```
std::string files;
auto opt = app.add_option("-f,--file,file", files, "File name");
opt->take_first();
opt->take_last();
opt->join();
```

The above code snipet will throw an exception from `multi_option_policy` method in `Option` class. In fact `take_first`, `take_last` and `join` can only be called once. According to the docs, these methods will only be used for vectors(may be flags). So it is better to hide these interfaces in other situations, or fix these bug.

To wholy avoid these interface conflict, a refactor to the `Option` level is needed. `Option` class itself or its internal implementation need different types to provide polymorphism. I want to help implement the enhancement, but some part of the source code is still unclear to me, especially `type_size_` `expected_` and their relationship with `multi_option_policy_`, would you give some hints about these? Thanks!

The following is a possible implementation. It seems like an enhanced version of CRTP, but with an important differnce. These base classes themselves can inherite from each other, so the code organization is similar to a virtual function based method, but with static dispatching.
Please gives your feedback, thanks again!

```
template<typename Host>
struct CommonStrategy {
    //common data
    bool configurable_{true};
    std::vector<std::string> snames_;

    //common behavior
    bool configurable() const   { return configurable_; }
    void configurable(bool val) { configurable_ = val;  }
};

template<typename Host>
struct OptionStrategy : public CommonStrategy<Host> {

    //customized behavior
    std::string get_names() {
        std::ostringstream out;
        //access common data
        auto &self = static_cast<Host&>(*this);
        for(auto &name:  self.snames) {
            out << name << ",";
        }
        return out.str();
    }
};

template<typename Host>
struct FlagStrategy : public CommonStrategy<Host> {

    //customized behavior
    std::string get_names() {
        std::ostringstream out;

        //access common data snames_
        auto &self = static_cast<Host&>(*this);
        for(auto &name:  self.snames_) {
            out << name << ",";
        }
        //access extra data fnames_
        for(auto &name : fnames_) {
            out << name << ",";
        }
        return out.str();
    }

    //extra data
    std::vector<std::string> fnames_;

    //extra behavior
    const std::vector<std::string> & get_fnames() const {
        return fnames_;
    }
};

template<template<typename> class Strategy>
struct Host : public Strategy<Host<Strategy>> {
};

int main() {
    Host<FlagStrategy> flag;
    flag.configurable(false);//call common method
    flag.get_names();//will prints name in a specific way
[...]

Issue #305, opened 2019-07-27
Apply validators to defaults if no value was set from the comand line


Ozaq:
A rather common use case for me is to validate the existence of files passed to my program via commandline arguments. CLI11 features `Opt::check(CLI::ExistingFile)` but It seems to not get called if the option was not present on the command line but there is still a default, e.g.
```
std::filesystem::path file{"myfile"};
CLI::Option* opt1 = app.add_option("--inifile", file, "Path to your inifile");
```
I saw the workaround in #115 so consider this just as a though for improvement.

phlptp:
So with #325 the value of the option specified as a default_str is run through the validators, which is wasn't previously.  It still isn't going to check the value just stored in a variable that might get assigned.

Issue #257, opened 2019-03-22
std::optional<std::vector<double>> doesn't compile


jiggersplat:
The following does not compile:

`app.add_option("-v,--vec", std::optional<std::vector<double>>, "some vector")->expected(3);
`

I can just check length as a work-around, but thought I would put this out there. Not sure this is a common or easy to support use-case.

phlptp:
I think right now optional only works with types that support >> operator.  I think it would take some template magic on the add_option templates to separate the actual storage type from the conversion types and even then I am not sure the it would work to assign the result to the optional containing a vector since you need an indirect to use the methods on an optional.  

In which case you best bet if you really want an optional would be to use
```cpp
std::optional<std::vector<double>> opt;
add_option_function<std::vector<double>>("-v,--vec", [&opt](auto &v){opt=v;}, "some vector")->expected(3);
```
This will only work on master right now, not 1.7 release

 @henryiii might have a better idea

henryiii:
Optional's of normal values are supported (>> operator), but not sure why you'd want an optional vector. Can't the vector just be empty if no arguments are passed?

phlptp:
this actually works now if you use the two parameter template for add_option, and you can specify expected_size(0) on the option if you actually wanted it to be optional.

PR #325, opened 2019-09-30
Type size refactor

phlptp:
If merged this PR will refactor the type_size and expected size and the multiOptionPolicy so they work independently and allow more flexible types in add_option. 

Also rework add_complex to actually allow things like `--comp 4+2i`  or `--comp -1.7-5j` in addition to the current pair of doubles support.

### DIFF
--- a/book/chapters/options.md
+++ b/book/chapters/options.md
@@ -165,7 +165,7 @@ Windows style options do not allow combining short options or values not separat
 
 ## Parse configuration
 
-How an option and its arguments are parsed depends on a set of controls that are part of the option structure.  In most circumstances these controls are set automatically based on the function used to create the option and the type the arguments are parsed into.  The variables define the size of the underlying type (essentially how many strings make up the type), the expected size (how many groups are expected) and a flag indicating if multiple groups are allowed with a single option.  And these interact with the `multi_option_policy` when it comes time to parse.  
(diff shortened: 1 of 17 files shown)


phlptp:
This is getting closer but wanted to see the coverage and let people start taking a look.

henryiii:

phlptp:
[clang tidy](https://dev.azure.com/CLIUtils/CLI11/_build/results?buildId=352) is not exactly intuitive what was going on there.  It is a nice check but it would be helpful if the actual warning was displayed along with the code, not sure if that is possible or this was a diff of some kind.

henryiii:
I think clang-tidy should print info messages while it builds, then the check that fails is just a diff. So if you look back in the build log, you should see a reason, I think.

henryiii:
I think clang-tidy should print info messages while it builds, then the check that fails is just a diff. So if you look back in the build log, you should see a reason, I think.

phlptp:
Yeah I guess it does, I was just looking at the step that showed the failure didn't occur to me to check the build for the warning.  I should look into how you did that,  might be useful to some of our other projects.

phlptp:
The basics of this PR are working through the details of isolating type size and expected size and allowing them to have a min and max independently.    In doing so this addresses #330,  #322,  and #296,  and gets at #257,  though I think with a little more work it might be possible to handle that without specifying the template arguments on add_option.  That will have to wait for a while though.  

There is also probably better treatment of #305 in this PR.  

There was some slightly odd behavior left in place to make sure it was consistent with previous behavior,  that might be able to be taken out in a 2.0 release if desired.

henryiii (approved):

Issue #326, opened 2019-10-01
non-const App::get_options


jbriales:
I was looking for how to iterate over all options (concretely, inside a option group), to set certain parameter of the option.
I found `get_options` but this is const-only.
Would it be possible to add a non-const version of `get_options` for an App?
I.e. just
`std::vector<Option *> get_options() { return options_; }`

phlptp:
My guess is this is another case of no one needing it so it was never added.  One other possibility would seem to match your use case would be an `apply` method that applies an operation to each contained option.

PR #329, opened 2019-10-05
Add non-const version of App::get_options

jbriales:
Addressing missing function as discussed in #326

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1545,6 +1545,22 @@ class App {
         return options;
     }
 
+    /// Non-const version of the above
+    std::vector<Option *> get_options(const std::function<bool(Option *)> filter = {}) {
+        std::vector<Option *> options(options_.size());
+        std::transform(std::begin(options_), std::end(options_), std::begin(options), [](const Option_p &val) {
+            return val.get();
+        });
+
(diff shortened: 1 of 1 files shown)


jbriales:
Any ideas/hints about the failures above?

phlptp:
The CI failures look like the same thing that is happening to the other PR's so I am assuming once those get fixed this can be rerun as well.

henryiii:

I'll do it if you need me to.

jbriales:
Seems it's all good now? :)

henryiii:
I'm surprised this still reports 100% coverage. Has a test started using this version, but not all of them? Odd. Regardless, I think it's fine.

henryiii:
Thanks!

phlptp:
Pretty sure a number of the tests that use that function would be based on non-const, but the actual formatters and several other functions use a const pointer , so they would still use the const version.    At some point a test probably should be added specifically to test the differences between the two versions.

jbriales:
I was also expecting test coverage would fail but I guess as you say the const vs non-const difference is subtle.
I created a new pull request to add the corresponding unit test in https://github.com/CLIUtils/CLI11/pull/337.

PR #337, opened 2019-10-26
Add test for non-const version of App::get_options

jbriales:
After https://github.com/CLIUtils/CLI11/pull/329 the get_options version being used in the test was the non-const.
Forced testing on non-const by using a const-ref to the app.
Also added testing to ensure the list of options of both const and non-const are the same as expected.


henryiii:
Thanks!

PR #340, opened 2019-11-01
Virtualized _parse_single and _parse_positional

KOLANICH:
It is needed to be able to substitute tham with own ones with custom logic that is run immediately when gets an arg.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -2425,7 +2425,7 @@ class App {
 
     /// Parse "one" argument (some may eat more than one), delegate to parent if fails, add to missing if missing
     /// from master return false if the parse has failed and needs to return to parent
-    bool _parse_single(std::vector<std::string> &args, bool &positional_only) {
+    virtual bool _parse_single(std::vector<std::string> &args, bool &positional_only) {
         bool retval = true;
         detail::Classifier classifier = positional_only ? detail::Classifier::NONE : _recognize(args.back());
         switch(classifier) {
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -2496,7 +2496,7 @@ class App {
     /// Parse a positional, go up the tree to check
     /// @param haltOnSubcommand if set to true the operation will not process subcommands merely return false
     /// Return true if the positional was used false otherwise
-    bool _parse_positional(std::vector<std::string> &args, bool haltOnSubcommand) {
+    virtual bool _parse_positional(std::vector<std::string> &args, bool haltOnSubcommand) {
 
         const std::string &positional = args.back();
 


KOLANICH:
Any progress  merging this?

phlptp:
I will let @henryiii make a decision on this but I am not sure I would recommend this.  Allowing those internal functions to be virtual automatically puts them into the public part of the API and they haven't been totally stable.  Modifying them could lead to all sorts of issues if the user functions didn't behave in the correct fashion or handle the arguments in the same way.  

I am just wondering if there might be another way of achieving your desired effect.  Perhaps @KOLANICH can expand on your use case and why the existing callbacks in the apps and options don't work for you.

KOLANICH:
>I am just wondering if there might be another way of achieving your desired effect. Perhaps @KOLANICH can expand on your use case and why the existing callbacks in the apps and options don't work for you.

In some parts there are, in some parts there are not. See

https://github.com/KOLANICH/kati/blob/CLI11/flags.cc

It is a rewrite of a plain C arguments parser (
https://github.com/google/kati/blob/master/flags.cc) into CLI11.  That parser mimics behavior of GNU Make in some part. That plain C parser has the following features:

* on an each arg code is immediately called
* that code modifies state
* based on that state args are either copied or not.
* there is no distinction between positional and keyed args.

So there are following ways to achieve these:
* modify CLI11 to have immediate callbacks for each argument
OR
* somehow preserve information about arguments full order and then recreate the copied command line based on this order.

But IMHO these all are too niche use cases to have code for them here. And noone wants to maintain an own fork. So the least invasive changes are to just make some methods virtual and allow users to implement the immediate callbacks mechanisms they want  by redefining them without of any guarantees of stability of these methods (I though expect breaking changes be very rare).

A better way is, of course, splitting `App` into multiple objects and letting a user to redefine each object.

henryiii:
I'm still in Australia and will try to consider this when I get back.

henryiii:
@phlptp What do you think? I'm not strongly against this; due to the leading underscore these functions should be considered unstable, but CLI11 tends to have a locked in version, and it's for a rather niche use case anyway. Since we already have a virtual table, I don't think there's a compile time or run time cost for adding this.

TL;DR: We can add it but it could break in the future and that's okay.

phlptp:
I think I am still against this.  I just think virtual functions automatically become part of the public interface whether they are private or not and should be stable through version numbers.  CLI11 hasn't followed semantic versioning in the past(maybe it should after 2.0) but maintaining stability in the public interface is still important and probably more so as more people use it.    longer term I would even be of the opinion to work toward making App non-virtual completely.  

still up to you though.

henryiii:
> work toward making App non-virtual completely

This would solve the move issue as well. Okay, let's not push for this, but rather try to solve the issues as they come up. It's very easy to maintain this small patch (two lines changed) if someone does want this badly enough.

PR #342, opened 2019-11-12
Handle float type in positive number (CLIUtils#328) 

ChristosT:
This PR enables checking positivity of float types , providing a fix to  CLIUtils#328.
Numbers are handled as `double`s in the same way that `CLI:Number` is structured https://github.com/CLIUtils/CLI11/blob/418b7175f5223aa28ab65744c76132fd05fd8f39/include/CLI/Validators.hpp#L364-L369
Also, a`CLI:NonNegative` is added and checks if  `number <=0` while `CLI:Positive` is modified to check whether `number>0`


henryiii (approved):

henryiii:
Thanks!

Issue #339, opened 2019-10-31
Stream operator o ostream breaks enum conversion


sobczyk:
Test code:
```.cpp
#include <CLI/CLI.hpp>
  
enum class Level : int { High, Medium, Low };

// this breaks the code
inline std::ostream& operator<<(std::ostream& o, const Level& l)
{
        switch(l) {
                case Level::High:
                        o << "High";
                        break;
                case Level::Medium:
                        o << "Medium";
                        break;
                case Level::Low:
                        o << "Low";
                        break;
        }
}

int main(int argc, char **argv) {
    CLI::App app;

    Level level;
    // specify string->value mappings
    std::vector<std::pair<std::string, Level>> map{
        {"high", Level::High}, {"medium", Level::Medium}, {"low", Level::Low}};
    // checked Transform does the translation and checks the results are either in one of the strings or one of the
    // translations already
    app.add_option("-l,--level", level, "Level settings")
        ->required()
        ->transform(CLI::CheckedTransformer(map, CLI::ignore_case));

    CLI11_PARSE(app, argc, argv);

    // CLI11's built in enum streaming can be used outside CLI11 like this:
    using namespace CLI::enums;
    std::cout << "Enum received: " << level << std::endl;

    return 0;
}

```
Running:
```
./a.out --help
Usage: ./a.out [OPTIONS]                                                                                                                                                                                           
                                                                                                                                                                                                                   
Options:                                                                                                                                                                                                           
  -h,--help                   Print this help message and exit                                                                                                                                                     
  -l,--level ENUM:value in {high->High,medium->Medium,low->Low} OR {High,Medium,Low} REQUIRED                                                                                                                      
                              Level settings

./a.out --level medium
Could not convert: --level = Medium
Run with --help for more information.
```

The function `template <typename T> inline CLI::detail::as_string(const T &v)`
uses the available ostream printer.
An option would be to add as_string for enums only.

It would be also nice if the help printer used the original ostream redirection, so we gen the real enum name.

sobczyk:
Another workaround would be to omit the key -> value -> string conversion, and use the key -> value from mapping directly

phlptp:
I don't think it would possible to not use the key->value->string since the value may require a different conversion when run through the rest of the processing,  trying to store an intermediate "value" of a different type would introduce all sorts of complexity and complications from transform ordering to the callback functions.    If a user wanted to do that it could be done inside a callback function though.

PR #348, opened 2019-11-26
Issue #339

phlptp:
add an example where the name of the enum is printed through a stream output function, which subverted the checkTransformer conversion and prevented conversion of the enumeration.

This PR deals with this situation for enumerations and other numeric types.    It does it by discriminating between to_string situations in which a value is needed vs one where the converted string is needed, by adding a value_string overload which routes to the to_string unless the value is numeric or an enumeration.


phlptp:
Fix Issue #339

henryiii:
Great, thanks!

PR #349, opened 2019-11-26
allow transform to work with arrays.  

phlptp:
The element_type trait require pointer_traits<T> to be defined for the type which it was for vector but not for array due to decay rules in C++, even though it was not used.  So Element type had to be split into two templates instead of a conditional.

Fix Gitter chat issue [November 25, 2019 10:09 AM](https://gitter.im/CLI11gitter/Lobby?at=5ddc18c155bbed7ade433f5b)


henryiii:
Is this ready, or were you planning on a couple more tests (from gitter)?

phlptp:
I probably should add a test with `const char *` arrays yet.  I will get to that this weekend

phlptp:
I think this is ready now.

henryiii (approved):

PR #357, opened 2019-11-30
Readme: Fix optional paragraph

cbachhuber:
- Added backticks around e.g. optional<int>, such that the angle brackets are rendered in markdown
- Changed link for further info to the suitable chapter, as I did not find info on optional or templates in the internals chapter
- std::optional<T> could be used as well, correct? Added to this paragraph


henryiii:
Thanks!

PR #358, opened 2019-12-01
Deprecated retired options

### LINKED ISSUE #275: Feature request: mark a flag as deprecated or retired
While maintaining a program that accepts CLI flags, one often wants to deprecate an option - still leaving it functional but providing a warning to the user, or retiring an option, meaning the flag is consumed but otherwise ignored.

Any thoughts on what that functionality could look inside CLI11?

phlptp:
adds helper functions for retired and deprecated options

Fixes #275

### DIFF
--- a/examples/retired.cpp
+++ b/examples/retired.cpp
@@ -17,5 +17,5 @@ int main(int argc, char **argv) {
 
     CLI::deprecate_option(opt2, "--not_deprecated");
 
-   CLI11_PARSE(app, argc, argv);
+    CLI11_PARSE(app, argc, argv);
 }
--- a/examples/retired.cpp
+++ b/examples/retired.cpp
@@ -1,6 +1,7 @@
 #include "CLI/CLI.hpp"
 #include <vector>
 
+// This example show the usage of the retired and deprecated option helper methods
 int main(int argc, char **argv) {
 
     CLI::App app("example for retired/deprecated options");
@@ -12,9 +13,14 @@ int main(int argc, char **argv) {
 
     app.add_option("--not_deprecated", x);
 
+    // specify that a no existing option is retired
     CLI::retire_option(app, "--retired_option");
+
+    // specify that an existing option is retired and non-functional this will replace the option with another that
+    // behaves the same but does nothing
     CLI::retire_option(app, opt1);
 
+    // deprecate an existing option and specify a recommended replacement
     CLI::deprecate_option(opt2, "--not_deprecated");
 
     CLI11_PARSE(app, argc, argv);
--- a/examples/retired.cpp
+++ b/examples/retired.cpp
@@ -24,4 +24,15 @@ int main(int argc, char **argv) {
     CLI::deprecate_option(opt2, "--not_deprecated");
 
     CLI11_PARSE(app, argc, argv);
+
+    if(!x.empty()) {
+        std::cout << "Retired option example: got --not_deprecated values:";
+        for(auto &xval : x) {
+            std::cout << xval << " ";
+        }
+        std::cout << '\n';
+    } else if(app.count_all() == 1) {
+        std::cout << "Retired option example: no arguments received\n";
+    }
+    return 0;
 }
--- a/examples/retired.cpp
+++ b/examples/retired.cpp
@@ -1,0 +1,21 @@
+#include "CLI/CLI.hpp"
+#include <vector>
+
+int main(int argc, char **argv) {
+
+    CLI::App app("example for retired/deprecated options");
+    std::vector<int> x;
+    auto opt1 = app.add_option("--retired_option2", x);
+
+    std::pair<int, int> y;
+    auto opt2 = app.add_option("--deprecate", y);
+
+    app.add_option("--not_deprecated", x);
+
+    CLI::retire_option(app, "--retired_option");
+    CLI::retire_option(app, opt1);
+
+    CLI::deprecate_option(opt2, "--not_deprecated");
+
+   CLI11_PARSE(app, argc, argv);
+}
--- a/examples/retired.cpp
+++ b/examples/retired.cpp
@@ -1,7 +1,7 @@
 #include "CLI/CLI.hpp"
 #include <vector>
 
-// This example show the usage of the retired and deprecated option helper methods
+// This example shows the usage of the retired and deprecated option helper methods
 int main(int argc, char **argv) {
 
     CLI::App app("example for retired/deprecated options");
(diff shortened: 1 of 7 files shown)


phlptp:
TODO
- [x] add documentation
- [x] add some comments and explanation to the example
- [x] verify code coverage
- [x] rename a few variable
- [ ] verify coverage

examples/retired.cpp
```diff
@@ -0,0 +1,27 @@
+#include "CLI/CLI.hpp"
+#include <vector>
+
+// This example show the usage of the retired and deprecated option helper methods
```
cbachhuber:
```suggestion
// This example shows the usage of the retired and deprecated option helper methods
```

examples/retired.cpp
```diff
@@ -0,0 +1,27 @@
...
+    CLI::App app("example for retired/deprecated options");
+    std::vector<int> x;
+    auto opt1 = app.add_option("--retired_option2", x);
+
+    std::pair<int, int> y;
+    auto opt2 = app.add_option("--deprecate", y);
+
+    app.add_option("--not_deprecated", x);
+
+    // specify that a no existing option is retired
```
cbachhuber:
```suggestion
    // specify that a non-existing option is retired
```

examples/retired.cpp
```diff
@@ -0,0 +1,27 @@
...
+    CLI::retire_option(app, "--retired_option");
+
+    // specify that an existing option is retired and non-functional this will replace the option with another that
```
cbachhuber:
```suggestion
    // specify that an existing option is retired and non-functional: this will replace the option with another that
```

examples/retired.cpp
```diff
@@ -0,0 +1,27 @@
...
+    // behaves the same but does nothing
+    CLI::retire_option(app, opt1);
+
+    // deprecate an existing option and specify a recommended replacement
```
cbachhuber:
```suggestion
    // deprecate an existing option and specify the recommended replacement
```

examples/retired.cpp
```diff
@@ -0,0 +1,27 @@
...
+    CLI::deprecate_option(opt2, "--not_deprecated");
+
+    CLI11_PARSE(app, argc, argv);
```
cbachhuber, line 26: Many other examples do a std::cout at the end, to show what is going on. Does that make sense here as well?

include/CLI/App.hpp
```diff
@@ -3029,6 +3029,79 @@ inline void TriggerOff(App *trigger_app, std::vector<App *> apps_to_enable) {
...
+    auto opt = app->get_option_no_throw(option_name);
+    if(opt != nullptr) {
+        retire_option(app, opt);
+        return;
+    }
+    auto opt2 = app->add_option(option_name, "option has been retired and has no effect")
+                    ->type_name("RETIRED")
+                    ->expected(0, 1)
+                    ->default_str("RETIRED");
+    Validator cv{[opt2](std::string &) {
```
cbachhuber: Why is this object called cv? Could you find a more epressive name?

include/CLI/App.hpp
```diff
@@ -3029,6 +3029,79 @@ inline void TriggerOff(App *trigger_app, std::vector<App *> apps_to_enable) {
...
+                 },
+                 ""};
+    cv.application_index(0);
+    opt2->check(cv);
+}
+
+/// Helper function to mark an option as retired
+inline void retire_option(App &app, Option *opt) { retire_option(&app, opt); }
+
+/// Helper function to mark an option as deprecated
```
cbachhuber:
```suggestion
/// Helper function to mark an option as retired
```

include/CLI/App.hpp
```diff
@@ -3029,6 +3029,79 @@ inline void TriggerOff(App *trigger_app, std::vector<App *> apps_to_enable) {
...
+    Validator cv{[opt2](std::string &) {
+                     std::cout << "WARNING " << opt2->get_name() << " is retired\n";
+                     return std::string();
+                 },
+                 ""};
+    cv.application_index(0);
+    opt2->check(cv);
+}
+
+/// Helper function to mark an option as deprecated
```
cbachhuber:
```suggestion
/// Helper function to mark an option as retired
```

include/CLI/App.hpp
```diff
@@ -3029,6 +3029,79 @@ inline void TriggerOff(App *trigger_app, std::vector<App *> apps_to_enable) {
...
+inline void deprecate_option(App *app, const std::string &option_name, const std::string &replacement = "") {
+    auto opt = app->get_option(option_name);
+    deprecate_option(opt, replacement);
+}
+
+/// Helper function to mark an option as deprecated
+inline void deprecate_option(App &app, const std::string &option_name, const std::string &replacement = "") {
```
cbachhuber, line 3054: My IDE complains that this function is never used. Is my IDE wrong? Manually, I also can't find usages and wonder how you get 100% coverage.

include/CLI/App.hpp
```diff
@@ -3029,6 +3029,79 @@ inline void TriggerOff(App *trigger_app, std::vector<App *> apps_to_enable) {
...
+    auto opt = app.get_option(option_name);
+    deprecate_option(opt, replacement);
+}
+
+/// Helper function to mark an option as retired
+inline void retire_option(App *app, Option *opt) {
+    App temp;
+    auto ropt = temp.add_option(opt->get_name(false, true))
```
cbachhuber: Why do you need ropt in addition to opt? Consider a more descriptive name for this object

cbachhuber (changes requested):
Thank you for this addition! Looks very good already, I only have minor remarks.

examples/retired.cpp
phlptp: Maybe something that prints out if the '--not_deprecated` or no options were passed, just so there is always guaranteed to be some output?

include/CLI/App.hpp
phlptp: I thought I was using it one of the tests but I will double check that and make sure

include/CLI/App.hpp
phlptp: We need to create a temporary psuedo copy of the option before destroying the original, and then copying the copy with the critical info.    I will rename it something more descriptive though.

include/CLI/App.hpp
phlptp: `custom validator`,  I will rename it

examples/retired.cpp
cbachhuber: Good suggestion, I like it!

include/CLI/App.hpp
cbachhuber: Thanks!

cbachhuber (approved):
Looks good, thanks! :+1:

henryiii (approved):

henryiii:
Looks good to me, are you ready to merge?

phlptp:
I think it is ready.

PR #360, opened 2019-12-03
Clang-tidy fixes

cbachhuber:
This PR fixes the clang-tidy issues in Option.hpp.

### DIFF
--- a/include/CLI/Option.hpp
+++ b/include/CLI/Option.hpp
@@ -87,7 +87,7 @@ template <typename CRTP> class OptionBase {
     // setters
 
     /// Changes the group membership
-    CRTP *group(std::string name) {
+    CRTP *group(const std::string& name) {
         group_ = name;
         return static_cast<CRTP *>(this);
     }
@@ -346,7 +346,7 @@ class Option : public OptionBase<Option> {
     bool empty() const { return results_.empty(); }
 
     /// This class is true if option is passed.
-    operator bool() const { return !empty(); }
+    explicit operator bool() const { return !empty(); }
 
     /// Clear the parsed results (mostly for testing)
     void clear() {
@@ -409,7 +409,7 @@ class Option : public OptionBase<Option> {
     bool get_allow_extra_args() const { return allow_extra_args_; }
 
     /// Adds a Validator with a built in type name
-    Option *check(Validator validator, std::string validator_name = "") {
+    Option *check(Validator validator, const std::string& validator_name = "") {
         validator.non_modifying();
         validators_.push_back(std::move(validator));
         if(!validator_name.empty())
@@ -418,7 +418,7 @@ class Option : public OptionBase<Option> {
     }
 
     /// Adds a Validator. Takes a const string& and returns an error message (empty if conversion/check is okay).
-    Option *check(std::function<std::string(const std::string &)> Validator,
+    Option *check(const std::function<std::string(const std::string &)>& Validator,
                   std::string Validator_description = "",
                   std::string Validator_name = "") {
         validators_.emplace_back(Validator, std::move(Validator_description), std::move(Validator_name));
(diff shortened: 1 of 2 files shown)


include/CLI/Option.hpp
```diff
@@ -418,7 +418,7 @@ class Option : public OptionBase<Option> {
     }
 
     /// Adds a Validator. Takes a const string& and returns an error message (empty if conversion/check is okay).
-    Option *check(std::function<std::string(const std::string &)> Validator,
+    Option *check(const std::function<std::string(const std::string &)> &Validator,
                   std::string Validator_description = "",
                   std::string Validator_name = "") {
         validators_.emplace_back(Validator, std::move(Validator_description), std::move(Validator_name));
```
phlptp: wonder if Validator should be moved here instead of copied, same with a few functions down.  Whatever is done it should probably be consistent on all functions

include/CLI/Option.hpp
```diff
@@ -663,13 +663,13 @@ class Option : public OptionBase<Option> {
     callback_t get_callback() const { return callback_; }
 
     /// Get the long names
-    const std::vector<std::string> get_lnames() const { return lnames_; }
```
phlptp, line 666: This might have been intended to be a const reference  `const std::vector<std::string> &`

cbachhuber: Good point, fixed. Also for get_snames(), but not for the remaining getters, as that causes CI to fail

include/CLI/Option.hpp
cbachhuber: I agree that this should be consistent. Reverted in this PR, maybe this should be done in a separate PR

include/CLI/Option.hpp
```diff
@@ -427,15 +427,15 @@ class Option : public OptionBase<Option> {
...
+    Option *transform(Validator Validator, const std::string &Validator_name = "") {
         validators_.insert(validators_.begin(), std::move(Validator));
         if(!Validator_name.empty())
             validators_.front().name(Validator_name);
         return this;
     }
 
     /// Adds a Validator-like function that can change result
-    Option *transform(std::function<std::string(std::string)> func,
+    Option *transform(const std::function<std::string(std::string)> &func,
```
phlptp, line 438: I have gone back and forth on whether std::function should be taken by value and moved or taken by constref and copied.  I think I usually default to take by value since a very common usages is to give a lambda function in the argument, but I am not entirely sure what is best here.

phlptp: actually in this case constref is probably best since we have to do a copy in the internal lambda (at least until C++14 is the minimum mode in CLI11)

include/CLI/Option.hpp
```diff
@@ -663,13 +663,13 @@ class Option : public OptionBase<Option> {
...
-    const std::vector<std::string> get_lnames() const { return lnames_; }
+    const std::vector<std::string> &get_lnames() const { return lnames_; }
 
     /// Get the short names
-    const std::vector<std::string> get_snames() const { return snames_; }
+    const std::vector<std::string> &get_snames() const { return snames_; }
 
     /// Get the flag names with specified default values
-    const std::vector<std::string> get_fnames() const { return fnames_; }
+    std::vector<std::string> get_fnames() const { return fnames_; }
```
phlptp: should this be constref return as well?

henryiii:
The CI should be checking clang-tidy (though it might have been lost during the transition of formatting to GitHub Actions).

phlptp:
I remember getting some PR's where the clang-tidy checks were triggered recently so I think it is still working, but I was also under the impression that was only checking a pretty limited subset of the clang-tidy checks and specifically ones where there was an auto fix that was possible.

include/CLI/Option.hpp
cbachhuber: Probably you're right. It won't make a huge difference I guess. And with this PR it is improved compared to before, so I suggest keeping it.

include/CLI/Option.hpp
cbachhuber: Right, implemented.

phlptp (approved):

PR #362, opened 2019-12-06
config file handling refactor

### LINKED ISSUE #214: Sort options in groups in INI output
To be consistent with --help output.

### LINKED ISSUE #293: Get loaded config file name
For logging reasons, I want to retrieve the loaded config filename. One solution would been to have a access to the `config_name_` variable but even then, if the default filename wasn't accessible, I wouldn't know (if the parameter `config_required` was not set).

Currently I have the following hack to get that information but it's pretty ugly:
```c++
    std::shared_ptr<CLI::App> _app = std::make_shared<CLI::App>("App");
    std::string _config_name = "./config.ini";
    _app->set_config("--config",
                   _config_name,
                   "Read a config file",
                   false);
    _config_name = (_app->get_option("--config")->results().empty())?_config_name:_app->get_option("--config")->results().at(0);
    std::ifstream test{_config_name};
    _config_name = (!test.good())?"None":_config_name;
```
It would be nice to have a nicer way to do that.

### LINKED ISSUE #335: Ignore superfluous options in INI
I don't think there currently is a way to ignore superfluous options in an INI file, that is options that are set within the INI, but are not used by the program. 

Use case:

I am writing an image encoder/decoder combo, which both have slightly different option requirements, but it would be useful if they could use the same INI files.

### LINKED ISSUE #345: _parse stays zero when parsing subcommands from INI
I use an INI file to configure my app and cmd line arguments to overwrite some of it's properties. 

Trying to figure out which options are set from the INI, by not passing any arguments from the cmd line,  I realized that ```app.got_subcommnad(sApp)``` is always ```false```. 

Additionally, the ```->required()``` flag that I set to some of the sub-commnads options is ignored at the INI level.

I temporarily fixed the issue by extending ```_parse_single_config``` to call ```->increment_parsed()``` for the sub-commands, i.e., 

```
    /// Fill in a single config option
    bool _parse_single_config(const ConfigItem &item, size_t level = 0) {
        if(level < item.parents.size()) {
            try {
                auto subcom = get_subcommand(item.parents.at(level));

               //TODO tmp fix
                subcom->increment_parsed();


                return subcom->_parse_single_config(item, level + 1);
            } catch(const OptionNotFound &) {
                return false;
            }
        }
``` 
and appears to resolve my issues. I don't really have an overview of the arch to figure out whether there are side effects by doing so. Any suggestions?

### LINKED ISSUE #364: Issues in saving/loading config with flag defaults
When saving config with default values, i.e. `app.config_to_str(true, true)`, flags get saved as false.
However when reading that same config, the flags get evaluated to true.

I created a toy test in https://github.com/CLIUtils/CLI11/pull/363 to quickly showcase this.

@phlptp  this might be (maybe) addressed in your WIP (https://github.com/CLIUtils/CLI11/pull/362)?

phlptp:
This PR if merged is an an overhaul of the configuration file system.   It should be backwards compatible with the existing way things worked modulo actual bug fixes.   The only default change was making the default to ignore extras in a configuration file.

This PR was driven by
 -  Fixing some of the issues that have come up over the last few months.  #335, #293, #364.  
-   An ability to actually trigger subcommand from config files so they can act like scripts #345
-  Support for TOML files
-  Clean up output #214 

The big changes
-  a configurable option in apps to allow them to be triggered by config files
-  Generalize the config-formatter with customization points for some characters like, the vector notation, the value separator, and comment character.  (to easily allow things like some Windows config files which use ':' as a value separator in the ini files.)  
-  Generalize the default reader to handle most TOML files (not embededed inline tables)  
-  predefine a TOML output module
-  support option groups and grouping in the config output generator 

refactor some of the configuration file handling code.  Make it easier to get the actual file that was processed, and allow extras in the config file to be ignored(default now), captured or errored.

- [x] Fix #335 
- [x] Fix #293 
- [x] Fix #345 
- [x] Fix #364
- [x] Fix the other issues created from fixing #364
- [x] get complete code coverage
- [x] generalized subcommand support in config files
- [x] update readme
- [x] update docbook
- [x] look at addressing #214
- [x] handle option groups correctly
- [x] configurable subcommands using `[subname]` notation
- [x] finalize code formatting

For handing of unrecognized options in ini files there are 3 options,  ignore(now default), capture(forward to app missing), and error.   The function specifying this can take a boolean(same behavior as before) or an enum.  The default behavior has changed but the behavior of the function calls is the same as before.

Closes #363 (test included). Closes #214.


phlptp:
I think this is ready for a review.

henryiii:
This is fantastic! I'll try to review tonight.

book/chapters/config.md
```diff
@@ -2,15 +2,49 @@
...
-is not found and this is set to true.
+You can tell your app to allow configure files with `set_config("--config")`. There are arguments: the first is the option name. If empty, it will clear the config flag. The second item is the default file name. If that is specified, the config will try to read that file. The third item is the help string, with a reasonable default, and the final argument is a boolean (default: false) that indicates that the configuration file is required and an error will be thrown if the file is not found and this is set to true.
+
+### Extra fields
+Sometimes configuration files are used for multiple purposes so CLI11 allows options on how to deal with extra fields
+
+```cpp
+app.allow_config_extras(true);
+```
+will allow capture the extras in the extras field of the app. (NOTE:  This also sets the allow_extras in the app to true)
```
henryiii:
```suggestion
will allow capture the extras in the extras field of the app. (NOTE:  This also sets the `allow_extras` in the app to true)
```

book/chapters/config.md
```diff
@@ -2,15 +2,49 @@
...
+
+for slightly finer control there is a scoped enumeration of the modes
+or
+```cpp
+app.allow_config_extras(CLI::config_extras_mode::ignore);
+```
+will completely ignore extras and is the default.
+
+```cpp
+app.allow_config_extras(CLI::config_extras_mode::capture);
```
henryiii, line 28: Is this equivalent to true? Below you mention the one that is equivalent to false, so it might be nice to mention that this is equivalent to true, as well.

book/chapters/config.md
```diff
@@ -23,12 +57,66 @@ in_subcommand = Wow
...
+vector = [1,2,3]
+
+# Section map to subcommands
+[subcommand]
+in_subcommand = Wow
+[subcommand.sub]
+subcommand = true # could also be give as sub.subcommand=true
+```
+
+The main differences are in vector notation and command character.  Note CLI11 is not a full TOML parser as it just reads values as strings.  It is possible (but not recommended to mix notation)
```
henryiii:
```suggestion
The main differences are in vector notation and comment character.  Note: CLI11 is not a full TOML parser as it just reads values as strings.  It is possible (but not recommended) to mix notation.
```

book/chapters/config.md
```diff
@@ -23,12 +57,66 @@ in_subcommand = Wow
...
 
 ## Writing out a configure file
 
 To print a configuration file from the passed arguments, use `.config_to_str(default_also=false, prefix="", write_description=false)`, where `default_also` will also show any defaulted arguments, `prefix` will add a prefix, and `write_description` will include option descriptions.
 
+### Customization of configure file output
+The default config parser/generator has some customization points that allow variations on the INI format.  The default formatter has a base configuration that matches the INI format.  It defines 5 characters that define how different aspects of the configuration are handled
+```cpp
+/// the character used for comments
+    char commentChar = ';';
```
henryiii: Could you fix the odd indentation here on the these lines?

henryiii (approved):
A few minor touchups recommended, otherwise it looks great.

Would it make sense to change the default in 2.0 to read and write TOML files, with INI being an option that has to be loaded? The formats are similar, and TOML is better defined than INI. I think vectors were not part of any INI format per say, and I came up with the current space separated method. So really the only difference is the comment character.

book/chapters/config.md
phlptp: It is not exactly equivalent.  For backwards compatibility setting to true also sets the allow_extras flag,  So it adds some coupling between the flags which may or may not be desired.  Using the enumeration doesn't have any coupling.

book/chapters/config.md
phlptp: yes

phlptp:
I think it would make sense to make the write output TOML compliant in 2.0.  I think as long as we leave the read side capable of reading both to the extent possible.  But switching the write to TOML at some point makes a lot of sense.

book/chapters/config.md
henryiii: Maybe adding this note would be useful, then. At least somewhere (in doxygen comment, for example).

phlptp:
I think I would recommend a 1.9 release before making the switch though.

henryiii:
I think we are close to a 1.9 release. I want to push that out soon then work on the precompile mode and Exception-free option, which are (my) two wishlist items for 2.0.

henryiii:
Should have merged this before #382. Will fix issues and merge.

phlptp:
I will work on #371 in the next day or so.  That would be the last thing I want to deal with for the near term that effects our uses.  If we can get that in before 1.9 then I use the main repo version in our code base instead of a customized one.  I want to do some work on the book after that but I think that should be mostly separate from the major refactoring you are thinking of.

henryiii:
Sounds good. Yes, that should not be touched by the restructure.

henryiii:
Thanks!!!

PR #374, opened 2019-12-27
add a cmakeLists for the book 

phlptp:
If merged this pull request will add a target in MSVC to link the book into the visual studio UI so it can be edited easily inside visual Studio.  

adds a CMakeLists in the book directory and an add_subdirectory command to use it inside the main CMakeLists.txt file.  

This doesn't actually build anything new just makes it much easier to edit the book in visual studio.

### DIFF
--- a/book/CMakeLists.txt
+++ b/book/CMakeLists.txt
@@ -1,0 +1,24 @@
+
+if(MSVC)
+
+    set(
+        book_sources
+        README.md
+        SUMMARY.md
+        chapters/advanced-topics.md
+        chapters/an-advanced-example.md
+        chapters/basics.md
+        chapters/config.md
+        chapters/flags.md
+        chapters/formatting.md
+        chapters/installation.md
+        chapters/internals.md
+        chapters/options.md
+        chapters/subcommands.md
+        chapters/toolkits.md
+        chapters/validators.md
+    )
+
+    add_custom_target(cli_book SOURCES ${book_sources})
+
+endif(MSVC)
--- a/book/CMakeLists.txt
+++ b/book/CMakeLists.txt
@@ -1,24 +1,10 @@
 
-if(MSVC)
+set(
+    book_sources
+    README.md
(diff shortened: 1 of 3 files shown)


phlptp:
I was starting to work with this as part of the config_file update but it seemed like something that could stand alone as it wasn't directly related to any config file changes.

book/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+
+if(MSVC)
```
henryiii: Is there a reason to protect it like this? Non-IDEs only build if ALL is set and other IDEs like Xcode May benefit too.

henryiii (commented):
What about using a glob? I don’t think I’d remember to add a new md file to the cmake lists.

book/CMakeLists.txt
phlptp: generally not needed for most linux builds so in our other applications we just protect things like this, to not add overhead on CMAKE that isn't necessary.  I took it off.

phlptp: As for using a glob, I generally avoid it.  it can get a little annoying when you are adding things and cmake won't regenerate the project, whereas with a listing of all files that isn't a problem.   I made a it a glob but left a list of the readme and summary so if more are added they can be added in there.

phlptp:
also it appears a new doxygen release was made (12-27-19) so the build script for doxygen broke

book/CMakeLists.txt
henryiii: Glob should not be used for the main build system, but if CMake is a secondary build system (such as for a Python package that has setup.py as the primary system, or for an extra feature like this one), then it's quite useful - as it is easy to forget to keep changes updated everywhere. CONFIGURE_DEPENDS really helps make it more useful, but since docs is a "fake" target, there is no reason to expect a "--build" should also check this and reconfigure.

henryiii (approved):

PR #377, opened 2019-12-30
Minor fixes in enum examples

cbachhuber:
- Initialize level variable (uninitialized before)
- Use std::map instead of std::vector<std::pair>
- Update comments
- Add enum_ostream to readme


henryiii (approved):

henryiii:
Thanks!

PR #379, opened 2019-12-30
All-Contributors

henryiii:
Adding bot & all-contributors list. Should be easier to maintain and check.


henryiii:
And thanks to @p-ranav for giving me the idea to do this. ;)

Issue #371, opened 2019-12-16
potential issues with default_val call


phlptp:
In looking at #369.  There are a few potential issues with using that function.  

It actually evaluates the callback, which works perfectly fine if CLI11 is assigning to a variable or the callback is empty.  In that context that is probably the correct thing to do.    But if the callback was user specified this could result in very unexpected behavior such as repeat callbacks or oddly timed callbacks if there was user defined dependencies.  

I am wondering if there needs to be a flag set up in the option definition that can discriminate between some of the CLI11 defined callbacks and user defined one and the function callback, so those user defined callbacks are not called at unexpected times.  

Also wondering if default_val should be a template that can take any type(that can be converted to a string), so if you had an option taking an integer you could do 
```
opt->default_val(5);
```

henryiii:
We could add such a flag.

Yes, we could add a template to that function. I originally did not because I thought we were going to add a template to Option at some point; I'm now rather thinking we will not, since by avoiding (public) templated classes we can precompile and same some time.

PR #387, opened 2020-01-01
default_val option call

phlptp:
Fix potentially improper callback calls for default_val Option function.  the update adds a flag variable to control it, makes default_val exception safe and a template to convert from actual value types.

add a variant of is_vector to check for const std::vector as well.
add a few tests for different modes of operation

Addresses #371

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,8 +331,8 @@ Before parsing, you can set the following options:
     `->capture_default_str()`: 🆕 Store the current value attached and display it in the help string.
 -   `->default_function(std::string())`: 🆕 Advanced: Change the function that `capture_default_str()` uses.
 -   `->always_capture_default()`: 🆕 Always run `capture_default_str()` when creating new options. Only useful on an App's `option_defaults`.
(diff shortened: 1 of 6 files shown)


phlptp:
- [x] Need to update readme yet before merging.  
- [x] Verify coverage

README.md
```diff
@@ -331,6 +331,8 @@ Before parsing, you can set the following options:
     `->capture_default_str()`: 🆕 Store the current value attached and display it in the help string.
 -   `->default_function(std::string())`: 🆕 Advanced: Change the function that `capture_default_str()` uses.
 -   `->always_capture_default()`: 🆕 Always run `capture_default_str()` when creating new options. Only useful on an App's `option_defaults`.
+-   `default_str()`:  Set the default string directly.  This string will also be used as a default value if no arguments are passed and the value is requested.
```
```suggestion
-   `default_str(string)`:  Set the default string directly.  This string will also be used as a default value if no arguments are passed and the value is requested.
```

henryiii (approved):
Looks good!

PR #394, opened 2020-01-08
Adding CMake 3.4 check

henryiii:
Reimplements and closes #321. Changes:

* `CLI11_CXX_STD` -> `CMAKE_CXX_STANDARD` (only changed if set)
* All versions of CMake checked to verify they configure (3.4 - 3.16)
* Sanitizers and json now are downloaded if requested, much smaller recursive git checkout
* Uses generator expressions (maybe too?) heavily
* `CLI11_BUILD_EXAMPLES_JSON` asks for json example
* As a nested project CLI11 should behave nicer


henryiii:
@all-contributors please add @slurps-mad-rips for packaging

henryiii:
@all-contributors please add @slurps-mad-rips for platform work

.github/workflows/tests.yml
```diff
@@ -28,14 +28,88 @@ jobs:
...
+    - name: Add wget
+      run: apt-get update && apt-get install -y wget
+    - name: Install Modern CMake
+      run: wget -qO- "https://cmake.org/files/v3.16/cmake-3.16.0-Linux-x86_64.tar.gz" | tar --strip-components=1 -xz -C /usr/local
+    - name: Configure
+      run: cmake -S . -B build -DCLI11_CUDA_TESTS=ON
+    - name: Build
+      run: cmake --build build -j2
+
+  cmake-config:
```
phlptp, line 43: how long does this take to run on all versions of cmake?

phlptp:
I think I want to try this out on my windows system tonight just to make sure it all works correctly.

  I like most of the changes,  I agree the heavy use of generator expressions may or may not be desirable.

.github/workflows/tests.yml
henryiii: About 6 seconds each. Plus a little for the initial docker build, total is under 2 minutes.

Screen Shot 2020-01-08 at 8 31 29 PM

henryiii:
> I think I want to try this out on my windows system tonight just to make sure it all works correctly.

Please do, most importantly, I don't want to break anything.

phlptp:
If the CLI11_INSTALL  is set to OFF I get
```
Selecting Windows SDK version 10.0.18362.0 to target Windows 10.0.18363.
CMake 3.15.2
Configuring done
CMake Error: INSTALL(EXPORT) given unknown export "CLI11Targets"
Generating done
```

I am guessing somewhere something is not acknowledging that flag

phlptp:
Ok, ran it through a bunch of different configurations on windows and MSYS.
There are a few warnings on some of the tests in Visual studio with C++17 mode enabled,  not ones that are easily fixable right now.  Everything else checks out as far as I can tell.  

The CLI11_INSTALL=OFF error should be fixed or the option removed before merging though.  

Longer term we may want to consider making the gtest options hidden so they don't even show up on the cmake-gui interface.

I think there might be a few things that need tweaking once CLI11 moves to have a compiled component(or at least the option of one)

phlptp (changes requested):
The CLI11_INSTALL=OFF error should be fixed or the option removed.

bruxisma:
> The CLI11_INSTALL=OFF error should be fixed or the option removed.

For what its worth, older versions of CMake always add their items to the `install` target when used as a subdirectory, unless `EXCLUDE_FROM_ALL` is used when `add_subdirectory` is called.

(Also, sorry for dropping off the face of the earth for a bit. Life has been getting in the way. Thanks @henryiii for taking care of this ☺️)

henryiii:
FWIW, macOS 10.14 + C++17 mode is broken:

```
In file included from /Users/runner/runners/2.163.1/work/1/s/tests/informational.cpp:4:
In file included from /Users/runner/runners/2.163.1/work/1/s/include/CLI/CLI.hpp:23:
/Users/runner/runners/2.163.1/work/1/s/include/CLI/Validators.hpp:276:34: error: 'status' is unavailable: introduced in macOS 10.15
    auto stat = std::filesystem::status(file, ec);
```

And the warnings on Windows for C++17 mode. Not related to this PR, though.

phlptp (approved):

Issue #396, opened 2020-01-09
C++17 mode


henryiii:
We have a warning or two in C++17 mode on windows, and macOS 10.14 does not support C++17 mode - we might be able to get this to pass by checking a feature macro instead of C++ mode for filesystem. See #394.

phlptp:
Here is the visual studio warning
```
45>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\optional(65,17): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message : see reference to function template instantiation 'std::_Optional_destruct_base<_Ty,true>::_Optional_destruct_base<double&>(std::in_place_t,double &)' being compiled
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :         with
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :         [
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :             _Ty=int
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :         ]
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message : see reference to function template instantiation 'std::_Optional_destruct_base<_Ty,true>::_Optional_destruct_base<double&>(std::in_place_t,double &)' being compiled
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :         with
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :         [
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :             _Ty=int
45>C:\Users\phlpt\Documents\CLI11\tests\OptionalTest.cpp(215): message :         ]
45>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\optional(201): message : see reference to function template instantiation 'std::_Optional_construct_base<_Ty>::_Optional_construct_base<double&>(std::in_place_t,double &)' being compiled
45>C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\VC\Tools\MSVC\14.23.28105\include\optional(201): message :         with
[...]

PR #397, opened 2020-01-09
Add std checking and try to fix filesystem being used on macOS 10.14

henryiii:
Addressing #396.

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -2,6 +2,7 @@
 // Distributed under the 3-Clause BSD License.  See accompanying
 // file LICENSE or https://github.com/CLIUtils/CLI11 for details.
 
+#include "CLI/Macros.hpp"
 #include "CLI/StringTools.hpp"
 #include "CLI/TypeTools.hpp"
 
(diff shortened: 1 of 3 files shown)


henryiii:
The problem is that XCode 11 has `<filesystem>`, but you can't use it unless you target macOS 10.15. This is building on 10.14 and targeting 10.14.

https://stackoverflow.com/questions/58667853/does-use-of-c17-stdfilesystem-require-macos-10-15-xcode-11-1

henryiii:
@phlptp, can you add the windows warning suppression here? I've fixed the filesystem one for macOS.

include/CLI/Validators.hpp
```diff
@@ -14,17 +15,19 @@
...
 // [CLI11:verbatim]
-#if(defined(__cplusplus) && __cplusplus >= 201703L) || (defined(_HAS_CXX17) && _HAS_CXX17 == 1)
-#define CLI11_CPP17
-#endif
 
 // C standard library
 // Only needed for existence checking
 #if defined CLI11_CPP17 && defined __has_include && !defined CLI11_HAS_FILESYSTEM
 #if __has_include(<filesystem>)
+// Filesystem cannot be used if targeting macOS < 10.15
```
phlptp, line 23: I think I would try the feature macro `__cpp_lib_filesystem` and make sure it is > 201703
https://en.cppreference.com/w/User:D41D8CD98F/feature_testing_macros.  I thought there was a few cases where clang defined the header but not all features in filesystem as well.

henryiii: That's not the problem here (though I can check that too, that's the first thing I tried). The problem is that on macOS, you can target different macOS versions. XCode 11 knows how to target 10.15, but the default is `-mmacosx-version-min=10.14`, since it's compiling on 10.14. The implementation of `<filesystem>` is fine and complete, but if you target < 10.15, you can't use filesystem since it has compiled components that are not available when linking.

Incomplete `<optional>` was a common issue, but I think they've been more careful about adding `<filesystem>` once it's complete.

phlptp: I have run into similar problems with <string_view> in the past.

PR #399, opened 2020-01-11
Enhancements in building deb packages:

KOLANICH:
* Fixed building deb packages, broken because of lack of contact field.
* Added some missing metadata.
* Set the name of according to the convention.


henryiii:
New file needs to be added to Conan. I need to make sure Conan gets checked on GA or Azure.

jzakrzewski:
I'm not sure if the Debian-specific options should be set in the CMakeLists.txt directly. I'd say it's the packagers job to set them on the command line.

KOLANICH:
>I'm not sure if the Debian-specific options should be set in the CMakeLists.txt directly. I'd say it's the packagers job to set them on the command line.

IMHO:
1. there should be official packages built by upstream
2. if we provide any automation to build distro-specific packages, we should set all the metadata needed to build them. If a maintainer wants to redefine some metadata, he can do it via CLI.:

>-D <var>:<type>=<value>, -D <var>=<value>
Create or update a CMake CACHE entry.

These cache enties have more priority than the non-cache entries we set in CMake code.

jzakrzewski:
I am not convinced but I really don't have enough experience in that area, so I just may be making fool of myself ;)

henryiii:
I think it's okay to set defaults for Debian as long as they are reasonable; these are CMake specific settings and also can be overridden. I don't build Debian packages, though, so could easily be convinced either way, especially with an example.

henryiii:
> there should be official packages built by upstream

Should we build deb's and rpm's in CI? And what about when we have an optional compiled component? (not a blocker for this PR)

henryiii:
@all-contributors please add @KOLANICH for platform work

henryiii:
We can tweak this later if needed, but since the variables are easy to override, and have poor defaults (`libcli11-dev` is a better default than the package name), I think this is okay for now. We might want to make sure rpm's are easy to generate too. Might be useful in 2.0 when packaging will supply a binary component (and thus be more useful than simply providing some headers).

Thanks!

KOLANICH:
Thanks.

Issue #378, opened 2019-12-30
Code style guide for CLI11


cbachhuber:
I'm unable to find a code style guide in this project. However, `git grep -n guide` shows that in App.hpp:62, @henryiii seemed to plan on using the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) (GSG). 

I suggest using a style guide for CLI11, as it is a prerequisite for consistent code. **If you agree**, the question would be how this should be implemented. I see three things to consider:
- What to use
  - we could use GSG entirely
  - Only a subset of GSG, which must then be documented
  - Almost everything from GSG, and document only deviations from GSG.
- Where to document: in Contributing.md?
- How to implement the new style guide
  - Existing code does not need to change immediately. New and changed code should comply with GSG. When renaming a member function/variable of a class to comply with e.g. capitalization, all member functions/variables should be renamed to avoid inconsistencies.
  - We could introduce one style guide rule after the other, and with each new rule, directly make the entire code base compliant. I would be happy to take the lead on this.

What do you think?

henryiii:
A few thoughts:

* Following Google's style guide entirely is not a requirement, a subset or collection of deviations would be fine
* As much as possible should be checked in CI - I'm happy to see the cpplint fork (https://github.com/cpplint/cpplint) supports `#pragma once`
* Staying fairly close to the current style would be good (but less important than the point above)
* Contributing.md would be a good place unless the guide becomes too large
* I'd be happy to have you take the lead on this. :)

Keep in mind, in the near(ish) future, the will be one large change to the codebase (#338), which will comprise version 2.0, and if any external changes are needed, they ideally would go in then. There will also be custom rules related to the single file generation script. But it will be hard to work in parallel during that change, so I'll probably announce when I start work on that.

The single file script already has custom rules (all CLI11 includes mush have `""`, all other includes `<>`), but those will probably become explicit markers like `// [cli11:includes]` or such.

cbachhuber:
Thanks for all your input! I will read into cpplint and propose how I would like to approach the code style topic in the next days/weeks.

Also thanks for you hint on #338, I will pause working on code while you're working on that.

cbachhuber:
Continuing on this topic, I would like to add a license message to the header files in `include/CLI` and to the respective files in `examples` to satisfy cpplint's `legal/copyright` check. 

[...]

PR #400, opened 2020-01-12
Add cpplint to CI

cbachhuber:
Addressing #378: add cpplint to the CI pipeline with a set of checks that require small changes (<10 lines) in existing code.

Currently, cpplint is expected to fail (solutions to these issues are already implemented, though not yet pushed). At this moment, this PR checks whether cpplint works in the azure pipeline and whether the return value of cpplint is used. Once that is verified, I'll push the code resolving the issues.

You can run cpplint on your machine by checking out this PR and executing

```
pip install cpplint
cpplint --counting=detailed --recursive examples include/CLI
```

Please review the checks that are so far excluded in file `CPPLINT.cfg`. We can discuss those checks in this or in future PRs.

Advice on integrating cpplint into CI is welcome, I'm an azure pipeline beginner. Hints on running the CI pipeline on my local machine are welcome as well :)


cbachhuber:
Command [cpplint --counting=detailed --recursive examples include/CLI](https://github.com/CLIUtils/CLI11/pull/400/files#diff-fec826feae04e51c0d94076385408bdcR39) does not seem to be called inside the pulled docker container: azure says that [command `cpplint` is unknown](https://dev.azure.com/CLIUtils/CLI11/_build/results?buildId=574&view=logs&j=86000acd-da14-5f31-9e9b-9ba718f07c68&t=74bd8289-07b2-5415-ee2d-d23b7def2fda&l=14).

How can I change the pipeline such that this command is called inside docker, where cpplint is available?

azure-pipelines.yml
```diff
@@ -31,6 +31,14 @@ jobs:
     - script: git diff --exit-code --color
       displayName: Check tidy
 
+- job: CppLint
+  pool:
+    vmImage: 'ubuntu-latest'
+    container: sharaku/cpplint:latest
```
henryiii:
```suggestion
  container: sharaku/cpplint:latest
```

The container is not part of the pool. If you unindent this, it should work.

henryiii:
Hint: Notice the first line of the job step: 

```
/bin/bash --noprofile --norc /home/vsts/work/_temp/6092fe76-9723-4597-b2a3-114e50bc8d41.sh
```

Does not indicate it's running through docker. Scrolling up, there's no container starting lines in previous steps either. So that should be an indicator that the container is not being used. Now I'm not sure why Azure didn't complain about the pool mapping having an extra unused key...

henryiii:
Here's Azure's container docs; it's quite hard to find, IMO, since "docker" brings up things like how to build a docker image, how to host one, etc, etc, etc. Not how to run on one. https://docs.microsoft.com/en-us/azure/devops/pipelines/process/container-phases?view=azure-devops

Like GitHub Actions (they are really reskinned versions of mostly the same thing), you can specify a container per-step as well.

cbachhuber:
> Hint: Notice the first line of the job step [...]

That did the trick, thanks a lot! I will also have a look at the link such that I can do fewer [WIP]-PRs in future ;)

I would like to ask your opinion on the flags that are [to be discussed](https://github.com/CLIUtils/CLI11/pull/400/files#diff-050687d2ee08c540986e001b5812566bR14): 
1. legal/copyright: requires a copyright message in the beginning of every file. This project is already covered by the LICENSE file, so this is not necessary imho.
2. runtime/references: would like to replace non-const references with pointers. Message example: `include/CLI/Config.hpp:86:  Is this a non-const reference? If so, make const or use a pointer: std::string &name  [runtime/references] [2]`. I personally like (non-const) references and don't see a threat in them, so I would not use this check.
3. runtime/int: Message example: `include/CLI/TypeTools.hpp:598:  Use int16/int64/etc, rather than the C type long  [runtime/int] [4]`. This flag makes sense, and `int64_t` works with `std::stoll()`. Only 4 changes required.
4. build/include_order: Message example: `include/CLI/TypeTools.hpp:7:  Found C++ system header after other header. Should be: TypeTools.h, c system, c++ system, other.  [build/include_order] [4]`. I think it would be nice to have [ordered includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes). 24 issues found.

I will follow your guidance on these flags. I can include them in this PR (increasing the diff) or create separate PRs.

henryiii:
1. We probably should have a little header at the top of each with copyright info, pointing at LICENSE.
2. This would be a major change, completely changing the whole API. Needs to stay off.
3. This would add a new include, but isn't unreasonable.
4. We should fix the include order.

At least point 4 can go in, and maybe 3 and/or 1.

phlptp:
I think I would recommend waiting on 1 and 3 until after a 1.9 release.    Any new header would need to be handled well in the single file generator.  I think it reasonable and probably desirable but there may be reasons to wait on it until after the single file generator is updated.  

3 is also probably desirable but may involve changing some of the API,  which might be more appropriate in the 2.0 release.  

2 I am also not sure I agree with.  I sort of understand why Google does it that way but the same rational doesn't apply to most other libraries.  pointers come with nullability concerns so it would require a lot of modifications that I am not sure are useful in CLI11.  

for 4 it is probably fine as long as it doesn't interfere with the clang-format since that has some header checks as well I believe.  (at least is does on our other repos).

cbachhuber:
Sounds very good! I'll do the following:
1. Wait until after 1.9 release
2. Put to unused checks
3. Looking at the issues reported by cpplint, this should be doable without an API change. If yes, I'm going to include it in this PR.
4. I'll try to include this check in this PR if there is no conflict with clang-format.

phlptp (approved):
The changes seem fine to me.  

I should look up why cpplint does the headers in that order.  I think I have always put the local headers first so they can't rely on the system headers.  But maybe there is a good reason for the opposite.

cbachhuber:
> The changes seem fine to me.
> 
> I should look up why cpplint does the headers in that order. I think I have always put the local headers first so they can't rely on the system headers. But maybe there is a good reason for the opposite.

Thanks for your review! Regarding your comment: you're right, and [Google style guidelines recommend it as you propose](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) for the related header (i.e. `a.hpp` should be the first include of `a.cpp`). The headers that are shifted down in this PR are other project header files, not the related header file. The reasoning is that other project header files are themselves responsible for managing their includes in the same way, I think.

jzakrzewski:
> other project header files are themselves responsible for managing their includes in the same way

I'd say it encourages bad practise. With the "other project headers" after system headers it's easy to miss an include and make the header not self-contained (like you have to include vector before your project header or the thing will not compile).

examples/enum_ostream.cpp
```diff
@@ -1,4 +1,4 @@
-#include <CLI/CLI.hpp>
+#include "CLI/CLI.hpp"
```
henryiii: Why is change? This is not a relative import, so it should be `<>`, IMO. The main header files use `""`, but that's a special case, since the single file header script uses it to detect local includes.

henryiii:
Putting the system includes before the other project headers does look odd and possibly error prone to me. Does Google explain why they recommend this? LLVM seems to do more explaining on the "why" of the recommendations.

jzakrzewski:
I don't see official explanation from Google but there https://stackoverflow.com/a/54348720 is one possible explanation (maybe not why that order but at least why they could consider it safe). It doesn't really speak to me though.
There seems to be projects that explicitly deviate from the guidelines in that matter exactly to avoid unwanted dependencies: https://github.com/googleapis/google-cloud-cpp/blob/master/doc/cpp-style-guide.md#order-of-includes

examples/enum_ostream.cpp
cbachhuber: My main goal here was consistency in the examples. The majority used `""`, so I went with that. But you're right, let me change them all to  `<>`.

cbachhuber:
I'm also unable to find a proper explanation for this in Google's C++ style guide. Also, cpplint's check `build/include_order` cannot be modified. Therefore, I suggest removing this check from cpplint.

examples/enum_ostream.cpp
cbachhuber: Actually, the changeset is already large enough in this PR. Let me make the includes in the examples consistent in another PR.

include/CLI/Timer.hpp
```diff
@@ -10,7 +10,7 @@
 #endif
 
 #include <array>
-#include <chrono>
+#include <chrono> //NOLINT(build/c++11)
```
henryiii:
```suggestion
#include <chrono> // NOLINT(build/c++11)
```

While in this case, this file is not added to the single header build, it's best to remain consistent in spacing here since the single header script makes a set out of the includes and will see different spacings as different headers.

henryiii:
> This would add a new include, but isn't unreasonable.

Any file that uses the c std types needs `<cstdint>`. It is very likely any other std library header will include that one, but let's be pedantic and add it to the files that now use it. Otherwise, it looks good to go!

cbachhuber:
Ah, thanks for reminding me of this. I read that one should use `int64_t` and companions in the `std::` namespace. This is because headers that define parts of C, such as `cstdint`, are required to provide them in the `std::` namespace, [and may (not must!) in addition add them to the global namespace](https://stackoverflow.com/a/44668189/7260972). The latter might change, however, so you're on the safe side when using these symbols as defined in the `std::` namespace.

There are few more lines that need to be changed to follow this rule; I'll open a separate PR for fixing those after this PR has been merged. Ok?

henryiii (approved):

PR #407, opened 2020-01-18
Fix includes in examples

cbachhuber:
Fix includes in examples, [as discussed](https://github.com/CLIUtils/CLI11/pull/400#discussion_r367678579):
- Consistently use `#include <CLI/CLI.hpp>` in examples (instead of `#include "CLI/CLI.hpp"`)
- Explicitly add used headers that were so far only implicitly included


henryiii (commented):
:+1:

henryiii (approved):

henryiii:
Thanks! I’ll tag and release soon. Will be traveling for a week and might have a chance to get a bit done toward 2.0.

Issue #411, opened 2020-01-22
Unclear exact semantics of add_flag(std::string, bool&), possible undefined results


himikof:
I was trying to understand the best way to use a simple boolean flag in this library, and found during my experiments that the following test surprisingly fails (`value` is not set to `false`):

```c++
TEST_F(TApp, EmptyBoolFlag) {
    bool value = true; // Should be ignored?
    app.add_flag("-b,--bool", value);
    args = {};
    run();
    EXPECT_EQ(0u, app.count("--bool"));
    EXPECT_FALSE(value);
}
```

There are a lot of examples of `add_flag` usage like (book/code/flags.cpp):
```c++
bool flag_bool;
app.add_flag("--bool,-b", flag_bool, "This is a bool flag");
```
Do they currently leave the value undefined if the flag is not passed? Or the usage is wrong?
I think that using a similar `add_option` overload without any default is incorrect, but `add_flag` seems like a different case.

phlptp:
The add_options will not modify the value if no input is passed.  add_flag does the same thing.  Flags can default to true/false or other so assuming something about it doesn't seem like the correct thing to do and could lead to issues.   

So basically if no option/flag is passed in the command line, the value is not modified.    It is possible to specify.

`->defaultval(false);  to the option that would set the value to false.  

You know as I think about this a little more the `defaultval` behavior should be a little clearer, and probably executed on reset as well so if the app is called multiple times the value is modified on reset to the default value.    A `forced()` option might also be appropriate so an option is always guaranteed to update the value or call a callback, regardless of whether an option is passed or not.  

Regardless at the moment the above behavior is as expected and designed.    Whether that should be case or allowed to be modified might be worth some discussion.

himikof:
So if this flag behavior is expected, all `add_flag(std::string, bool&, ...)` calls without an explicit `= false` initialization (and without `default_val`) are broken and should be fixed? It seems that there are currently such examples in both docs and tests. I don't think this behavior is clear from the documentation now.

It would be unfortunate if the most simple way to use `add_flag` API is also wrong, users need to remember either to initialize to variable manually, or to always call `add_flag(...)->default_val(false);`

Also, the API invariant of "do not modify the value passed by reference if `option->count() == 0`" already does not hold for `add_flag(std::string, int&)`: the counter is eagerly set to zero there, and I assume that is intended. So the code
```
int flag_count;
app.add_flag("--count", flag_count);
```
works, but the same code with `bool` does not.

[...]

PR #416, opened 2020-01-26
Value initialization

phlptp:
Based on #411 This PR tries to clear up some things from the book and README.  

And go through all the tests and examples, and even though I am pretty sure there were no actual examples of undefined behavior, the potential was there in some cases.  And the use of non-initialized values in many of the tests probably isn't the best example as pointed out by @himikof.  

So this PR removes the uninitialized values I could find and tries to standardize the initialization to use brace initialization for most cases where the initialization is a constant.

### DIFF
--- a/book/chapters/flags.md
+++ b/book/chapters/flags.md
@@ -7,7 +7,7 @@ The most basic addition to a command line program is a flag. This is simply some
 The simplest way to add a flag is probably a boolean flag:
 
 ```cpp
-bool my_flag;
+bool my_flag=false;
 app.add_flag("-f", my_flag, "Optional description");
 ```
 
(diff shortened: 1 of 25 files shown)


phlptp:
This will probably take a few more commits to clean up things a little more and get the formatting correct again.

book/chapters/flags.md
```diff
@@ -19,12 +19,38 @@ This will bind the flag `-f` to the boolean `my_flag`. After the parsing step, `
...
 
 ```cpp
-int my_flag;
+int my_flag{0};
 app.add_flag("-f", my_flag, "Optional description");
 ```
 
 After the parsing step, `my_flag` will contain the number of times this flag was found on the command line, including 0 if not found.
 
+## arbitrary Type flags
```
henryiii:
```suggestion
## Arbitrary type flags
```

henryiii (approved):

PR #423, opened 2020-02-05
Container options

phlptp:
If merged this pull request will generalize the types that can be used in add_option to any container.  

A container in this case is any type with a `value_type` member and `iterator` member, an`end()` method, a `clear()` method, and an `insert` method.  This supports all the standard containers,  though not stack, or queue since they don't support an insert operation.  Should support most other containers as well if they have a similar interface to the standard library ones.  

Also support any type with a `value_type` defined.   And supports complex number type (any type with real(), and imag() methods directly in add_option.  

One other change is updated google test v1.10.0 to better support templated tests and match the current documentation for google test, so we don't have to figure out where old docs are for getting google test to work.


phlptp:
Todo:
- [x] test with some boost containers
- [x] make work with something like std::map<std::string, std::pair<int,int>>  
- [x] handle the general wrapper types in the type conversion
- [x] remove the vector specific operations
- [x] adjust the priorities for int_constructible and double_constructible classification types
- [x] update readme
- [x] update the book chapter on options to match current master
- [x] get compiling on all tested compilers
- [x] 100% coverage
- [x] merge commits and write a summary commit message

phlptp:
@henryiii  this is getting closer to being ready to go.   I am still working on the readme and book probably for a couple days yet.  

Had a few questions though.  With this PR complex numbers(without requiring the `<complex>` header be added) are supported directly in add_option.  So we can deprecate add_complex, but I wasn't sure if this was something we wanted to do now or in perhaps a later PR.  

To handle some options nicely like `vector<vector<X>>` some sort of user specified segment separator needs to be defined to be able to separate sections on the command line.  I was thinking `&&` but I don't have a strong preference.  
for examples

`--vec 1 2 4 7 && 3 5 --other A`  

would create a vector with two element vectors one with 4 element and 1 with 2.   But I am not sure if `&&` is the best or something else would be better.

jzakrzewski:
@phlptp Maybe rather something that does not require quoting in standard shells?

phlptp:
> @phlptp Maybe rather something that does not require quoting in standard shells?

That is a very good point that I had not been considering thus far.

phlptp:
So looking at a list of [bash special character](https://www.oreilly.com/library/view/learning-the-bash/1565923472/ch01s09.html)  
would indicate `&~#$*()\{}{};<>/?!` would probably not be used since they have different interpretations in shell scripting

That could mean something like `%%` might work,  a single `+` might work,  `++` is used as a subcommand terminator so wouldn't work in this context.  `#` or `##` might be possible since this would never begin a line and thus wouldn't be considered a comment line

phlptp:
I think I am leaning towards `%%` as a separator for subvector elements.  That doesn't seem to be used anywhere else or have meaning in command line contexts.

jzakrzewski:
Bash, and Windows CMD seems to be OK with that.

BTW: # does not have to be at the beginning of the line to start a comment in bash.

README.md
```diff
@@ -213,7 +214,7 @@ app.add_flag(option_name,
              help_string="")
 
 app.add_flag(option_name,
-             variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or any singular object with a defined conversion from a string like add_option
+             variable_to_bind_to, // bool, int, float,complex, containers, enum, or string-like, or any singular object with a defined conversion from a string like add_option
```
```suggestion
             variable_to_bind_to, // bool, int, float, complex, containers, enum, or string-like, or any singular object with a defined conversion from a string like add_option
```

README.md
```diff
@@ -195,14 +195,15 @@ While all options internally are the same type, there are several ways to add an
...
-               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕.
+               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕. Also supported are complex numbers🚧, wrapper types🚧, and containers besides vector🚧 of any other supported type
                help_string="")
 
 app.add_option_function<type>(option_name,
                function <void(const type &value)>, // type can be any type supported by add_option
                help_string="")
 
-app.add_complex(... // Special case: support for complex numbers
+app.add_complex(... // Special case: support for complex numbers ⚠️. complex numbers are now fully supported in the add_option so this function is redundant.
```
```suggestion
app.add_complex(... // Special case: support for complex numbers ⚠️. Complex numbers are now fully supported in the add_option so this function is redundant.
```

README.md
```diff
@@ -195,14 +195,15 @@ While all options internally are the same type, there are several ways to add an
 app.add_option(option_name, help_str="")
 
 app.add_option(option_name,
-               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕.
+               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕. Also supported are complex numbers🚧, wrapper types🚧, and containers besides vector🚧 of any other supported type
```
```suggestion
               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕. Also supported are complex numbers🚧, wrapper types🚧, and containers besides vector🚧 of any other supported type.
```

henryiii (approved):
I don't see any major issues; we can polish a little as we go if needed, but over all looks very good.

PR #424, opened 2020-02-06
Add copyright and copyright check

cbachhuber:
Continuing #378 and #400: this PR enables the cpplint check `legal/copyright` and prepends the copyright message agreed on in #378 to each `hpp` and `cpp` file in `include/CLI/` and in `examples/`.

Feel free to be pedantic about the copyright message; I have a script to automatically prepend a message to all above files.

PR #425, opened 2020-02-06
Fixing appveyor by using Python 3

henryiii:
The old python setting wasn't working, it was still using Python 2.


henryiii:
I believe this fixes the CI issue seen in #424.

henryiii:
#423 and #424 need to be rebased.

PR #429, opened 2020-02-12
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

### DIFF
--- a/.github/actions/cmake_config/entrypoint.sh
+++ b/.github/actions/cmake_config/entrypoint.sh
@@ -10,7 +10,7 @@ rm -rf cmake_dir/* build_tmp/*
 v=$1
 fn=cmake-$v-Linux-x86_64.tar.gz
 
-if [ ! -f cmake_souces/$fn ]; then
+if [ ! -f cmake_sources/$fn ]; then
     wget -qO cmake_sources/$fn "https://cmake.org/files/v${v%.*}/$fn"
 fi
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -551,7 +551,7 @@ Lots of cleanup and docs additions made it into this release. Parsing is simpler
 ## Version 0.3: Plumbum compatibility
 
 * Added `->requires`, `->excludes`, and `->envname` from [Plumbum](http://plumbum.readthedocs.io/en/latest/)
-* Supports `->mandatory` from Plubmum
+* Supports `->mandatory` from Plumbum
 * More tests for help strings, improvements in formatting
 * Support type and set syntax in positionals help strings
 * Added help groups, with `->group("name")` syntax
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -333,7 +333,7 @@ This is a quick patch release that makes LICENSE part of the single header file,
 
 ### Version 1.5.1: Access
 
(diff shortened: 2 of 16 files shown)


.github/actions/cmake_config/entrypoint.sh
```diff
@@ -10,7 +10,7 @@ rm -rf cmake_dir/* build_tmp/*
 v=$1
 fn=cmake-$v-Linux-x86_64.tar.gz
 
-if [ ! -f cmake_souces/$fn ]; then
+if [ ! -f cmake_sources/$fn ]; then
```
jsoref, line 13: This worries me

docs/Doxyfile
```diff
@@ -280,7 +280,7 @@ OPTIMIZE_OUTPUT_VHDL   = NO
 # parses. With this tag you can assign which parser to use for a given
 # extension. Doxygen has a built-in mapping, but you can override or extend it
 # using this tag. The format is ext=language, where ext is a file extension, and
-# language is one of the parsers supported by doxygen: IDL, Java, Javascript,
+# language is one of the parsers supported by doxygen: IDL, Java, JavaScript,
```
jsoref, line 283: Some projects have asked me to drop changes to `Doxyfile` as it tends to be just copied from elsewhere. I'm happy to do that if requested (or similarly for other files) -- just let me know.

jsoref (commented):
Fwiw, I'm not a fan of the use of `subcomms` instead of `subcommands` or `subcmds`, but I was afraid to make that change, so I left them alone.

CHANGELOG.md
```diff
@@ -551,7 +551,7 @@ Lots of cleanup and docs additions made it into this release. Parsing is simpler
 ## Version 0.3: Plumbum compatibility
 
 * Added `->requires`, `->excludes`, and `->envname` from [Plumbum](http://plumbum.readthedocs.io/en/latest/)
-* Supports `->mandatory` from Plubmum
+* Supports `->mandatory` from Plumbum
```
henryiii, line 554: Impressed it caught this one

.github/actions/cmake_config/entrypoint.sh
henryiii: It was an attempt to cache, IIRC, and doesn't really affect anything (other than failing to cache).

docs/Doxyfile
henryiii: I think it's okay; eventually I'll switch to using the CMake 3.1x Doxygen tools, but until then, I don't think the Doxyfile will get regenerated.

henryiii (approved):

henryiii:
@allcontributers, add @jsoref for tools

henryiii:
@all-contributors, please add @jsoref for tools

CHANGELOG.md
jsoref: The tooling just reports tokens not in the dictionary, since neither are in the dictionary, I'm presented w/ two fairly close tokens in a long list. I just check the context for both and pick one as the more likely real token. As a human, I do reasonably well at picking, but I do fail periodically. Also, since I tend to do this late at night, I may skip over some set of tokens deciding "they're probably real".

PR #430, opened 2020-02-12
conan: Do not run tests when cross compiling

geir-t:
Running conan create with a profile made for cross compiling would fail since the tests is not compiled for build OS

Open to other ways of solving this issue

### DIFF
--- a/test_package/conanfile.py
+++ b/test_package/conanfile.py
@@ -1,4 +1,4 @@
-from conans import ConanFile, CMake
+from conans import ConanFile, CMake, tools
 import os
 
 
@@ -16,5 +16,6 @@ class HelloTestConan(ConanFile):
         self.copy("*.dylib*", dst="bin", src="lib")
 
     def test(self):
-        os.chdir("bin")
-        self.run(".%sexample" % os.sep)
+        if not tools.cross_building(self.settings):
+            os.chdir("bin")
+            self.run(".%sexample" % os.sep)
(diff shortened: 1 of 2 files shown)


henryiii:
@all-contributors, please add @geir-t for platform

henryiii:
I'll try to test this locally soon, then merge. Thanks! The build should not run tests, you are correct.

henryiii:
(And, technically, you don't need to build, since the "build" process itself only builds tests - this will probably change in 2.0 though)

geir-t:
Thank you for the quick response. Ok, then I'll leave the build step alone for now.

test_package/conanfile.py
```diff
@@ -16,5 +16,6 @@ def imports(self):
         self.copy("*.dylib*", dst="bin", src="lib")
 
     def test(self):
-        os.chdir("bin")
-        self.run(".%sexample" % os.sep)
+        if not tools.cross_building(self.settings):
```
henryiii, line 19: Can this be `get_env("CONAN_RUN_TESTS", True)` too?

geir-t: It looks like that variable needs to be defined in the profile, which for us it's currently not. The cross_building func deduces it given the target profile and current host machine so I think that is the right choice. Hopefully that will work out of the box for more people. 

I took the liberty of changing to the cross_building check in the main conanfile for the same reasons listed above.

henryiii: Okay, though I'd recommend bring it up on the Conan docs, to see if the suggestion can be changed.

henryiii:
Thanks!

PR #434, opened 2020-02-17
Add and fix cpplint whitespace/comments

cbachhuber:
Finalizing the work started in #400 (and part of #378): this PR enables the cpplint check `whitespace/comments` and fixes corresponding issues.

Also, it aligns the clang-format style to 2 spaces between code and trailing comments.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -75,7 +75,7 @@ SortIncludes:    true
 # SpaceBeforeAssignmentOperators: true
 SpaceBeforeParens: Never
 # SpaceInEmptyParentheses: false
-# SpacesBeforeTrailingComments: 1
+SpacesBeforeTrailingComments: 2
 # SpacesInAngles:  false
 # SpacesInContainerLiterals: true
 # SpacesInCStyleCastParentheses: false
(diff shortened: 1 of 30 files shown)


cbachhuber:
clang-format now requires test code also in proper formatting. Should we add folder `tests` to cpplint checks? I just checked, with the current cpplint configuration, there are 23 legal/copyright issues in `tests`, and 11 other issues.

phlptp:
In general I think it is a good idea to treat the test code the same as the library code, so I would be in favor of  running any checks on the test code.    With the knowledge that I am pretty sure there are a few cases where we are purposely doing something Abnormal for the purpose of testing some aspect of the library, so there will likely be a few places where we have bypass the checks.

phlptp:
I would enable the test checks in a separate PR though.

henryiii:
I second that. Tests and (usually) examples should be included in checks. Second PR is best.

henryiii (approved):

cbachhuber:
Got it, I'll create a separate PR for including `tests` in cpplint and fixing the corresponding issues :+1:

PR #442, opened 2020-03-07
Check tests with cpplint

cbachhuber:
As agreed in #434, tests are now being checked by cpplint as well.

@henryiii: I read that you're currently quite busy, so feel free to take your time reviewing this :)


tests/TimerTest.cpp
```diff
@@ -1,10 +1,16 @@
...
+// under NSF AWARD 1414736 and by the respective contributors.
+// All rights reserved.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
 #include "CLI/Timer.hpp"
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
-#include <chrono>
+#include <chrono>  // NOLINT(build/c++11)
```
phlptp: Why are there potential warnings on these headers?  regex I understand somewhat since that was a little late coming to some c++11 compilers, but chrono?

cbachhuber: Good point. Cpplint's warning for this line is `<chrono> is an unapproved C++11 header.`

Googling that yields [this discussion](https://stackoverflow.com/questions/33653326/google-style-guide-chrono-is-an-unapproved-c11-header) as best source. The line of argument there is that in the [chromium project](https://chromium-cpp.appspot.com/), chrono overlaps with the time API of chromium. [Cpplint's source code](https://github.com/cpplint/cpplint/blob/master/cpplint.py#L6128) does not give any more details on this.

It seems that the line of argument against chrono is very weak for general C++11 projects. We could think about removing cpplint check `build/c++11`. The other warnings of this check are because of `regex` (reasonable) and `thread` (again, overlap in chromium).

phlptp: I think I would lean towards removing the check.  We will catch any use of regex from some of the compilers that don't support it, and the others are all fine in all known compilers CLI11 support, so having an extra odd comment seems mildly annoying and not helpful.

cbachhuber: I agree. I pushed the changes to this PR, considering that the PR is conceptually not big.

phlptp (approved):

henryiii:
Can you resolve conflicts? Then we can get this in. :)

henryiii:
Thanks!

Issue #445, opened 2020-03-14
Undefined reference to std::filesystem::status in C++17


timothyqiu:
I'm using CLI11 1.9.0 with gcc 8.3.0 on Debian 10, and I tried the single header version, a manual install, and the Conan package.

If the main project is in C++17, then by default, using CLI11 produces an undefined reference error when linking:

```
/usr/bin/ld: CMakeFiles/demo.dir/main.cpp.o: in function `CLI::detail::check_path(char const*)':
main.cpp:(.text._ZN3CLI6detail10check_pathEPKc[_ZN3CLI6detail10check_pathEPKc]+0x38): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&, std::error_code&)'
/usr/bin/ld: CMakeFiles/demo.dir/main.cpp.o: in function `std::filesystem::__cxx11::path::path<char const*, std::filesystem::__cxx11::path>(char const* const&, std::filesystem::__cxx11::path::format)':
main.cpp:(.text._ZNSt10filesystem7__cxx114pathC2IPKcS1_EERKT_NS1_6formatE[_ZNSt10filesystem7__cxx114pathC2IPKcS1_EERKT_NS1_6formatE]+0x64): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This is because gcc < 9 requires a separate library `libstdc++fs` for `std::filesystem`.

I think this is a CLI11 issue because when using CMake, the target `CLI11::CLI11` is supposed to specify all its required libraries, so a `target_link_libraries(demo PRIVATE CLI11::CLI11)` is enough for the user.

The current workaround is to manually adding the missing libraries or defining `CLI11_HAS_FILESYSTEM=0` to disable the `std::filesystem` approach.

A possible solution is to detect the correct libraries, as discussed [here](https://gitlab.kitware.com/cmake/cmake/issues/17834). For example using something similar to this [FindFilesystem.cmake](https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake).

phlptp:
Despite some trouble with azure, we are able to replicate the issue so it will be fixed soon.  Thanks for identifying the issue.    See #446

PR #446, opened 2020-03-14
add windows latest and gcc 8 builds to azure

phlptp:
Add a MSVC 2019 build and a gcc 8 build for use with C++17.

Test issue #445

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -143,11 +143,6 @@ target_compile_options(CLI11_warnings
 add_library(CLI11 INTERFACE)
 add_library(CLI11::CLI11 ALIAS CLI11) # for add_subdirectory calls
 
-#if (CMAKE_CXX_STANDARD GREATER 16)
-#    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 8)
-#       target_link_libraries(CLI11 INTERFACE -lstdc++fs)
-#    endif()
-#endif()
 # Duplicated because CMake adds the current source dir if you don't.
 target_include_directories(CLI11 INTERFACE
      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,7 +145,7 @@ add_library(CLI11::CLI11 ALIAS CLI11) # for add_subdirectory calls
 
 if (CMAKE_CXX_STANDARD GREATER 16)
     if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 8)
-       target_link_libraries(CLI11 PUBLIC -lstdc++fs)
+       target_link_libraries(CLI11 INTERFACE -lstdc++fs)
     endif()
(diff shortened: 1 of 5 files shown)


phlptp:
Why is azure not running?

henryiii:
Good question. Last run is on Wednesday.

henryiii:
I think it might just be a hiccup? You could try repushing. I'll add you to the Azure admins if you PM me your azure ID (username or whatever it uses).

phlptp:
So as far as I can tell something must have changed in azure pipelines that they don't automatically run on PR's unless you specify that in the config file.  

I added three tests.  a windows latest for visual studio 2019, which It didn't look like we were testing otherwise and probably should.  
A gcc 8 C++17 which is the trigger for this that was failing and is now passing.  

and a clang8 C++17 build which I thought might fail for the same reason as GCC 8 but didn't,  So not sure what is up with that yet.  

We can leave in the new clang8 build or not.   The fix was to specifically target GCC 8 and C++17 and add the specific header.  

I probably should add a build note to the read me about the potential issue yet.

phlptp:
fixes #445

CMakeLists.txt
```diff
@@ -143,6 +143,11 @@ target_compile_options(CLI11_warnings
 add_library(CLI11 INTERFACE)
 add_library(CLI11::CLI11 ALIAS CLI11) # for add_subdirectory calls
 
+if (CMAKE_CXX_STANDARD GREATER 16)
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 8)
```
henryiii: Why is this "not" version equal 8? This only needs to be added for C++17 & GCC < 9.

henryiii:
Wouldn't it be better to modify the automatic selection of the filesystem header to not turn on for GCC < 9? Then a user could force it on, but if not, then it doesn't try to use it. That also fixes the single file header.

CMakeLists.txt
phlptp: I think it only really applies to GCC 8.  GCC 7 doesn't have <filesystem> and GCC 9 doesn't have a problem, so this was a the condition that needed the extra linking target.

phlptp:
We would have to figure out the condition for the libstdc++ that matched Gcc 8, but it could be done that way as well.   I suppose from a single header perspective that would be better since it wouldn't need the extra linking library.

CMakeLists.txt
henryiii: But shouldn't it be `if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 8)`?

phlptp: Yes, so why is then working on gcc 8?   I am trying with the in code check, if that works we can probably remove it from CMAKE.

henryiii: The version equal bothers me; `VERSION_LESS 9` would be better. I assume this isn't expecting 8.0.0?

henryiii:
Looks good; does the special instructions need to be updated?

Also, maybe of interest to @cbachhuber, I think we should add `IndentPPDirectives: AfterHash` to make this easier to read. (not in this PR, though).

phlptp:
I did update the readme note regarding gcc 8,  if you think it needs to be clearer we can edit it.

PR #449, opened 2020-03-28
Add char type

phlptp:
Add support for a char type in the options.  

This was technically supported before but in that the option type was accepted but it was always interpreted as an int, which is not necessarily what is desired.  The added conversion interprets single character strings as the character, and otherwise defaults to a numerical conversion.  

This issue came up a couple times in our usage, when character inputs unexpectedly produced errors when trying to define an option for a separation character or delimiter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ While all options internally are the same type, there are several ways to add an
 app.add_option(option_name, help_str="")
 
 app.add_option(option_name,
-               variable_to_bind_to, // bool, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕. Also supported are complex numbers🚧, wrapper types🚧, and containers besides vector🚧 of any other supported type.
+               variable_to_bind_to, // bool, char(see note)🚧, int, float, vector, enum, or string-like, or anything with a defined conversion from a string or that takes an int 🆕, double 🆕, or string in a constructor. Also allowed are tuples 🆕, std::array 🆕 or std::pair 🆕. Also supported are complex numbers🚧, wrapper types🚧, and containers besides vector🚧 of any other supported type.
                help_string="")
 
 app.add_option_function<type>(option_name,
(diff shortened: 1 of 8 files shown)


phlptp:
This has become a little more complicated than originally intended.   The issue arose with handling enumerations with a char type, which wouldn't want to allow single characters.  So that required discriminated those cases in the lexical cast.  Which led to a separate function for integral conversions and some other code cleanup that could be done as a result.  which led to some tweaks to some of the validators.

phlptp:
One remaining question:  In the char conversions with single digits '0' - '9'  should these be treated as the numerical equivalent or the character?   treating them as numbers would be completely backwards compatible,  but questionable whether that is needed or not when moving to 2.0.   Not sure I have a strong preference,  my use case wouldn't be impacted  and enum:char are handled separately.

henryiii:
So would 03 be the way to get the char value 3? Then 3 probably could the be char 3, as otherwise you can't get the char 3 without looking up it's numerical code?

phlptp:
If we treated single numerical values as character ['0'-'9']  the ways to get the numerical value if that is what you wanted would be, using 3 as an example:
"03", "+3", "0x03"  I think the octal and binary equivalent might work as well

henryiii:
Do you have a stronger opinion now about the best choice here, or are you still torn?

phlptp:
My inclination is that taking it as an actual char is preferable, since otherwise it would be complicated to actually get the character '3',  

my only hesitancy is that it does break backwards compatibility.   which should be fine if the next release is 2.0.

henryiii:
We may make a 1.9.1, but I think we've already sailed past 1.10. The downside is I'll have to get two or three large changes in that I really want in 2.0.

henryiii:
>  I think we've already sailed past 1.10

Feel free to correct me if I'm wrong, I need to compile the changelog.

phlptp:
yeah I think there were a few changes that already broke backwards compatibility in a few circumstances, so we were definitely on the path to a 2.0 release.

phlptp:
@henryiii Was there anything you wanted me to do with this or are you just waiting for a 2.0 release?

henryiii:
I was a little indecisive about "1" being char 1 or value 1. Rereading it, I think the point that there's a way to get a number, but not a good way to get a char is reasonable enough to argue for this being best.

henryiii (approved):

PR #452, opened 2020-04-09
Version add

### LINKED ISSUE #436: add a version callback similar to help
I am wondering how much use a dedicated version option, like help/help_all/config would be.  It is similar to help in that it should bypass the other error checks and required options.  

I know we end up putting it in most use cases in some way or another,  having it be embedded in CLI11 would standardize and make it easier for everyone.  

I would picture something like
an overload set
```cpp
app.set_version(string version_string);
app.set_version(std::function<std:;string()> version_callback);  
app.set_version(string version_args, string version_string);
app.set_version(string version_args, std::function<std:;string()> version_callback);  
```

this would add a version option ("-v,--version" that triggers a version_exception and is triggered like help does currently.

phlptp:
Add a version operation, similar to config and help, to print out the version and exit.  Supports adding the version via a string or callback.  

We had been doing this in our application and it might seem to be something that is pretty common to want in applications

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1577,6 +1577,23 @@ class App {
         return formatter_->make_help(this, prev, mode);
     }
 
+    /// Displays a version string
+    std::string version() const {
+        std::string val;
+        if(version_ptr_ != nullptr) {
+            auto rv = version_ptr_->results();
+            version_ptr_->clear();
+            version_ptr_->add_result("true");
+            try {
+                version_ptr_->run_callback();
+            } catch(const CLI::CallForVersion &cfv) {
+                val = cfv.what();
+            }
+            version_ptr_->clear();
+            version_ptr_->add_result(rv);
+        }
+        return val;
+    }
     ///@}
     /// @name Getters
     ///@{
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -742,6 +742,17 @@ class App {
 
         return version_ptr_;
     }
+    /// Set the version as a string
+    App *version(const std::string &version) {
+        set_version_flag("-v,--version", version);
+        return this;
+    }
+
+    /// Set the version as a function
+    App *version(std::function<std::string()> vfunc) {
+        set_version_flag("-v,--version", vfunc);
+        return this;
+    }
 
   private:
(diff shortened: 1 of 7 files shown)


include/CLI/App.hpp
```diff
@@ -703,6 +706,56 @@ class App {
...
+                [vfunc]() { throw(CLI::CallForVersion(vfunc(), 0)); },
+                "display program version information and exit");
+            version_ptr_->configurable(false);
+        }
+
+        return version_ptr_;
+    }
+    /// Set the version as a string
+    App *version(const std::string &versionString) {
+        set_version_flag("-v,--version", versionString);
```
nightlark: `-v` is commonly also used for verbose by programs (including cp, rm, wget, curl, mv).

jzakrzewski: So is `-V` basically it's a guess which one will be equivalent to `--version` and which to `--verbose`. Each default is equally good/bad.

phlptp: ok, I removed the "-v" from the default since it seems that could mean any number of different things from common applications.

phlptp:
closes #436

henryiii:
I am not really sure we need to have the `version` shortcut. It only adds the "--version" for you - which I was going to consider suggesting as an argument, but then that's exactly what `set_version_flag` is. There's also some ambiguity in whether it is setting the version or accessing the version (which is not needed, but it's not completely clear). Thoughts?

phlptp:
I am fine removing it, hadn't thought about the ambiguity on whether it would be accessing the version.

phlptp:
Do we want to keep the `std::string version() const` operation?

henryiii:
No (I'd missed that, actually), but that's what I would expect "version()" to do; the dual use might be also a bit confusing.

phlptp:
removed all `version`  methods

henryiii:
I'm sorry, I meant "Yes, we could keep it" - somehow I stared with No then (I think) explained why we should keep it. I was just saying that version() returning a string and version(string) could possibly be confusing, so it was another reason that it might be nice to remove version(string).

phlptp:
ok, I will that one back in :)

henryiii (approved):

PR #455, opened 2020-04-22
fix for the cmake_config tests

phlptp:
fix a repo change for git in ubuntu18.04 that impacted the cmake_config tests


henryiii:
Thanks! In-hospital test of our flow meter tomorrow, after that might have a little time. First batch ships May 1st, so after that I'll have more time (at least I hope so, getting new software onto nurse stations will not be easy, and new software onto the devices will be a real pain. So it needs to work before shipping!)

phlptp:
I use this same test in a few other repositories, and they broke a few days ago with a priority fix to the git in the docker image, figured the fix would be needed here too.

phlptp:
Good luck!  getting hardware to work with software is always a different level of challenge.

PR #459, opened 2020-05-07
try clang 10

phlptp:
Add a test build to explore #458.

PR #467, opened 2020-05-31
Use e.get_name instead of dynamic_cast

### LINKED ISSUE #466: CLI11 does not compile without rtti
My project is compiled without rtti. I get the following error when trying to use CLI11 version 1.9.0:
```
.../src/bin/CLI11.hpp: In member function ‘int CLI::App::exit(const CLI::Error&, std::ostream&, std::ostream&) const’:
.../src/bin/CLI11.hpp:5863:54: error: ‘dynamic_cast’ not permitted with -fno-rtti
         if(dynamic_cast<const CLI::RuntimeError *>(&e) != nullptr)
                                                      ^
```
Is RTTI required to use CLI11?

certik:
Also use std::static_pointer_cast instead of std::dynamic_pointer_cast

Fixes #466

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -64,6 +64,10 @@ jobs:
         vmImage: 'windows-2019'
         cli11.std: 20
         cli11.options: -DCMAKE_CXX_FLAG="/std:c++latest"
+      Linux17nortti:
+        vmImage: 'ubuntu-latest'
+        cli11.std: 17
+        cli11.options: -DCMAKE_CXX_FLAG="-fno-rtti"
   pool:
     vmImage: $(vmImage)
   steps:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -63,11 +63,11 @@ jobs:
       Windowslatest:
         vmImage: 'windows-2019'
         cli11.std: 20
-        cli11.options: -DCMAKE_CXX_FLAG="/std:c++latest"
+        cli11.options: -DCMAKE_CXX_FLAGS="/std:c++latest /EHsc"
       Linux17nortti:
         vmImage: 'ubuntu-latest'
         cli11.std: 17
-        cli11.options: -DCMAKE_CXX_FLAG="-fno-rtti"
+        cli11.options: -DCMAKE_CXX_FLAGS="-fno-rtti"
   pool:
     vmImage: $(vmImage)
   steps:
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1606,7 +1606,12 @@ class App {
 
     /// Access the config formatter as a configBase pointer
     std::shared_ptr<ConfigBase> get_config_formatter_base() const {
+        // This is safer as a dynamic_cast if we have RTTI, as Config -> ConfigBase
+#if defined(__cpp_rtti) || (defined(__GXX_RTTI) && __GXX_RTTI) || (defined(_HAS_STATIC_RTTI) && (_HAS_STATIC_RTTI == 0))
+        return std::dynamic_pointer_cast<ConfigBase>(config_formatter_);
+#else
         return std::static_pointer_cast<ConfigBase>(config_formatter_);
+#endif
     }
 
     /// Get the app or subcommand description
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1374,20 +1374,20 @@ class App {
     int exit(const Error &e, std::ostream &out = std::cout, std::ostream &err = std::cerr) const {
 
         /// Avoid printing anything if this is a CLI::RuntimeError
-        if(dynamic_cast<const CLI::RuntimeError *>(&e) != nullptr)
+        if(e.get_name() == "RuntimeError")
             return e.get_exit_code();
 
-        if(dynamic_cast<const CLI::CallForHelp *>(&e) != nullptr) {
+        if(e.get_name() == "CallForHelp") {
             out << help();
             return e.get_exit_code();
         }
 
-        if(dynamic_cast<const CLI::CallForAllHelp *>(&e) != nullptr) {
+        if(e.get_name() == "CallForAllHelp") {
             out << help("", AppFormatMode::All);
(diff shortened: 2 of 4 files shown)


henryiii:
What about the 4 dynamic_casts in Option? I think those can be converted, too? We might need to be a little careful with needs/excludes, but it's not actually checking a type, but just making sure it exists.

certik:
I did the changes in this PR in the latest release of `CLI11.hpp`. This was all that was needed for `CLI11.hpp` to compile. Is the Option a new addition to CLI11?

include/CLI/App.hpp
```diff
@@ -1606,7 +1606,7 @@ class App {
 
     /// Access the config formatter as a configBase pointer
     std::shared_ptr<ConfigBase> get_config_formatter_base() const {
-        return std::dynamic_pointer_cast<ConfigBase>(config_formatter_);
+        return std::static_pointer_cast<ConfigBase>(config_formatter_);
```
certik, line 1613: @henryiii I am not sure about this change --- is it safe?

henryiii:
Have you added any options? It's templated internally, so it might not compile unless you add at least one option. But otherwise, it should be there. `git grep dynamic_cast` to show them - again, I think two of them are completely trivial to change, and the other two might need a nullptr check or something like that.

include/CLI/App.hpp
henryiii: This is a cast up, so yes it is. Only casts down need the dynamic casts.

henryiii:
> Codacy was unable to analyse your pull request

This sounds like Codacy's error, not ours. Not sure why that counts as an error. Ignore it.

I think the Windows GoogleTest error is from #459 and unrelated - not sure why it's breaking now and not then.

include/CLI/App.hpp
phlptp: I don't think this entirely safe.  ConfigBase inherits from Config.  So if someone put in a custom Config object this would be expected to return null.    On the other hand if you were doing that you should never use this function.  So it isn't unreasonable to have this function commented out in the case of turning off RTTI.

henryiii: You are correct; I chose very confusing names there and managed to confuse myself with them. Then this needs an `#if` for RTTI.

phlptp:
If we are going to support a no  RTTI build we probably should have a test that builds with those flags to make sure everything is running correctly.

certik:
Just a heads up that I implemented the above changes in our project in this MR:

https://gitlab.com/lfortran/lfortran/-/merge_requests/364

in particular, here is the commit that does the changes:

https://gitlab.com/lfortran/lfortran/-/commit/4321ad350f9e6be0a9bdf95977a67f029ecb2b9a

and here is the main program that uses CLI11:

https://gitlab.com/lfortran/lfortran/-/blob/921886c90b8f7fea2ebb7539349249f83ccb4873/src/bin/lfortran.cpp#L278

and things compile without RTTI (our LLVM dependency requires to compile without RTTI) into a statically linked binary and everything seems to work on all platforms. The only other thing we had to do was to explicitly link against the `stdc++fs` library on some platforms, as one of our CI builds failed (in Alpine):

https://gitlab.com/lfortran/lfortran/-/commit/e1f0914c87bc704c0e7feeed6fbc2743caf1db71#00763928ad5c162833c2e7ea2f675238a7b9ed22

It might be worth mentioning this fact somewhere in the CLI11 README, as it took me a while to figure this out.


Thanks for this excellent library! It seems to do exactly what I've been looking for.

certik:
@henryiii, the changes in `Option.hpp` do not seem to be needed --- they compile just fine without RTTI on my machine. I assume that's because the compiler is clever enough to know it's an upcast?

henryiii:
I believe so, at least GCC claims it that dynamic_cast is usable for an upcast with RTTI off. But there's no need to use dynamic_cast for an upcast, so it's clearer to use static_cast.

azure-pipelines.yml
```diff
@@ -63,7 +63,11 @@ jobs:
       Windowslatest:
         vmImage: 'windows-2019'
         cli11.std: 20
-        cli11.options: -DCMAKE_CXX_FLAG="/std:c++latest"
+        cli11.options: -DCMAKE_CXX_FLAGS="/std:c++latest /EHsc"
```
henryiii, line 66: @phlptp This was not working before, due to a typo (missing S). Now we need EHsc, maybe because this replaces the default flags? Just checking to make sure it is fine.

henryiii:
I *think* I have a working RTTI off test build now. If this builds, I may disable the RTTI fix temporarily just to ensure the build breaks.

henryiii:
> there's no need to use dynamic_cast for an upcast

The fact that we are not checking to see if it's a null pointer should be a giveaway that we aren't actually doing a dynamic cast. ;)

henryiii:
It works! The no RTTI build breaks with the change removed. :)

include/CLI/App.hpp
henryiii: Added a note to the doc. It you have RTTI off and have a non-standard formatter, this will return garbage.

phlptp: I think if it is a static cast it will return garbage with a custom formatter that is not based on ConfigBase regardless of RTTI.

phlptp: Might be worth renaming in CLI11 2.0 to make it a little clearer about the inheritance tree here.

certik:
@henryiii thanks for working on this, and incorporating these changes.

azure-pipelines.yml
phlptp: That explains a few things.  I suppose in general requiring /EHsc could be potentially problematic but probably not for anything CLI11 does by itself.

include/CLI/App.hpp
henryiii: That's why I have an #if to switch it to `dynamic_cast` if RTTI is on. In 2.0, we can rename, and also just return the unmodified base pointer, and let the user do the cast.

PR #469, opened 2020-06-01
Simplify clang tidy config file

cbachhuber:
The current clang-tidy setup treats all enabled clang-tidy warnings as errors. Simplifying the clang-tidy config file to keep this behavior while improving readability and ease of maintenance.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -15,17 +15,7 @@ modernize*,
 readability-container-size-empty,
 '
 
-WarningsAsErrors: '
--*,
-google-*,
--google-runtime-references,
-llvm-include-order,
-llvm-namespace-comment,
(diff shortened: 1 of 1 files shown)


henryiii:
`result_of` error on windows-latest shows up in other PRs.

phlptp:
looks like a google test issue rather than something in the CLI11 library.  The Windows latest must have updated the version of visual studio to used.  Previously the only failure occurred on the preview release.  We might have to update the google test submodule to master to fix it since the fix in google test wasn't made until late April.

henryiii:
Yes, you are right - I forget it was in google-test. We can pick a commit to checkout, or we can just bump to the newer google-test on this one job. Google usually has a live-at-head policy so master should be fairly stable.

PR #475, opened 2020-06-10
Fix internal header include paths.

samhocevar:
The extra leading "CLI/" part of include directives prevents the inclusion
of CLI.hpp from a relative directory without an extra -I or /I compiler
directive.


henryiii:
I think this looks fine to me. Probably can be back ported to 1.9.1.

henryiii:
Backported.

PR #484, opened 2020-06-24
Allow_extra args and single vector argument

phlptp:
allow the use of `allow_extra_args(false)` to specify only a single argument at a time with a vector output.

based on [gitter conversation](https://gitter.im/CLI11gitter/Lobby?at=5ef1f99047fdfd21edde2901)

Including operation tweak and test case


include/CLI/App.hpp
```diff
@@ -2761,7 +2761,10 @@ class App {
         }
         int min_num = (std::min)(op->get_type_size_min(), op->get_items_expected_min());
         int max_num = op->get_items_expected_max();
-
```
henryiii, line 2764: Could you add a little comment here, just like the block below, mentioning what this is here for?

README.md
```diff
@@ -322,6 +322,7 @@ Before parsing, you can set the following options:
 -   `->ignore_case()`: Ignore the case on the command line (also works on subcommands, does not affect arguments).
 -   `->ignore_underscore()`: Ignore any underscores in the options names (also works on subcommands, does not affect arguments). For example "option_one" will match with "optionone".  This does not apply to short form options since they only have one character
 -   `->disable_flag_override()`: From the command line long form flag options can be assigned a value on the command line using the `=` notation `--flag=value`. If this behavior is not desired, the `disable_flag_override()` disables it and will generate an exception if it is done on the command line.  The `=` does not work with short form flag options.
+-   `->allow_extra_args(true/false)`: If set to true the option will take an unlimited number of arguments like a vector, if false it will limit the number of arguments to the size of the type used in the option.  Default value depends on the nature of the type use, containers default to true, others default to false.
```
henryiii: We should add an emoji here for the in-progress feature. Though, I'm tempted to back port this to v1, probably would have to be a small 1.10 release.

tests/OptionTypeTest.cpp
```diff
@@ -838,3 +838,16 @@ TEST_F(TApp, tupleTwoVectors) {
     EXPECT_EQ(std::get<0>(cv).size(), 2U);
     EXPECT_EQ(std::get<1>(cv).size(), 3U);
 }
+
+TEST_F(TApp, vectorSingleArg) {
```
henryiii, line 842: Shouldn't we have a test with a size 2+ vector as well?

include/CLI/App.hpp
phlptp: done

README.md
phlptp: done

tests/OptionTypeTest.cpp
phlptp: good call! on the additional test,  had to change the condition to make it work for that

include/CLI/App.hpp
```diff
@@ -2761,7 +2761,12 @@ class App {
...
+        // check container like options to limit the argument size to a single type if the allow_extra_flags argument is
+        // set.
+        if(max_num >= detail::expected_max_vector_size / 16 && !op->get_allow_extra_args()) {
```
henryiii, line 2766: Why 16?

phlptp: No very good reason.  I realized if I was going to make this < it had to be at least /4 since I use the max_size/2 in another spot and figured it should be a multiple of 2.  So 4/8/16/32/64 probable doesn't make any difference right now.

henryiii:
By the way, I'm seeing if the problem with Clang 3.4's docker image can be fixed upstream. https://github.com/silkeh/docker-clang/issues/7

henryiii:
Should be fixed in the clang docker images now. Can you retrigger the build?

Issue #486, opened 2020-07-06
Providing several config files


jbriales:
I read through the docs but didn't found anything about this:
Would it be possible in CLI11 to take several input config files, and just concatenate their entries?
Currently if I provide `--config` more than once I just get:
> --config: At Most 1 required but received 2

The motivation would be to be able to reuse multiple subsets of options and combine them.

phlptp:
I don't think it is currently handled but I don't think the modification would be that hard.

PR #494, opened 2020-07-29
Add an ability to handle multiple config files

phlptp:
Address Issue #486 

Allow support for multiple config files

```
set_config("--config")->expected(1, 3);
```
will allow upto expected max config files.


phlptp:
- [x] update docs
- [x] verify formatting and coverage

phlptp:
@henryiii I think this is ready now

henryiii (approved):
Looks good to me! One bool -> auto I'm not really fond of, but it's okay.

PR #497, opened 2020-08-07
Customize text between option name and description

jakoblover:
Not sure if "option_text" is the best description of this text, but this PR allows you to customize this text using `->option_text("(ARG)")` for example. As of now, this text is generated automatically and can't be modified.


henryiii:
To verify, you want more customization than is possible with `type_name()`?

jakoblover:
Yep. Type_name only sets the type before the = sign. I want to customize that text completely, because I want to be in control over what is displayed there. I want a different format that is not trivial to change programatically in the formatter

henryiii (approved):

PR #499, opened 2020-08-12
Clarify inclusion of CLI11 via CMake

### LINKED ISSUE #498: Install instructions CMake
I use CLI11 in many small independent projects (folders). Thus, I was trying to install it as a shared folder accessible by other users, for example /opt/CLI11

Then, in my projects, I do:
`find_package(CLI11 CONFIG REQUIRED)`
as suggested in the Readme, specifying the proper CLI11_DIR path.

However, I get the following error:
`CLI11Config.cmake not found.`

If I check the build directory, only the following exist:
`CLI11ConfigVersion.cmake  CLI11Targets.cmake  CMakeDoxygenDefaults.cmake  cmake_install.cmake  CPackConfig.cmake  CPackSourceConfig.cmake  CTestCustom.cmake`

Am I doing something wrong? (I think the README.md is a bit unspecific in this respect.)

ferdymercury:
Installing is not optional if you want find_package to work
Specify other options when using shared folders (system-wide) rather than local ones.
Fixes https://github.com/CLIUtils/CLI11/issues/498

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,11 +133,25 @@ There are some other possible "features" that are intentionally not supported by
 
 ## Install
 
-To use, there are two methods:
+To use, there are several methods:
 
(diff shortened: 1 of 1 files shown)


henryiii (approved):
Looks good to me. Thanks!

PR #509, opened 2020-09-03
fix name of make make option opts function

mbhall88:
The function `make_opts` does not exist. I think what the docs were implying though was `make_option_opts`? At least using that name gave me the functionality the docs were talking about.


henryiii:
Looks like a typo to me, thanks!

henryiii:
@all-contributors please add @mbhall88 for doc

PR #516, opened 2020-09-11
Print input when IsMember validation fails

### LINKED ISSUE #515: Show value passed for failing `isMember` validation.
Currently, when the membership check fails and CLI11 reports the error, the actual value is not shown but only ` not in {...}`.
Change should be here, I guess. I can open a pull request in the upcoming days when I'm on a machine with git.

https://github.com/CLIUtils/CLI11/blob/c1666528ef0b1b5ddeacc60fba0f01f67cc7c7e8/include/CLI/Validators.hpp#L749

Dr-ZeeD:
Fixes #515

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -746,9 +746,7 @@ class IsMember : public Validator {
             }
 
             // If you reach this point, the result was not found
-            std::string out(" not in ");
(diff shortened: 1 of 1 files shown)


henryiii (approved):

henryiii:
Looks good to me. :) Thanks!

PR #520, opened 2020-10-01
Support for Atomic types in flags and options

### LINKED ISSUE #496: add_flag doesn't work well with std::atomic<bool>
ran into an issue with the add_flag not handling atomic<bool> properly.  
It seems to trigger a deleted assignment operator in atomic<bool>.

I made a workaround in the application for now but I will be looking into this soon.

phlptp:
Fixes #496 

dealing with atomic types is kind of a pain.  

This PR add support for them.  And add the tests for them as well.  

The main issue is that atomic types don't support direct assignment only assignment of the underlying type so there was a need to distinguish between the two type in a couple different locations that utilize the wrapper type.


henryiii (approved):

henryiii:
Thanks!

PR #521, opened 2020-10-10
Don't copy ConfigItems when iterating

alexdewar:
Use a const reference to avoid unnecessary copies. Also constify the
argument to _parse_config.


henryiii:
Sorry for the delay, I thought I merged this...

henryiii:
Thanks!

PR #523, opened 2020-10-15
create pkg-config file

djerius:
Create a pkg-config metadata file so that CLI11 can be found by pkg-config

### DIFF
--- a/cmake/CLI11GeneratePkgConfig.cmake
+++ b/cmake/CLI11GeneratePkgConfig.cmake
@@ -1,0 +1,6 @@
+configure_file("cmake/CLI11.pc.in" "CLI11.pc" @ONLY)
+
+install(FILES "${PROJECT_BINARY_DIR}/CLI11.pc"
+  DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
+
+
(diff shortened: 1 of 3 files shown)


cmake/CLI11GeneratePkgConfig.cmake
```diff
@@ -0,0 +1,9 @@
+set(DEST_DIR "${CMAKE_INSTALL_PREFIX}")
+set(PACKAGE_VERSION "${VERSION_STRING}")
+
+CONFIGURE_FILE("cmake/CLI11.pc.in" "CLI11.pc" @ONLY)
```
henryiii:
```suggestion
configure_file("cmake/CLI11.pc.in" "CLI11.pc" @ONLY)
```

cmake/CLI11GeneratePkgConfig.cmake
```diff
@@ -0,0 +1,9 @@
+set(DEST_DIR "${CMAKE_INSTALL_PREFIX}")
+set(PACKAGE_VERSION "${VERSION_STRING}")
```
henryiii: Why not use `@PROJECT_VERSION@` (which is automatically set by CMake), and just use it inside the `.pc.in` file? And use `@CMAKE_INSTALL_PREFIX@` too? If this is a convention, it could stay that way, but others, it seems to be simpler without adding new variables?

djerius: Just ignorance on my part of CMake.  I've pushed a version with your suggestions.

henryiii:
Thanks!

Issue #492, opened 2020-07-29
More convenient type deduction for parameters


ulfllorenz:
Currently, when I want some type safety, I have to write something like that:

`
int myParameter;
myApp.add_option("--foo", myParameter, "");
`

This is a bit annoying, because I have to write two lines of code instead of one; in general, I would prefer to use the option directly, but as far as I understand, this is the syntax needed to get CLI11 to do the cast checking. Is there a concise alternative solution possible, at least for simple (numeric) types; maybe an obvious extension would be user-defined types with a certain notion of well-behaved.

As a related issue, and possibly a solution, the various number checkers are not typed. For example, CLI11::PositiveNumber reads in an argument as a double, but I happen to require an int. So in addition, it would be nice to have the various checkers typed properly, maybe even have a type checker to skip the explicit adding of a variable reference.

phlptp:
I think this has been addressed with the addition of the type validator.  So I am going to close it.

PR #526, opened 2020-10-21
add a type validator 

phlptp:
add a typed Validator that checks for specific types conversion.  This could be useful if a storage variable was not used for the option but the result needs to be a specific type.

This is to address the issue raised in #492

Also move the number validator to use the typed checker instead of a separate class.

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -419,53 +419,6 @@ class IPV4Validator : public Validator {
     }
 };
 
-/// Validate the argument is a number and greater than 0
-class PositiveNumber : public Validator {
-  public:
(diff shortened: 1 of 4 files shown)


include/CLI/Validators.hpp
```diff
@@ -451,7 +451,7 @@ class NonNegativeNumber : public Validator {
         };
     }
 };
-
+/*
```
henryiii: Can we just delete this, please? :)

phlptp:
@henryiii  I think this one is ready too

henryiii (approved):
Looks good, I like that a few of the existing validators can use it instead.

Issue #522, opened 2020-10-10
Help subcommand 


Leon0402:
Hi, 

I was wondering if it's currently possible to add a help subcommand rather than a help flag (or in addition to it). Manually adding it doesn't work: 

```
auto helpCommand = parser.add_subcommand("help", "Shows the help-page");
helpCommand->callback([this]() {
    std::cout << this->parser.help();
});
```
This will execute basically `Program help --help`, so the help menu of the subcommand help. 

To give a prominent example, where such an approach is used: git 
Instead of `git --help` you can execute `git help`. 

It would be nice if this help subcommand worked in the same way as the flag, so: 
`Program subcommand --help` could be written as `Program help subcommand`
Git doesn't offer this feature, but something similar. You can for instance get more detailed information on a topic (and not on a subcommand) for instance `git help workflow`, while `git help commit` doesn't work. Providing this option as well might be interesting as well, but I personally don't need it and I guess this would require more effort to implement as well. 

Thanks!

phlptp:
I don't know why this wouldn't work,  what goes wrong with the code above?    When using it like that you will need to make sure the lifetime of the this pointer at least matches that of the parser that is executing the parse.  

If you just wanted it to behave exactly like call for help I would probably just do
```
auto helpCommand = parser.add_subcommand("help", "Shows the help-page");
helpCommand->callback([]() {
    throw CLI::CallForHelp();
});
```

or even 
```
auto helpCommand = parser.add_subcommand("help", "Shows the help-page");
helpCommand->preparse_callback([](std::size_t /*unused*/) {
    throw CLI::CallForHelp();
});
```

which will short circuit any further processing.

Then CLI11_PARSE will work identically to the --help built in function

Leon0402:
Hi, 

thanks for the comment! As I said, it doesn't work (at least the way I do it), because it's inside a subcommand.

Imagine what would happen if you call "MyApp subcommand --help". It wouldn't show the complete help menu, but just the part relevant parts for that subcommand.
The same happens here. The command "MyApp help" is "translated "to ""MyApp help --help" rather than just "MyApp --help". 

The output is: 
```
Shows the help-page
Usage: MyProgramm [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
```
So as you can see it shows above the "Usage" part the message of the help subcommand rather than the description. And it also doesn't show any subcommand (which is exactly what happens if you call "MyProgram subcommand --help"))

[...]

PR #529, opened 2020-10-22
Silent subcommands

phlptp:
add a silent option to subcommands to prevent the use from showing up in the subcommands list if used.

This is in response to Issue #522.

Adds an ability to silence subcommands and prevent them from showing up in the used subcommands list.
This allows things like

```c++
    auto sub1 = app.add_subcommand("help")->silent();
    sub1->parse_complete_callback([]() { throw CLI::CallForHelp(); });
```


book/chapters/subcommands.md
```diff
@@ -112,3 +112,18 @@ Here, `--shared_flag` was set on the main app, and on the command line it "falls
 
 This is a special mode that allows "prefix" commands, where the parsing completely stops when it gets to an unknown option. Further unknown options are ignored, even if they could match. Git is the traditional example for prefix commands; if you run git with an unknown subcommand, like "`git thing`", it then calls another command called "`git-thing`" with the remaining options intact.
 
+### Silent subcommands
+
+Subcommands can be modified by using the `silent` option.  This will prevent the subcommand from showing up in the get_subcommands list.  This can be used to make subcommands into modifiers For example a help subcommand might look like
```
henryiii:
```suggestion
Subcommands can be modified by using the `silent` option.  This will prevent the subcommand from showing up in the get_subcommands list.  This can be used to make subcommands into modifiers. For example, a help subcommand might look like
```

include/CLI/App.hpp
```diff
@@ -222,7 +222,11 @@ class App {
     App *parent_{nullptr};
 
     /// Counts the number of times this command/subcommand was parsed
-    std::size_t parsed_{0};
+    std::uint32_t parsed_{0U};
+
+    /// indicator that the subcommand is silent and won't show up in subcommands list
+    /// This is potentially useful as a modifier subcommand
+    bool silent_{false};
```
henryiii, line 223: This should be moved up to be with the other 8-bit members; if bools are placed together the compiler is more likely to pack them into 8 bits each, rather than adding alignment spacing. Technically, we probably should try to order the struct members large to small, but optimizing this is not really that important - it's just nice to do it right when we add something.

include/CLI/App.hpp
```diff
@@ -2390,8 +2404,8 @@ class App {
     ///
     /// If this has more than one dot.separated.name, go into the subcommand matching it
     /// Returns true if it managed to find the option, if false you'll need to remove the arg manually.
-    void _parse_config(const std::vector<ConfigItem> &args) {
-        for(const ConfigItem &item : args) {
+    void _parse_config(std::vector<ConfigItem> &args) {
+        for(ConfigItem item : args) {
```
henryiii: Why this change?

phlptp: recommended by clang-tidy or some other static analysis tools I ran

include/CLI/App.hpp
phlptp: I was trying to put it into a gap in alignment, I rearranged a bit

include/CLI/App.hpp
phlptp: I am going to change it back, it might have been a merge conflict or something

phlptp:
I think this is ready now @henryiii 
I also upgraded a version number with the cuda build for a github change

henryiii (approved):

PR #533, opened 2020-10-30
Fix variable name.

dbarowy:
Unless there is more CLI11 magic than I know about, the variable name in the example is a typo.


henryiii:
@all-contributors please add @dbarowy for docs

PR #542, opened 2020-11-25
Add some extra tests from debugging

phlptp:
This PR adds some extra tests cases that passed right away, but are related to various issues that have come up


henryiii (approved):
Extra tests are always welcome - if you want, feel free to merge something like this without review. :)

PR #584, opened 2021-05-06
fix: remove duplicated call to subcommand's final_callback

### LINKED ISSUE #572: final_callback of a nested subcommand will be called twice if parse_complete_callback exist
```c++
#include "CLI11.hpp"

    CLI::App app("Vision Application");
    app.set_help_all_flag("--help-all", "Expand all help");
    app.add_flag("--version", "Get version");

    CLI::App *cameraApp = app.add_subcommand("camera", "Configure the app camera");
    //cameraApp->require_subcommand(0, 1);  // 0 (default) or 1 camera

    cameraApp->parse_complete_callback( []() {
        std::cout << " camera's parse complete callback" << std::endl;
        } );

    //std::string mvcamera_config_file = "mvcamera_config.json";
    CLI::App *mvcameraApp = cameraApp->add_subcommand("mvcamera", "MatrixVision Camera Configuration");
    //mvcameraApp->add_option("-c,--config", mvcamera_config_file, "Config filename", true)->check(CLI::ExistingFile);
    //std::string mock_camera_path;
    //mockcameraApp->add_option("-p,--path", mock_camera_path, "Path")->required()->check(CLI::ExistingPath);

    CLI::App *mockcameraApp = cameraApp->add_subcommand("mock", "Mock Camera Configuration")
        ->final_callback( []() {
            std::cout << "mock's final callback" << std::endl;
            } );

    CLI11_PARSE(app, argc, argv);
}
```

**_Execution:_**
./nested camera mock
**_Output:_**
camera's parse complete callback
mock's final callback
mock's final callback

The final_callback of mockcameraApp will be called twice under this case. But if I remove the parse_complete_callback of cameraApp, the final_callback of mockcameraApp will only be called once as expected.

dryleev:
Fixes #572.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -113,6 +113,7 @@ class App {
 
     /// This is a function that runs when parsing has finished.
     std::function<void()> parse_complete_callback_{};
+
     /// This is a function that runs when all processing has completed
     std::function<void()> final_callback_{};
 
@@ -1963,6 +1964,7 @@ class App {
             app->_configure();
         }
     }
+
     /// Internal function to run (App) callback, bottom up
     void run_callback(bool final_mode = false, bool suppress_final_callback = false) {
         pre_callback();
@@ -1983,7 +1985,7 @@ class App {
 
         // finally run the main callback
         if(final_callback_ && (parsed_ > 0) && (!suppress_final_callback)) {
-            if(!name_.empty() || count_all() > 0) {
+            if(!name_.empty() || count_all() > 0 || parent_ == nullptr) {
                 final_callback_();
             }
         }
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1964,7 +1964,7 @@ class App {
         }
     }
     /// Internal function to run (App) callback, bottom up
-    void run_callback(bool final_mode = false) {
+    void run_callback(bool final_mode = false, bool suppress_final_callback = false) {
         pre_callback();
         // in the main app if immediate_callback_ is set it runs the main callback before the used subcommands
         if(!final_mode && parse_complete_callback_) {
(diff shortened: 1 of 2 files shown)


henryiii:
Nice, thanks! Could you add a little test for this? (I can add it later if not, though it might be able to be merged faster if you add it).

henryiii:
Added tests, and fixed another bug, I think - if this is the main app, the `final_callback` should always run. Currently it was only running if at least one more subcommand was passed. @phlptp could you verify that this looks correct?

include/CLI/App.hpp
```diff
@@ -1963,27 +1964,28 @@ class App {
...
-                subc->run_callback(true);
+                subc->run_callback(true, suppress_final_callback);
             }
         }
 
         // finally run the main callback
-        if(final_callback_ && (parsed_ > 0)) {
-            if(!name_.empty() || count_all() > 0) {
+        if(final_callback_ && (parsed_ > 0) && (!suppress_final_callback)) {
+            if(!name_.empty() || count_all() > 0 || parent_ == nullptr) {
```
henryiii, line 1988: Here, specifically, final_callback wasn't always running on the main app.

tests/SubcommandTest.cpp
```diff
@@ -1862,3 +1862,63 @@ TEST_CASE_METHOD(ManySubcommands, "defaultEnabledSubcommand", "[subcom]") {
...
+
+    subsub->parse_complete_callback([&subsub_compl]() { subsub_compl++; });
+    subsub->final_callback([&subsub_final]() { subsub_final++; });
+
+    SECTION("No specified subcommands") {
+        args = {};
+        run();
+
+        CHECK(app_compl == 1);
+        CHECK(app_final == 1);
```
henryiii, line 1894: Without the above fix, this is 0, not 1. The rest of them were correct.

PR #592, opened 2021-06-02
fix: commenting of descriptions with embedded newlines

paddy-hack:
Anything written in `write_descriptions`-guarded code blocks should assume that whatever text is provided by library users may contain embedded newlines.  This fixes the places where this wasn't done.

### DIFF
--- a/include/CLI/Config.hpp
+++ b/include/CLI/Config.hpp
@@ -278,7 +278,7 @@ ConfigBase::to_config(const App *app, bool default_also, bool write_description,
     bool defaultUsed = false;
     groups.insert(groups.begin(), std::string("Options"));
     if(write_description && (app->get_configurable() || app->get_parent() == nullptr || app->get_name().empty())) {
-        out << commentLead << app->get_description() << '\n';
+        out << commentLead << detail::fix_newlines(commentLead, app->get_description()) << '\n';
     }
     for(auto &group : groups) {
         if(group == "Options" || group.empty()) {
--- a/tests/ConfigFileTest.cpp
+++ b/tests/ConfigFileTest.cpp
@@ -1705,6 +1705,16 @@ TEST_CASE_METHOD(TApp, "TomlOutputHiddenOptions", "[config]") {
     CHECK(std::string::npos == loc);
 }
 
+TEST_CASE_METHOD(TApp, "TomlOutputAppMultiLineDescription", "[config]") {
+    app.description("Some short app description.\n"
+                    "That has multiple lines.");
+    run();
+
+    std::string str = app.config_to_str(true, true);
+    CHECK_THAT(str, Contains("# Some short app description.\n"));
+    CHECK_THAT(str, Contains("# That has multiple lines.\n"));
+}
+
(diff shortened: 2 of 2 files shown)


phlptp (approved):

paddy-hack:
FYI, the failed appveyor job appears to have been caused by a failed download of a build requisite, `catch.hpp`.  I have no trouble downloading it manually.  The `sha256sum` of my download matches what the build expects ...

henryiii:
Could you add a little test for this? Otherwise, I will, but it will likely get merged faster if you add one. There's a file with config tests in it.

paddy-hack:
Although I'm pretty sure I could implement a few tests, I am pressed for time and wouldn't know where to start to begin with.  I guess things will move along faster if you add the tests.

paddy-hack:
Just noticed the merge conflict.  Appears to have been introduced due to #593 submitted by me :open_mouth:  Should be straightforward to fix.

henryiii:
Can you paste an example config file to use in test, possibly? I don't mind making the test, but having an example file will help.

paddy-hack:
I had opportunity to fiddle with my application's command line so put some time aside to add tests for my changes.  Looking over a few existing tests was enough to get me up to speed :sweat_smile: 

Adding tests, I discovered that only the first hunk of my merge request is strictly needed.  It will address all multi-line *descriptions*.  The other two hunks address multi-line *group names* and multi-line *subcommands*.  The latter would be extremely cumbersome to use on the command-line.  The former, multi-line group names, is very unlikely.

This made me wonder what to do.  Currently, one is able to set multi-line group names and subcommands.  If one does, `config_to_str()` produces content that is *not* usable as a configuration file.  Should I just leave the `fix_newlines()` in or would it be better to add validation that prevents group names and subcommands from being multi-line strings?

paddy-hack:
Oops!  Just noticed that subcommands already check for embedded newlines and throw when they find one.  That leaves group names.  I'm inclined to add validation for those rather than allow embedded newlines.  What do you think?

phlptp:
My inclination would be to put the same check on group names as is on subcommands.

paddy-hack:
A quick search shows that `App::group()` and `OptionBase::group()` need to add validation.  Short of more or less copy-and-pasting the validation from `App::add_subcommand()`, any suggestions to implement this without code duplication?

paddy-hack:
Hmm, group names in the tests use spaces.  These aren't allowed in subcommands so that needs a bit of extra tinkering ...  
Working on it, will probably end up using copy-and-paste and leave it for others to refactor away the duplication :wink:

paddy-hack:
I'll do the group name validation in a separate PR.

phlptp:
we had some discussion earlier about relaxing the restrictions on subcommand names, to allow other characters,   basically everything but `=:!{}` and newlines and not starting with `/` or `-`.   So I am guessing we can merge the checks at that point.

paddy-hack:
OK.  Then I won't be submitting a group name validation PR.  Saves me some work :wink:

henryiii:
Fantastic, thank you!

Issue #537, opened 2020-11-18
config_to_str() ignores options with empty strings


andioz:
Hi there,

I'm experimenting with CLI11 version 1.9.1. I used boost::program_options for some years now and I'm looking for something smarter, with less cryptic syntax and less boilerplate code. Typically I offer a `--print` option in my programs, to print the current configuration to the console, and a `--config` option for reading such an config file in.

This works so far, but I have one problem, and cannot find a simple way to solve this. I want *all* options in the configuration file, including default values *and* empty strings. But the latter are not printed. Furthermore, there is a different format used for single character strings. Here is the `simple` example with my extensions: 

```c++
// Copyright (c) 2017-2020, University of Cincinnati, developed by Henry Schreiner
// under NSF AWARD 1414736 and by the respective contributors.
// All rights reserved.
//
// SPDX-License-Identifier: BSD-3-Clause

#include <CLI/CLI.hpp>
#include <iostream>
#include <string>

int main(int argc, char **argv) {

    CLI::App app("K3Pi goofit fitter");

    app.add_flag("-p,--print", "Print configuration and exit")->configurable(false); // NEW: print flag

    std::string file;
    CLI::Option *opt = app.add_option("-f,--file,file", file, "File name")->capture_default_str(); // NEW: capture_default_str()

    int count{0};
    CLI::Option *copt = app.add_option("-c,--count", count, "Counter")->capture_default_str(); // NEW: capture_default_str()

    int v{0};
    CLI::Option *flag = app.add_flag("--flag", v, "Some flag that can be passed multiple times")->capture_default_str(); // NEW: capture_default_str()

    double value{0.0};  // = 3.14;
    app.add_option("-d,--double", value, "Some Value")->capture_default_str(); // NEW: capture_default_str()

    CLI11_PARSE(app, argc, argv);
    
    if (app.get_option("--print")->as<bool>()) { // NEW: print configuration and exit
        std::cout << app.config_to_str(true, false);
        return 0;
    }

    std::cout << "Working on file: " << file << ", direct count: " << app.count("--file")
              << ", opt count: " << opt->count() << std::endl;
    std::cout << "Working on count: " << count << ", direct count: " << app.count("--count")
              << ", opt count: " << copt->count() << std::endl;
    std::cout << "Received flag: " << v << " (" << flag->count() << ") times\n";
    std::cout << "Some value: " << value << std::endl;

    return 0;
}
```
Here is the output for different values:

```
$ examples/simple -p # 1
[...]

PR #599, opened 2021-06-16
fix a few issues with config file output, 

phlptp:
namely an ability to print a default empty string, and to adjust the quote characters used when quoting strings or characters.

Address #537


henryiii:
I've added pre-commit.ci, so any fixable corrections should happen automatically now (though, sadly, no clang-format at the moment).

henryiii (approved):
Could you add a mention in the README?

PR #600, opened 2021-06-16
refactor!: remove add_complex

henryiii:
Followup to #597.


phlptp (approved):
I like being able to remove special cases!!!

Issue #553, opened 2021-01-25
cannot show help if config file is required


ntessore:
If a config file is required, it is impossible to print the help, even though the user is told to print the help.

```cpp
#include "CLI11.hpp"

int main(int argc, char* argv[])
{
    CLI::App app;

    app.set_config("--config")
        ->required();

    CLI11_PARSE(app, argc, argv);

    return 0;
}
```

```console
$ ./test
no specified config file was not readable (missing?)
Run with --help for more information.
$ ./test --help
no specified config file was not readable (missing?)
Run with --help for more information.
```

trumpetsven:
The Problem also appears for "--version".

PR #606, opened 2021-06-18
The help flag didn't get processed when a config file was required.

phlptp:
This fixes issue #553

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -2248,10 +2248,27 @@ class App {
 
     /// Process callbacks and such.
     void _process() {
-        _process_config_file();
-        _process_env();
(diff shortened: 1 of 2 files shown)


include/CLI/App.hpp
```diff
@@ -2248,10 +2248,27 @@ class App {
...
+        }
+        // callbacks and help_flags can generate exceptions which should take priority over the config file error if one
+        // exists
         _process_callbacks();
         _process_help_flags();
+
+        if(caught_error) {
+            throw CLI::FileError(std::move(fe));
+        }
+
```
henryiii, line 2271: I'd be tempted to use a unique_ptr here (really need an optional), but this should get refactored by the error handling update whenever that comes, so I think this is fine.

PR #611, opened 2021-07-12
chore: bump to 2.0.0

henryiii:
- chore: drop unused py file
- chore: bump version to 2.0.0


henryiii:
@phlptp Waiting for a sign-off from you that we are good to release.

phlptp (approved):
I am good with this.  The changelog and the changes here look good.    I haven't looked through the book completely yet, but that can wait I think If any changes are needed.    So I think it is time for a release.

PR #613, opened 2021-07-13
[doc] Updated documentation to include cmake info

### LINKED ISSUE #500: Usage with cmake
Hi, 

I just wanted to mention that CLI11 works perfectly with FetchContent of cmake, which is in my opinion the easiest and best way of inclusion. Presumably this could be added to the docs?

oparkins:
- Added example CMake instructions to fetch CLI11 and add it to a project.
- Added additional information about what header to use successfully.
- This is what works internally on some projects. Others may wish to test to make sure the CMake gods are satisfied. 

Closes CLIUtils/CLI11#500


henryiii:
I feel like this is a bit much for the readme, it would be better to put into the book and then link to it? Also not sure about recommending master, though we'd need to keep it up to date if we recommend a tag.

phlptp:
changes are included in #901

PR #617, opened 2021-07-14
docs: add more notes about releases

henryiii:
Added note about bug https://github.com/softprops/action-gh-release/issues/108 that wipes the release titles.


henryiii:
The upload is also broken, it's not adding CLI11.hpp. :(

henryiii:
Looks like after the move, there are no badges for Conan IO packages: https://github.com/conan-io/conan-center-index/issues/1042

PR #621, opened 2021-07-15
chore: codacity

henryiii:
@phlptp, you might want to review the last commit here, as some of those are touching things you've worked on.

- docs: fix some Codacity recommendations
- chore: update copyright year
- style: more codacity fixes
- style: fix issues reported by Codacity


include/CLI/App.hpp
```diff
@@ -1763,7 +1763,7 @@ class App {
         if(name_.empty()) {
             return std::string("[Option Group: ") + get_group() + "]";
         }
-        if(aliases_.empty() || !with_aliases || aliases_.empty()) {
+        if(aliases_.empty() || !with_aliases) {
```
henryiii, line 1766: Assuming this was not trying to check something else?

include/CLI/Validators.hpp
```diff
@@ -1113,7 +1113,6 @@ inline std::pair<std::string, std::string> split_program_name(std::string comman
                     esp = end + 1;
                     if(embeddedQuote) {
                         vals.first = find_and_replace(vals.first, std::string("\\") + keyChar, std::string(1, keyChar));
-                        embeddedQuote = false;
```
henryiii, line 1116: This is never used. The variable gets reinitialized to false after this ends.

tests/AppTest.cpp
```diff
@@ -1933,7 +1933,6 @@ TEST_CASE_METHOD(TApp, "AllowExtrasArgModify", "[app]") {
     app.allow_extras();
     app.add_option("-f", v2);
     args = {"27", "-f", "45", "-x"};
-    auto cargs = args;
```
henryiii, line 1936: Unused, hopefully this was not supposed to be used later for a missing test?

include/CLI/App.hpp
phlptp: I suspect it was copied and ended up duplicated condition

tests/AppTest.cpp
phlptp: Seeing how the method names is argModify I could see that being potentially useful, but it look like the test didn't end up using it and just making the check with a constructed vector.

tests/HelpersTest.cpp
```diff
@@ -998,6 +998,10 @@ TEST_CASE("Types: TypeName", "[helpers]") {
     CHECK(umapName == "[INT,[TEXT,FLOAT]]");
 
     vclass = CLI::detail::classify_object<std::atomic<int>>::value;
+    CHECK(CLI::detail::object_category::wrapper_value == vclass);
+
+    std::string atomic_name = CLI::detail::type_name<std::atomic<int>>();
+    CHECK(atomic_name == "INT");
```
henryiii: @phlptp This actually reports `CLI::detail::object_category::other` and "TEXT" on some compilers (older clang and GCC). Any quick thoughts on if that is easily fixable / important?

phlptp: Things got a little weird with older compilers and atomics and the template checks.   I think it still works when actually used just in a roundabout fashion through the stream operators and I am not sure the effort to fix the template names for those compilers is worth it.

henryiii: Okay, we can just check for both then.

PR #623, opened 2021-07-19
style: markdown checking

henryiii:
This adds a couple of markdown checkers that should help keep most of the issues Sourcery is seeing down. Quite a few fixes.

- style: add mdlint
- style: add remarklint


henryiii:
Manually adding remarklint until an official hook is added https://github.com/remarkjs/remark-lint/issues/227.

Issue #580, opened 2021-04-21
Crash when using a slash character as the name of a subcommand


suzukieng:
Just stumbled upon this, while migrating a HTTP/REST based API to a CLI utility (yes, these kinds of projects exist).

I attempted to name a subcommand "legacy/status" and got a crash (Visual Studio 10):

```
Unhandled exception at 0x759FA6E2 in xxx.exe: Microsoft C++ exception: CLI::IncorrectConstruction at memory location 0x010AF178.
```

Regardless of it being a good or bad idea to have a slash as part of a command name, a better error message or something in the (excellent!) documentation might help.

phlptp:
I am going through some of the issues today,  I am going to try to see what is up with this one and fix it if possible or at least make a better error message

henryiii:
I've added a "triage" label for things that are likely bugs and should be checked before release, but not verified quite yet (say by running locally or adding a test). Feel free to add that to any issues you see that look like they need to be addressed but you don't have time to do it ATM. :) If verified, they can become "bug".

phlptp:
I don't think allowing a slash in general is a good idea.  It can be used in shells and reinterpreted.  That being said there are applications where it and other characters could be useful and allowed in options and subcommands.  

Not sure what a good way of allowing that in code without library modification yet.  
But the code that controls the validity is 

```cpp
/// Verify the first character of an option
template <typename T> bool valid_first_char(T c) {
    return std::isalnum(c, std::locale()) || c == '_' || c == '?' || c == '@';
}

/// Verify following characters of an option
template <typename T> bool valid_later_char(T c) { return valid_first_char(c) || c == '.' || c == '-'; }
```

It would be possible to add additional characters in that part of the code to allow other characters in a slightly modified library

henryiii:
Would there be harm in including slashes in `valid_later_char`?

phlptp:
can't use `'\'` since that is an escape character on many shells.  `'/'` as a first character can't be allowed since that is often used on windows systems as an argument.    But as a later character nothing pops out as a reason why it can't be allowed.  

If we want to I can add it pretty quickly

henryiii:
If someone passes `\n`, then we get the escaped version. If someone passes `\\`, we see `\`. So on _our_ side, it doesn't seem like we should necessarily disallow it? Agree about the first character, just thinking for later characters. Also for a space, such as `' '`, one could write `./app 'my command'`, which would be a terrible CLI, but technically valid.

[...]

PR #627, opened 2021-07-21
Relaxed option naming

phlptp:
If merged this pull request will relax the restrictions on option naming and subcommand naming.  

Allowing inclusion of all characters except
starting with:

- `-` since we need to recognize `-` and `--` as different options and `--` has some special meaning
- `\n` since that would cause all sorts of chaos with the formatters and other things
- `!`  used as a special character for setting default values
- ` `  allowing this would create the potential for difficult to debug user entry errors for example `"-- input, input"` would be adding an extra space accidentally, But if space were allowed this would read fine as a single almost impossible to use option.  

and anywhere in the option
- `\n`  same as above
- `=`, and `:`  they are fine as a first character but after that they are used as a value separator so can't be allowed in an option
- `{`  `{}` is used to indicate default values
- ` `  allowing this would create the potential for difficult to debug user entry errors for example `"--input input"` would be forgetting a comma to separate an option from the positional.  But if space were allowed this would read fine as a single almost impossible to use option.  

this will address #580 
also allow lots of amusing things like `--:)`  as an option

A couple questions in #592 we talked about adding a check on group_names and a few other things.  particularly in regard to newlines.  I need to do that yet.  

But in regard to subcommands  through the `add_subcommand` I am inclined to leave the check for spaces in the subcommand name in place.  The reason has to do with error catching.    Most subcommand names don't have spaces but their description does quite frequently.  Meaning the check on spaces catches some potential errors in the API call.    Removing this restriction would likely cause additional errors that might be hard for a user to spot initially.    

Then the question is should we allow newlines in App names?    If not might be worth adding an additional validator to catch that and use it for app names, aliases, and group names.


phlptp:
TODO
- [x] readme updates
- [x] additional test cases for subcommands
- [x] validators for group names, app name, aliases
- [x]  book updates

phlptp:
Added a validator for group_names, aliases, and option group names.  But not app names.  App names can still have new lines, so it is possible to get a subcommand with a new line but not easy.    Should we restrict this as well?

book/chapters/flags.md
```diff
@@ -120,4 +120,4 @@ Flag int: 3
 Flag plain: 1
 ```
 
-[^1]: It will not inherit this from the parent defaults, since this is often useful even if you don't want all options to allow multiple passed options.
+\[^1\]: It will not inherit this from the parent defaults, since this is often useful even if you don't want all options to allow multiple passed options.
```
henryiii, line 123: Guessing this was required for some reason?

book/chapters/options.md
```diff
@@ -26,12 +26,12 @@ You can use any C++ int-like type, not just `int`. CLI11 understands the followi
 | complex-number | std::complex or any type which has a real(), and imag() operations available, will allow 1 or 2 string definitions like "1+2j" or two arguments "1","2" |
 | enumeration | any enum or enum class type is supported through conversion from the underlying type(typically int, though it can be specified otherwise) |
 | container-like | a container(like vector) of any available types including other containers |
-| wrapper | any other object with a `value_type` static definition where the type specified by `value_type` is one of type in this list |
+| wrapper | any other object with a `value_type` static definition where the type specified by `value_type` is one of the type in this list, including std::atomic<>|
```
henryiii:
```suggestion
| wrapper | any other object with a `value_type` static definition where the type specified by `value_type` is one of the type in this list, including `std::atomic<>` |
```

include/CLI/App.hpp
```diff
@@ -978,13 +967,13 @@ class App {
         if(!subcommand_name.empty() && !detail::valid_name_string(subcommand_name)) {
             if(!detail::valid_first_char(subcommand_name[0])) {
                 throw IncorrectConstruction(
-                    "Subcommand name starts with invalid character, allowed characters are [a-zA-z0-9]+'_','?','@' ");
+                    "Subcommand name starts with invalid character, '!' and'-' are not allowed");
```
```suggestion
                    "Subcommand name starts with invalid character, '!' and '-' are not allowed");
```

include/CLI/Option.hpp
```diff
@@ -94,6 +94,9 @@ template <typename CRTP> class OptionBase {
 
     /// Changes the group membership
     CRTP *group(const std::string &name) {
+        if(!detail::valid_alias_name_string(name)) {
+            throw IncorrectConstruction("group names may not contain newlines or null characters");
```
henryiii:
```suggestion
            throw IncorrectConstruction("Group names may not contain newlines or null characters");
```

book/chapters/flags.md
phlptp: In some parsers it interprets that as a link and give syntax warnings

PR #631, opened 2021-08-02
add a more clear force callback

phlptp:
and callback on parse modifier for options.

The method of forcing a callback that exists is not clear from the issues raised around it.  This should clarify that ability and make it more explicit.  It also adds an ability to trigger the callback for each parse of an option.    This opens up new avenues of processing and CLI modification as well as inline validation.


phlptp:
TODO:
- [x] test cases
- [x] docs
- [x] book updates
- [x] pass checks

include/CLI/Option.hpp
```diff
@@ -666,7 +685,7 @@ class Option : public OptionBase<Option> {
     /// The maximum number of arguments the option expects
     int get_type_size_max() const { return type_size_max_; }
 
-    /// The number of arguments the option expects
+    /// return the inject_separator flag
```
henryiii:
```suggestion
    /// Return the inject_separator flag
```
```suggestion
    /// return the inject_separator flag
```

henryiii (approved):

henryiii:
WIP icons missed, added in #638.

PR #633, opened 2021-08-10
only add CONFIGURE_DEPENDS if main project

### LINKED ISSUE #632: Do not define CONFIGURE_DEPENDS for library usage
If you include CLI11 in projects, it is a bit confusing, that it uses `CONFIGURE_DEPENDS` in its globbing, as normally the user do not add files to libraries. I would suggest adding the idiomatic check `IF (CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)` before the for-loop, to only "reglob", if this is the main project.

https://github.com/CLIUtils/CLI11/blob/4d34311cb16a42487c9f1ce587f25ebada5a2bb3/CMakeLists.txt#L157-L163

BenjaminBeichler:
- reglobbing of files is confusing, if CLI11 is used as library, as typically the files won't change

fixes #632


henryiii (approved):

henryiii:
Thanks!

PR #642, opened 2021-09-23
max/min macro conflict

phlptp:
This PR fixes some potential conflicts when using some validators with some windows headers which define max and min as macros.

### DIFF
--- a/include/CLI/Validators.hpp
+++ b/include/CLI/Validators.hpp
@@ -467,35 +467,35 @@ class Range : public Validator {
     /// Note that the constructor is templated, but the struct is not, so C++17 is not
     /// needed to provide nice syntax for Range(a,b).
     template <typename T>
-    Range(T min, T max, const std::string &validator_name = std::string{}) : Validator(validator_name) {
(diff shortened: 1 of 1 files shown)


phlptp:
With the 2.1 release I could finally get rid of my custom version of CLI11 in HELICS.  Ran into the issue of macros on Windows which this PR fixes.

include/CLI/Validators.hpp
```diff
@@ -467,35 +467,35 @@ class Range : public Validator {
     /// Note that the constructor is templated, but the struct is not, so C++17 is not
     /// needed to provide nice syntax for Range(a,b).
     template <typename T>
-    Range(T min, T max, const std::string &validator_name = std::string{}) : Validator(validator_name) {
+    Range(T minVal, T maxVal, const std::string &validator_name = std::string{}) : Validator(validator_name) {
```
```suggestion
    Range(T min_val, T max_val, const std::string &validator_name = std::string{}) : Validator(validator_name) {
```
(And for the others). I think this is more consistent.

phlptp: made the change

Issue #648, opened 2021-10-02
config parsing crashes on empty arrays


puchneiner:
We can take this test https://github.com/CLIUtils/CLI11/blob/8b785a6c7d5b0af5f3ba0003d7a9e43d08d6f3be/tests/ConfigFileTest.cpp#L1009-L1028

then add
```c++
out << "four=[]\n";
```
and
```c++
CHECK(four == std::vector<std::string>({}));
```

The test won't pass, application crashes with segmentation fault.

So any user can crash the app by making a pretty valid config file

kvtb:
changing `->required()` to `->take_all()->expected(0,999)` does not help.
although, I think, it should work with `->required()` too, as the option is provided, even if its value is an empty array

PR #649, opened 2021-10-03
add tests which suppose to pass

nora-puchreiner-ai:
There is no fix yet, just tests which the fixed version expected to pass
Related https://github.com/CLIUtils/CLI11/issues/648

### DIFF
--- a/tests/ConfigFileTest.cpp
+++ b/tests/ConfigFileTest.cpp
@@ -1021,7 +1021,7 @@ TEST_CASE_METHOD(TApp, "TOMLStringVector", "[config]") {
     }
 
     std::vector<std::string> zero, one, two, three;
-    app.add_option("--zero", zero)->required();
+    app.add_option("--zero", zero)->required()->expected(0,99);
     app.add_option("--one", one)->required();
     app.add_option("--two", two)->required();
     app.add_option("--three", three)->required();
@@ -1051,7 +1051,7 @@ TEST_CASE_METHOD(TApp, "IniVectorCsep", "[config]") {
     }
 
     std::vector<int> zero, one, two, three;
-    app.add_option("--zero", zero)->required();
+    app.add_option("--zero", zero)->required()->expected(0,99);
     app.add_option("--one", one)->required();
     app.add_option("--two", two)->expected(2)->required();
     app.add_option("--three", three)->required();
--- a/tests/ConfigFileTest.cpp
+++ b/tests/ConfigFileTest.cpp
@@ -1051,8 +1051,8 @@ TEST_CASE_METHOD(TApp, "IniVectorCsep", "[config]") {
     }
 
     std::vector<int> zero, one, two, three;
-    app.add_option("--zero", three)->required();
-    app.add_option("--one", three)->required();
+    app.add_option("--zero", zero)->required();
+    app.add_option("--one", one)->required();
     app.add_option("--two", two)->expected(2)->required();
     app.add_option("--three", three)->required();
 
--- a/tests/ConfigFileTest.cpp
+++ b/tests/ConfigFileTest.cpp
@@ -1014,16 +1014,22 @@ TEST_CASE_METHOD(TApp, "TOMLStringVector", "[config]") {
         std::ofstream out{tmptoml};
         out << "#this is a comment line\n";
         out << "[default]\n";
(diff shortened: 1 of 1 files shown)


henryiii:
Great, thanks, this is the best way to file an issue (with a failing test to start the process)! I expect the issue partially comes from the fact it's hard/impossible to make this on the command-line; not sure what it means to pass an empty required option. (Though, you can do "flag-style" options in which case this would be possible). I expect we rely on the fact somewhere that there's at least one argument that needs to be fixed.

henryiii:
I'm seeing the message:

```
/Users/henryschreiner/git/software/CLI11/tests/ConfigFileTest.cpp:1007: FAILED:
--zero: At least 1 required but received 0
```

That seems like it is the right error if there is no way to pass an empty argument on the command line? Now if this is allowed with no argument, that should also be allowed to be empty in the TOML. Also, wasn't this a segfault at some point, or maybe it's just how I compiled it in Debug?

puchneiner:
Ok, I will add `->expected(0,99)` to the test.
it won't make them pass anyway.

I tried to fix it, but hit the same problem - zero-argument is often abused to differ flag and option, although there is a dedicated `flag_like_` field introduced for that purpose.
So it required minor refactoring, not just a fix.

puchneiner:
Yes, it is difficult to pass zero-array in command-line, and that is another issue - to invent a way to do it.
But CLI11 might be used as config-only parser, where limitations brought from the the realm of command line look strange

puchneiner:
Actually, I hit this problem parsing command-line, not the config :)
I am using 2 `CLI::App`, one for config file - with 100s of options and another for command-line, with the most crucial ones.
Also, command-line has the ability to prepend a virtual line to the config using `--option key value`
and `--option key ""` unexpectedly resulted in vector of size 1 containing an empty string.
Then I tried to discover how to set value to empty vector and fall to the abyss on how much lines have to be fixed :(

henryiii:
> and `--option key ""` unexpectedly resulted in vector of size 1 containing an empty string.

Why is this unexpected? That's exactly what you are telling it to do. If you want to strip empty items from your vector, you can do that. The only way to get an empty vector on the command line is to make this a "flag-option" hybrid (which is a bit ugly conceptually, but we support it for most uses). This is a weird use though; it's the only way that `count` would not equal the number of items in the vector. "requires" on top of this is really weird; you require it but you can't rely on it having anything in it?

It still sounds like there's a segfault lurking somewhere, have you been able to reproduce that?

kvtb:
> Why is this unexpected? That's exactly what you are telling it to do.

Because it prepends to config parser as `key=`
Config lines look like `key="hello" "world"` - 2-item array, `key="hello"` - 1-item array, `key=""` - 1-item array with empty string
Then the question which deserves to be in the IQ test card: how `key=` is to be parsed? I'd say to empty array.
But it is parsed as 1-item array with empty string, the same like `key=""`

phlptp:
from a parser perspective I would say `key=""` and `key=`   on a line have to be equivalent.   I am not sure `key=`  is technically valid toml but CLI is a bit more lax, and such the presence or absences of quotes should be ignored and handled equivalently.  The only way I would see to get an empty array would be to force the array syntax as `key=[]`

phlptp:
I poked at this a little, and basically what the parser is doing is stripping off the array indicators and splitting up the string based on the array separator char.  

Which essentially means `key="",  key=,  key=[]` are all equivalent.    having those be different would require empty tests for each of those conditions.  I think `key=""` and `key= `  have to be equivalent otherwise it is really inconsistent with passing an empty string.  

Now when the value You are asking CLI to parse is a vector,  that is a container of strings.  Each element is a string.  So either you get 1 or more strings in the vector or nothing was passed.  

the config parser has no knowledge of the ultimate conversion,  so we can't treat anything differently based on whether the output is vector or not,  so an array in config has to be equivalent to giving multiple arguments (either separately or at once on the command line).   

The only thing I could see doing differently in the config reader is 
for `key=` and `key=[]`  is to essentially ignore the argument and treat it as if it weren't passed.  Anything else would require inconsistencies in how values get treated which I suspect would lead to a whole host of unintended consequences.

henryiii:
I think this was fixed, so closing. Let me know if any tests, etc. are useful from here.

PR #660, opened 2021-10-24
Empty vector

phlptp:
this is a follow-on to #649.  

What I did was fix a few of the bugs in the parsing with flag like variables.  
add a check in the processing of vector outputs, that catches a 
"{}" as the empty vector and returns it.  

So in a config file `""`, `[]`, `{}`  will trigger the default value for flag like options( expected_min==0)  so to get an empty vector as an option for a given option.  the default value must be an empty vector, or the `default_str` set to "{}"  

There are also some checks to prevent an empty vector from getting generated on an output with a minimum expected greater than 0.  

this needs more testing to make sure it works from the command line as well, and I think there are a few edge cases that need to be dealt with yet, but @puchneiner is correct in that there does need to be a mechanism for generating empty vectors as results.  I don't think it needs to be easy, as if it was too easy it could cause seg faults if people are not expecting an empty vector to be allowed.    Basically if you are adding an option and do not explicitly specify a min expected as 0 and specify a default value as empty, there should be no way to generate an empty vector as an output.

### DIFF
--- a/tests/ConfigFileTest.cpp
+++ b/tests/ConfigFileTest.cpp
@@ -1021,7 +1021,7 @@ TEST_CASE_METHOD(TApp, "TOMLStringVector", "[config]") {
     }
 
     std::vector<std::string> zero, one, two, three;
-    app.add_option("--zero", zero)->required();
+    app.add_option("--zero", zero)->required()->expected(0,99);
     app.add_option("--one", one)->required();
     app.add_option("--two", two)->required();
     app.add_option("--three", three)->required();
@@ -1051,7 +1051,7 @@ TEST_CASE_METHOD(TApp, "IniVectorCsep", "[config]") {
     }
 
     std::vector<int> zero, one, two, three;
-    app.add_option("--zero", zero)->required();
+    app.add_option("--zero", zero)->required()->expected(0,99);
     app.add_option("--one", one)->required();
     app.add_option("--two", two)->expected(2)->required();
     app.add_option("--three", three)->required();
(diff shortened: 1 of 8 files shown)


tests/ConfigFileTest.cpp
```diff
@@ -1014,16 +1014,28 @@ TEST_CASE_METHOD(TApp, "TOMLStringVector", "[config]") {
...
+        out << "nzero={}\n";
+        out << "one=[\"1\"]\n";
         out << "two=[\"2\",\"3\"]\n";
         out << "three=[\"1\",\"2\",\"3\"]\n";
     }
 
-    std::vector<std::string> two, three;
+    std::vector<std::string> nzero, zero1, zero2, one, two, three;
+    app.add_option("--zero1", zero1)->required()->expected(0, 99)->default_str("{}");
+    app.add_option("--zero2", zero2)->required()->expected(0, 99)->default_val(std::vector<std::string>{});
```
henryiii, line 1028: What happens if you add `->expected(0, 99)` but don't set default_str? Can whatever happens for that be tested?

phlptp: I added a test for that.  But that is basically the previous behavior and will result in a vector with a single empty string.

phlptp: The idea here is that there does need to be a way to get an empty vector as an output, but it shouldn't be default or easy.

henryiii: Seems like adding an empty string from a TOML `[]` is also problematic. Would is be possible to make this an error if a default is not specified? `[""]` would be a valid default to get the old behavior?

henryiii:
At this point, I think this is a strict improvement, though making `[]`/`{}` an error rather than equivalent to `[""]` seems like a valid improvement eventually. Just need a rebase to fix the README conflict.

PR #663, opened 2021-10-26
run some tests and fixes for C++20

### LINKED ISSUE #662: add_flag is ambiguous with c++20
Code:
```
#include <CLI/CLI.hpp>

int main()
{
    auto app = CLI::App{"test"};
    auto flag = app.add_flag("--flag", "desc");
    return 0;
}

```

Compiling with
`g++ -std=c++20 cli11.cpp`

results in
```
cli11.cpp: In function 'int main()':
cli11.cpp:6:29: error: call of overloaded 'add_flag(const char [7], const char [5])' is ambiguous
```

CLI11 is version 2.1.2. g++ is 11.2.0.

phlptp:
Check #662 and fix a windows compile issue with C++20.

### DIFF
--- a/tests/OptionalTest.cpp
+++ b/tests/OptionalTest.cpp
@@ -239,8 +239,7 @@ TEST_CASE_METHOD(TApp, "BoostOptionalEnumTest", "[optional]") {
     auto dstring = optptr->get_default_str();
     CHECK(dstring.empty());
     run();
-    auto checkOpt = static_cast<bool>(opt);
-    CHECK(!checkOpt);
+    CHECK_FALSE(checkOpt);
 
     args = {"-v", "3"};
     run();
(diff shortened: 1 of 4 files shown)


tests/OptionalTest.cpp
```diff
@@ -239,16 +239,19 @@ TEST_CASE_METHOD(TApp, "BoostOptionalEnumTest", "[optional]") {
     auto dstring = optptr->get_default_str();
     CHECK(dstring.empty());
     run();
-    CHECK(!opt);
+    auto checkOpt = static_cast<bool>(opt);
+    CHECK(!checkOpt);
```
henryiii:
```suggestion
    CHECK_FALSE(checkOpt);
```
?

henryiii (approved):

Issue #551, opened 2021-01-21
map config groups to option groups instead of subcommands


ntessore:
Is there currently a way to map config file groups (i.e. `[mygroup]`) to option groups defined in an app?

Use case is a program with a number of subcommands that share a very large number of options, which are read from a config file, and are grouped for clarity.

phlptp:
I believe you can use the alias option on option_groups,  this will allow the group name to be used in the config.

```
app.add_option_group("ogroup")->alias("ogroup");
```

```
[default]
val=1
[ogroup]
val2=2
```

PR #666, opened 2021-11-03
add a few tests related to github issues

### LINKED ISSUE #556: gcc 9.3 warns with paranoid -Wstrict_overflow=5 about App::parse()
On a nearly empty sample
```
int main(int argc, const char* argv[]) {
    CLI::App app{"Sample"};
    try{
        app.parse(argc, argv);
    } catch (const CLI::ParseError &e) {
        return app.exit(e);
    }
```
gcc with `-O2` and `-Wstrict-overflow=5` complains:
```
[build] /home/vsol/repo/lfmb/build/RelWithDebInfo/_deps/cli-src/include/CLI/App.hpp:1222:10: error: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Werror=strict-overflow]
[build]  1222 |     void parse(int argc, const char *const *argv) {
[build]       |          ^~~~~
[build] /home/vsol/repo/lfmb/build/RelWithDebInfo/_deps/cli-src/include/CLI/App.hpp:1222:10: error: assuming signed overflow does not occur when changing X +- C1 cmp C2 to X cmp C2 -+ C1 [-Werror=strict-overflow]
```
The error position is confusing but actually it seems to be the loop:
```
        for(int i = argc - 1; i > 0; i--)
            args.emplace_back(argv[i]);
```
Possible fix is 
```
        for(auto i = static_cast<std::size_t>(argc - 1); i > 0u; i--)
            args.emplace_back(argv[i]);
```

### LINKED ISSUE #569: default value for flag is not shown in help
The default value for a flag is not shown in the help string. 

For example I have a flag ``--alignment`` with a negate flag ``--no-alignment`` and the help output is:

```
  --alignment,--no-alignment{false}
                              Do Alignment Processing
```

But there is nothing that shows the default value.

### LINKED ISSUE #669: Cannot compile with /GR- (RTTI disabled) and _HAS_STATIC_RTTI=0 with MSVC compiler
I get the following error compiling a CLI11 example with `/GR-` (RTTI disabled) and also with `-D_HAS_STATIC_RTTI=0`:

```
CLI11.hpp(6763,25): error C2280: 'std::shared_ptr<CLI::ConfigBase> std::dynamic_pointer_cast<CLI::ConfigBase,CLI::Config>(const std::shared_ptr<CLI::Config> &) noexcept': attempting to reference a deleted function
```

The condition check for whether to use a dynamic or static cast is the issue here:
https://github.com/CLIUtils/CLI11/blob/815553211b644b2a64b5dc671c602233c49c065f/include/CLI/App.hpp#L1546-L1550

MSVC does not appear to modify the value of [`__cpp_rtti`](https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#__cpp_rtti) which remains `1`, but rather sets [`_CPPRTTI`](https://docs.microsoft.com/en-us/cpp/build/reference/gr-enable-run-time-type-information) instead. Visual Studio's intellisense doesn't seem to correctly interpret this value but it does in fact change depending on `/GR[-]`.

If my understanding is correct, the condition for `_HAS_STATIC_RTTI` should also be checking for a value of `1`, not `0`.

### LINKED ISSUE #670: subcommand callback executes twice
The following code executes the g1c1 callback twice

```
#include <CLI/CLI.hpp>
#include <iostream>

int main(int argc, char* argv[]) {
  auto app = CLI::App{"An app."};
  auto group1 = app.add_option_group("FirstGroup", "First options group.");
  auto group2 = app.add_option_group("SecondGroup", "Second options group.");

  auto g1c1 = group1->add_subcommand("g1c1", "First group first sub-command")
                  ->callback([]() {
                    std::cout << "First group first sub-command executed.\n";
                  });

  auto g2c1 = group2->add_subcommand("g2c1", "Second group first sub-command");

  CLI11_PARSE(app, argc, argv);
  return 0;
}
```

This bug was reported by @bimalgaudel on [gitter](https://gitter.im/CLI11gitter/Lobby?at=618dd0d0fb8ca0572b1862e8)

phlptp:
add test related to issue #556,  #569, #551

Changes the way default_str is displayed in the help messages from "=XXXX" to "[XXXX]" to make it clearer what the purpose of the value is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,16 @@ is not passed, or every time the option is parsed.
 [#656]: https://github.com/CLIUtils/CLI11/pull/656
 [#657]: https://github.com/CLIUtils/CLI11/pull/657
 
+## Version 2.1.3: Bug Fixes and Tweaks
+
+* Change the way the default value is displayed in the included help text generation from `=XXXXX` to `[XXXXX]` to clean up some situations in which the text looked awkward and unclear [#666][]
+* Fix a bug where a subcommand callback could be executed multiple times if it was a member of an option group [#666][]
+* Fix an issue where the detection of RTTI being disabled on certain visual studio platforms did not disable the use of dynamic cast calls [#666][]
+* Add additional tests concerning the use of aliases for option groups in config files [#666][]
+* Resolve strict-overflow warning on some GCC compilers [#666][]
+
+[#666]: https://github.com/CLIUtils/CLI11/pull/666
+
 ## Version 2.0: Simplification
 
 This version focuses on cleaning up deprecated functionality, and some minor
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -102,6 +102,7 @@ jobs:
       gcc9:
         containerImage: gcc:9
         cli11.std: 17
+        cli11.options: -DCMAKE_CXX_FLAGS="-Wstrict-overflow=5"
       gcc11:
         containerImage: gcc:11
         cli11.std: 20
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1213,8 +1213,8 @@ class App {
         }
 
         std::vector<std::string> args;
-        args.reserve(static_cast<std::size_t>(argc) - 1);
-        for(int i = argc - 1; i > 0; i--)
+        args.reserve(static_cast<std::size_t>(argc) - 1U);
+        for(auto i = static_cast<std::size_t>(argc) - 1U; i > 0U; --i)
             args.emplace_back(argv[i]);
         parse(std::move(args));
     }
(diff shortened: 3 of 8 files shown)


phlptp:
Closes #556
Closes #569

phlptp:
Try to address #669

phlptp:
Closes #670

henryiii:
> Changes the way default_str is displayed in the help messages from "=XXXX" to "[default=XXXX]"

I don't like embedding an English label like this into CLI11; it's nice to avoid text that might need to be translated. Argparse uses "[XXXX]", that might be clearer that the current form?

henryiii (approved):
Would you mind summarizing the changes for the changlog? Changing the default help print, a fix related to RTTI, and a few extra tests?

henryiii (approved):

PR #668, opened 2021-11-06
add an option to validate optional arguments like in a vector.

phlptp:
This can resolve some issues with separating positionals from vector arguments


phlptp:
For example 
```
std::vector<std::string> vec;
    std::vector<int> ivec;
    app.add_option("pos", vec);
    app.add_option("--args", ivec)->check(CLI::Number);
    app.validate_optional_arguments();
```

can now parse 
`--args 1 2 three four`

properly withour error

henryiii (approved):
I worry about increasing the already quite large API surface, but this looks good. I'm guessing the reason (other than backward compatibility) that this would not be on by default or always on and not a setting would be the error messages are likely poorer with it on?

phlptp:
Having it on by default could lead to some confusing results, and it does add computation and the main reason it is needed is symmetry with vector positional arguments in the API, and for situations with positionals and vector arguments.

PR #678, opened 2021-11-27
fix: avoid a warning about useless move

### LINKED ISSUE #677: GCC warning useless-cast
When compiling with gcc and `-Wuseless-cast`, this triggers a warning (or error with `-Werror`).
```
CLI/App.hpp:2273:24: error: useless cast to type ‘class CLI::FileError’ [-Werror=useless-cast]
 2273 |             throw CLI::FileError(std::move(fe));
      |       
```
Just `throw fe;` should be fine.

henryiii:
Closes #677. This is a more straight forward way to do this, and not that much duplication (three lines shorter).

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -2252,26 +2252,23 @@ class App {
 
     /// Process callbacks and such.
     void _process() {
-        CLI::FileError fe("ne");
-        bool caught_error{false};
         try {
             // the config file might generate a FileError but that should not be processed until later in the process
             // to allow for help, version and other errors to generate first.
             _process_config_file();
+            
             // process env shouldn't throw but no reason to process it if config generated an error
             _process_env();
         } catch(const CLI::FileError &fe2) {
-            fe = fe2;
-            caught_error = true;
+            // callbacks and help_flags can generate exceptions which should take priority
+            // over the config file error if one exists.
+            _process_callbacks();
+            _process_help_flags();
+            throw;
         }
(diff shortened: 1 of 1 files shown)


include/CLI/App.hpp
```diff
@@ -2252,27 +2252,24 @@ class App {
...
             // process env shouldn't throw but no reason to process it if config generated an error
             _process_env();
         } catch(const CLI::FileError &fe2) {
-            fe = fe2;
-            caught_error = true;
+            // callbacks and help_flags can generate exceptions which should take priority
+            // over the config file error if one exists.
+            _process_callbacks();
+            _process_help_flags();
+            throw;
```
Henny022, line 2267: this should be `throw fe2;` to be equivalent to before

include/CLI/App.hpp
```diff
@@ -2252,27 +2252,24 @@ class App {
...
-        CLI::FileError fe("ne");
-        bool caught_error{false};
         try {
             // the config file might generate a FileError but that should not be processed until later in the process
             // to allow for help, version and other errors to generate first.
             _process_config_file();
+
             // process env shouldn't throw but no reason to process it if config generated an error
             _process_env();
         } catch(const CLI::FileError &fe2) {
```
```suggestion
        } catch(const CLI::FileError &) {
```

include/CLI/App.hpp
henryiii: Casting up then throwing was a mistake anyway. We shouldn't lose the subclass of one is used.

PR #679, opened 2021-12-02
Add WASI build support

thewtex:
Addresses:
```
  wasm-ld: error:
  CMakeFiles/WebAssemblyInterfaceHeaderTest1.dir/test/WebAssemblyInterfaceHeaderTest1.cxx.o:
  undefined symbol:
  std::__2::__fs::filesystem::__status(std::__2::__fs::filesystem::path
  const&, std::__2::error_code*)

  /usr/wasi-sdk-14.0/share/wasi-sysroot/include/c++/v1/filesystem:256:3:
  error: "The Filesystem library is not supported by this configuration
  of libc++"
  # error "The Filesystem library is not supported by this configuration
  of libc++"
```


henryiii (approved):
Thanks!

Issue #682, opened 2021-12-09
Allow to set a config root folder with set_config


Firefly35:
Sometimes an application will host several configurations (profiles) in its configuration folder (for instance in .app/profiles/...).
However, set_config doesn't provide a way to add a root_config folder where a profile would be searched for if it isn't found from its filename.

phlptp:
What do you think of the solution in #698?

Firefly35:
It's beyond my expectations, moreover it allows to specify several paths to
look for. It is a great solution to the issue I set.
Thanks for this proposition.


Le lun. 31 janv. 2022 à 23:45, Philip Top ***@***.***> a
écrit :

PR #683, opened 2021-12-09
feat[#682]: add the ability to set a config root folder. 

Firefly35:
Using set_config will search for the config filename from the current run directory of the application.
When the file is not found in this situation, it is searched for from the config_root_folder when it is set.


Firefly35:
Reopening the pull request from another branch to follow CLI11 guidelines

PR #684, opened 2021-12-09
Add the ability to set a config root folder in set_config

Firefly35:
Using set_config will search for the config filename from the current run directory of the application.
When the file is not found in this situation, it is searched for from the config_root_folder when it is set.
This pull request fixes the #682 issue.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -931,6 +931,14 @@ class App {
         return config_ptr_;
     }
 
+    Option *set_config(std::string option_name = "",
+                       std::string default_filename = "",
+                       const std::string &help_message = "Read an ini file",
+                       bool config_required = false) {
+
+        return set_config(option_name, default_filename, help_message, "", config_required);
+    }
+
     /// Removes an option from the App. Takes an option pointer. Returns true if found and removed.
     bool remove_option(Option *opt) {
         // Make sure no links exist
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -901,10 +901,10 @@ class App {
 #endif
 
     /// Set a configuration ini file option, or clear it if no name passed
-    Option *set_config(std::string option_name = "",
-                       std::string default_filename = "",
-                       const std::string &help_message = "Read an ini file",
-                       const std::string &default_config_root_folder = "",
+    Option *set_config(std::string option_name,
+                       std::string default_filename,
+                       const std::string &help_message,
+                       const std::string &default_config_root_folder,
                        bool config_required = false) {
 
         // Remove existing config if present
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -257,6 +257,10 @@ class App {
     /// Pointer to the config option
     Option *config_ptr_{nullptr};
 
+    /// base folder to search config for
+    /// when it is empty, search only from config_ptr_->default_str();
+    std::string default_config_folder_ = "";
+
     /// This is the formatter for help printing. Default provided. INHERITABLE (same pointer)
     std::shared_ptr<Config> config_formatter_{new ConfigTOML()};
 
@@ -900,6 +904,7 @@ class App {
     Option *set_config(std::string option_name = "",
                        std::string default_filename = "",
                        const std::string &help_message = "Read an ini file",
+                       const std::string & default_config_root_folder = "",
                        bool config_required = false) {
 
         // Remove existing config if present
@@ -907,6 +912,9 @@ class App {
             remove_option(config_ptr_);
             config_ptr_ = nullptr;  // need to remove the config_ptr completely
         }
+        
+        // Set the default config folder
+        default_config_folder_ = default_config_root_folder;
 
         // Only add config if option passed
         if(!option_name.empty()) {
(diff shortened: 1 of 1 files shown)


phlptp:
I am not sure I like this change.  Adding paths here seems problematic to me.  One thing I could see being useful is making support in the config_file for a ';' separated list.  So it could check down the list until it found one and use that one.  That would be a little more general support and still give the flexibility that this PR is proposing without adding a new function signature.

Firefly35:
In fact this functionality is a rather good solution for tools that host
several configurations in a $home/.app/configs/...
Calling the tool anywhere from the command line and switching config can -
with the PR - just use the config "name" instead of providing full paths to
every config file.
It's a common pattern for command line tools to have a centralized
configurations/profiles folder hosted in a "root" config folder.
For the moment I maintain my own fork, but I'd prefer if in a way or
another the functionality makes its way in CLI11.
Why do you find adding a "standardized path" searched for when the config
file was not found from its filepath is problematic ?

Le lun. 13 déc. 2021 à 18:14, Philip Top ***@***.***> a
écrit :

phlptp:
Might be just that the applications I work on are all cross platform so there is no standardized locations and having an absolute path in a binary that could be installed in countless different ways leads to all sorts of problems.   Or even a relative one would have to defined in a per operating system basis due to different path formats in use on different operating systems.  

I suppose that might be an applications problem vs a CLI11 problem though.

Firefly35:
Indeed. And the root path built is per os (in fact the same on unixes
including Linux and macos, but the home dir is built differently for
windows).
You can refer to the remaken application to see how I set it (see
CmdOptions.cpp) here
https://github.com/b-com-software-basis/remaken/tree/release/1.9.x

Le lun. 13 déc. 2021 à 22:27, Philip Top ***@***.***> a
écrit :

include/CLI/App.hpp
```diff
@@ -2018,8 +2034,19 @@ class App {
                 return;
             }
             for(auto rit = config_files.rbegin(); rit != config_files.rend(); ++rit) {
-                const auto &config_file = *rit;
+                std::string config_file = *rit;
                 auto path_result = detail::check_path(config_file.c_str());
+                if(path_result == detail::path_type::nonexistent && !default_config_folder_.empty()) {
+                    // Try to find the config file from its name in default config folder
+                    std::string config_file_path = default_config_folder_;
+                    if(default_config_folder_.back() != '/' && default_config_folder_.back() != '\\') {
```
phlptp, line 2042: would this check be better in the `set_config` function to only do it once rather than every element of a loop?

phlptp:
Wow, and I thought we had a lot of options in our applications. 

 I am fine with it,  let @henryiii decide to merge it if he thinks it appropriate.  
I know we went through a lot of revisions into CLI11 to not have to use a custom version and get the needed changes merged in, so thanks,

henryiii:
`set_config`'s arguments change in a non-backward compatible way and there are no tests. But IMO this is not needed at all; can't you just add a transform to the option that injects the default folder? I'd much rather build things out of existing components vs. add new API, especially new API that is a positional argument (so unreadable). A nice test + docs for how to do it would be great, though.

Firefly35:
Did you look deeply to the PR ? set_config has always the previous API signature, and there is an overload with the additionnal default_config_root_folder argument.
Moreover, I don't see how I can put a transform option as the config_root_folder is needed to figure out the root folder to search the config for while using set_config, AFAIK this happens before any option is parsed.

henryiii:
> there is an overload

I missed the overload. But that makes this potentially confusing, as well, since not only are there no names on the arguments in C++, the order is not stable either.

> Moreover, I don't see how I can put a transform option as the config_root_folder is needed to figure out the root folder to search the config for while using set_config, AFAIK this happens before any option is parsed.

Maybe we should try to fix that instead? At least I think we should carefully consider trying that path first. The config option is already special, so I think we could build on that.

phlptp:
I kind of like Henry's idea of enabling validators and transforms on the config option.  That is probably an oversight on our part.  Then I think a built in transform that checks paths and if necessary adds a default path might be quite useful in other contexts as well.

Firefly35:
I let you choose the best way to provide this feature in future CLI11
releases.
For now, we live with the workaround I quickly implemented :).
Concerning overloading in C++, there is IMHO no issue about arguments as
the method signature differs on the number and kind of arguments.
I don't think there can be any ambiguity between f(str,str,cstr,bool) and
f(str,str,cstr,cstr,bool) ?
Argument order would change if the calling convention was changed at
compile time, but as CLI11 is header only the calling convention would be
the same throughout the translation unit,
so I don't get where the argument order would be an issue ? (provided the
calling convention is the same in all libs/applications, but with different
calling conventions, not only CLI11 would be impacted, every argument
forwarding between the app and library would be impacted, no ?)


Le ven. 28 janv. 2022 à 17:47, Philip Top ***@***.***> a
écrit :

phlptp:
Just an update,  transforms work fine on the config file option.  They are executed before an options value gets used, which in the case of config is just before the file is processed, so a transform on the option from set_config works as expected and can be used.

henryiii:
My worry was in teaching it / reading it; default arguments are easier, though they still require a user look them up, but different signatures require each signature to be documented, for readers to be familiar with both, and count parameters, etc. Not that it can't be handled by a compiler.

phlptp:
I am going to close this as a similar capability was merged in #698.

PR #686, opened 2021-12-10
Hide silent subcommands from the help.

Anteru:
Right now there doesn't seem to be a way to hide a subcommand from the help list. `silent` is documented as removing it from `get_subcommands`, but because help calls that with a custom filter function (via `Formatter::make_subcommands`), every subcommand is included there. This PR hides `silent` subcommands from the help output. I hope that matches the original intent, if not, I guess a new option "hidden from help" or so could be added.

### DIFF
--- a/book/chapters/subcommands.md
+++ b/book/chapters/subcommands.md
@@ -126,3 +126,5 @@ This would allow calling help such as:
 ./app help
 ./app help sub1
 ```
+
+Silent subcommands also don't appear in the help text.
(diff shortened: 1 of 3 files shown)


conr2d:
You can hide a subcommand from the help list by setting its group to empty string("").

``` c++
auto cmd = app.add_subcommand("debug", "debugging mode");
cmd->group("");
```

henryiii:
The docs need updating, but yes, empty string rather than a special "silent" group was what was chosen.

PR #689, opened 2021-12-12
Update transform validators documentation

### LINKED ISSUE #688: Transform validators documentation needs an update.
In #687 I asked for help because I did not understand what the base function of validators should return, or how to mutate an option's value.

JonathanGzzBen:
Close #688 
Adds an example on how to use transform validator objects and clarifies
the usage of the return type of the base function.


phlptp:
I was looking at this a little more.   I think part of the confusion is that the transform method that takes a lambda  uses a different signature from the `std::function` used in the validator object.    So you can't just take that same lambda function and stick it in a Validator object even though they are compatible.  You get very different results.  

Also if the string return from the validator is not empty, then an exception gets generated from that.  So it is permissible but not necessary to generate an exception for failed validation inside the function.  You can just return an error message as the result.  

Not sure if there is a better way to do that or not?   The best I can think of right now is to add an error flag to the validator, so if  validator failed then that flag must be set and then the input is always const and the output is either a message or the transformed value depending on whether the flag was set

henryiii:
Returning the error message is preferred, since CLI11 might be able to run without exceptions enabled one day. That's why that signature is used.

I think this is a clear upgrade, but we could follow up with information about the validator std::function?

phlptp (approved):

Issue #696, opened 2022-01-26
Symlink tests/mesonTest/subprojects/CLI11 causes directory cycle


boris-kolpackov:
The `tests/mesonTest/subprojects/CLI11` symlink points to its parent directory (`../../..` to be precise) which causes a directory cycle that in turn can cause traversal issues for various tools. In case of `build2` you get a warning, which is annoying:

```
warning: directory cycle caused by symlink build/.bpkg/tmp/78c1e67e6423/upstream/tests/mesonTest/subprojects/CLI11/
  info: symlink target build/.bpkg/tmp/78c1e67e6423/upstream/
```

Also note that directory symlinks on Windows are tricky (if at all available): https://build2.org/article/symlinks.xhtml#windows

jspelletier:
Hello,
This symlink is causing job failures in gitlab when this repo is used as submodules and the git strategy is clone. It seems this can also happen in fetch strategy if runner was previously in clone mode.

I don't know why you need this file but couldn't you instead create it at runtime when running tests? Please consider removing this symlink!

The symlink causes errors such as the following:
Remove-Item : There is a mismatch between the tag specified in the request and the tag present in the reparse point
```
At C:\Users\xxxxxxx\AppData\Local\Temp\build_script368801463\script.ps1:224 char:3
+   Remove-Item -Force -Recurse "C:\gitlab-runner\builds\CsRJCNVP\0\yyyy...
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Remove-Item], Win32Exception
    + FullyQualifiedErrorId : System.ComponentModel.Win32Exception,Microsoft.PowerShell.Commands.RemoveItemCommand
```
 gitlab is unable to delete the submodule folder and this causes errors later on when trying to execute git clone operations
 
 Thanks

PR #697, opened 2022-01-27
remove the symbolic link in the meson test

phlptp:
Remove the symbolic link for Issue #696

Add a symlink generation command in the meson CI test


henryiii:
Calling it `.gitkeep` is somewhat standard. And you weren't able to mkdir it, I guess?

phlptp:
trying the mkdir now

phlptp:
looks like the mkdir from the command line works as well, so that is probably the best solution here

jspelletier:
Note: the symlink should be deleted after the test or the problem with gitlab will persist if someone has a gitlab CI executing that test as a submodule. in my case since I am not executing the tests, I think that would work

henryiii:
> gitlab CI executing that test as a submodule

Why would gitlab CI pick up a line from Azure pipelines?

azure-pipelines.yml
```diff
@@ -82,15 +82,22 @@ jobs:
...
+      displayName: generate CLI11 symlink
     - script: meson build
       displayName: Run meson to generate build
       workingDirectory: tests/mesonTest
     - script: ninja -C tests/mesonTest/build
       displayName: Build with Ninja
     - script: ./tests/mesonTest/build/main --help
       displayName: Run help
+    - script: rm -rf tests/mesonTest/subprojects
+      displayName: remove symlinks
```
henryiii: Cleaner without these lines, IMO.
```suggestion
```

henryiii (approved):

PR #698, opened 2022-01-28
Add a transform/validator that checks for files on a default path.

### LINKED ISSUE #682: Allow to set a config root folder with set_config

### LINKED ISSUE #694: Unexpected behavior with add_flag and std::optional<bool>
Using a recent (2021-12-27) 9ec0ba7 build, under msvc14.1x64 (2017)

I want to represent a `std::optional<bool>` flag so I tried

```c++
std::optional<bool> opt_bool;
CLI::App app = ...
assert(opt_bool.has_value() == false); 
CLI::Option* opt = app.add_flag("--flag,!--no-flag", opt_bool, "My flag desc.");
assert(opt_bool.has_value() == false); // incorrectly (to me) is true
```

but `opt_bool.has_value()` is always `true` after `add_flag(..)`

I can work around this via

1) using `opt->count() > 0` to reset the optional or
2) 
```c++
auto callback =  [&opt_bool](int value){
	if (value == 1) opt_bool = true;
	else if (value == 2) opt_bool = false;
};
CLI::Option* opt = app.add_flag("--flag{1},--no-flag{2}", callback, "My flag desc.");
```

but was wondering if the first approach was meant to work as I had hoped?

Here's a simple project that demonstrates it: https://godbolt.org/z/cxzaY3o3s

phlptp:
This is an alternative to #684 and closes #682. It add a transform/validator that checks for a file and on a default path.  
This if used with a config ptr can be used to add a default path, or any number of default paths for that matter.  

I think it has the same capabilities as #684 without adding any additional API calls and adding a new validator which may have other uses as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -428,7 +428,7 @@ CLI11 has several Validators built-in that perform some common checks
 * `CLI::ExistingDirectory`: Requires that the directory exists.
 * `CLI::ExistingPath`: Requires that the path (file or directory) exists.
 * `CLI::NonexistentPath`: Requires that the path does not exist.
-* `CLI::FileOnDefaultPath`: Best used as a transform, Will check that a file exists either directly or in a default path and update the path appropriately.  See [Transforming Validators](#transforming-validators) for more details
+* `CLI::FileOnDefaultPath`: 🚧 Best used as a transform, Will check that a file exists either directly or in a default path and update the path appropriately.  See [Transforming Validators](#transforming-validators) for more details
 * `CLI::Range(min,max)`: Requires that the option be between min and max (make sure to use floating point if needed). Min defaults to 0.
 * `CLI::Bounded(min,max)`: Modify the input such that it is always between min and max (make sure to use floating point if needed). Min defaults to 0.  Will produce an error if conversion is not possible.
 * `CLI::PositiveNumber`: Requires the number be greater than 0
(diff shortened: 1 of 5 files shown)


phlptp:
This will need a couple more test cases and some docs before merging but wanted to see what @henryiii  and Firefly35 thought of this alternative.

henryiii:
This looks good, works within the existing infrastructure, and can be expanded to look in several locations (like `~/.config` and [`./.config`](https:dot-config.github.io), etc). I think a bit of docs showing how to use it would be useful.

README.md
```diff
@@ -795,6 +798,13 @@ app.set_config("--config")->expected(1, X);
 
 Where X is some positive number and will allow up to `X` configuration files to be specified by separate `--config` arguments.  Value strings with quote characters in it will be printed with a single quote. All other arguments will use double quote.  Empty strings will use a double quoted argument. Numerical or boolean values are not quoted.
 
+NOTE:  Transforms and checks can be used with the option pointer returned from set_config like any other option to validate the input if needed.  It can also be used with the built in transform `CLI::FileOnDefaultPath` to look in a default path as well as the current one.  For example
+
+```cpp
+app.set_config("--config")->transform(CLI::FileOnDefaultPath("/to/default/path/"));
+```
+See [Transforming Validators](#transforming-validators) for additional details on this validator.
```
```suggestion

See [Transforming Validators](#transforming-validators) for additional details on this validator.
```

book/chapters/config.md
```diff
@@ -4,6 +4,22 @@
...
+if it is desired that config files be searched for a in a default path the `CLI::FileOnDefaultPath` transform can be used.
+
+```cpp
+app.set_config("--config")->transform(CLI::FileOnDefaultPath("/default_path/"));
+```
+
+This will allow specified files to either exist as given or on a specified default path.
+
+```cpp
+app.set_config("--config")->transform(CLI::FileOnDefaultPath("/default_path/"))->transform(CLI::FileOnDefaultPath("/default_path2/",false));
```
henryiii: Would it make sense to add a transform(A, B, ...) override so that this could be written:

```cpp
app.set_config("--config")->transform(
    CLI::FileOnDefaultPath("/default_path2/",false),
    CLI::FileOnDefaultPath("/default_path/"));
```

Regardless, I'd use a little formatting to make this more readable:

```suggestion
app.set_config("--config")
    ->transform(CLI::FileOnDefaultPath("/default_path/"))
    ->transform(CLI::FileOnDefaultPath("/default_path2/",false));
```

README.md
henryiii: A few construction emojis are needed, too.

henryiii (approved):

book/chapters/config.md
phlptp: I don't think added an overload would be that useful.  Something like this is already possible
```cpp
app.set_config("--config")->transform(
    CLI::FileOnDefaultPath("/default_path2/") |
    CLI::FileOnDefaultPath("/default_path/"));
```

using the existing `|` operation,  Which I think is a bit more obvious that it is an or operation vs an and operation,  So I don't think an overload would add much.

henryiii: Nice, I like the or. :)

sg-james:
Apologies for bumping, but I believe this PR might have accidentally closed issue #694 by accident and meant to close issue #682.

henryiii:
Corrected, thanks!

PR #702, opened 2022-02-09
ci: move MSVC 2015 to GHA

henryiii:


.github/workflows/tests.yml
```diff
@@ -63,6 +69,26 @@ jobs:
...
+        toolset: "14.0"
+
+    - name: Configure
+      run: pipx run cmake -S . -B build
+
+    - name: Build
+      run: pipx run cmake --build build -j2
+
+    - name: Test
+      run: pipx run cmake --build -t test
```
henryiii:
```suggestion
      run: pipx run cmake --build build -t test
```

henryiii:
Looks like it is using Visual Studio 16 2019 anyway.

.github/workflows/tests.yml
```diff
@@ -63,6 +69,26 @@ jobs:
       run: meson compile -C build-meson
 
 
+  msvc-2015:
+    name: MSVC 2015 test
+    runs-on: windows-2019
```
henryiii, line 74:
```suggestion
    runs-on: windows-2017
```

.github/workflows/tests.yml
```diff
@@ -63,6 +69,26 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: ilammy/msvc-dev-cmd@v1
+      with:
+        toolset: "14.0"
```
henryiii:
```suggestion
        toolset: "15.0"
```

.github/workflows/tests.yml
```diff
@@ -63,6 +69,26 @@ jobs:
...
 
+  msvc-2015:
+    name: MSVC 2015 test
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: ilammy/msvc-dev-cmd@v1
+      with:
+        toolset: "15.0"
```
henryiii:
```suggestion
        toolset: "15"
```

.github/workflows/tests.yml
```diff
@@ -63,6 +69,22 @@ jobs:
...
 
 
+  msvc-2015:
+    name: MSVC 2017 test
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Configure
+      run: pipx run cmake -S . -B build -G "Visual Studio 15 2017" -A x64
```
henryiii:
```suggestion
      run: pipx run cmake -S . -B build -G "Visual Studio 14 2015" -A x64
```

henryiii:
It does look like there's no way to get MSVC < 2019 from GH after they remove `windows-2016`, which used to be the place to get both. We can hold off here for another month, and after 2016 is gone, see what the landscape is. I'd rather hope MS is successful and ppl no longer care about <2019.

.github/workflows/tests.yml
```diff
@@ -63,6 +69,22 @@ jobs:
...
 
 
+  msvc-2015:
+    name: MSVC 2015 test
+    runs-on: windows-2019
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Configure
+      run: cmake -S . -B build -G "Visual Studio 14 2015" -A x64
```
henryiii:
```suggestion
      run: cmake -S . -B build -G "Visual Studio 14 2015"
```

PR #706, opened 2022-02-23
Fix typo in Macros.hpp

RangeMachine:


yuriy-odonnell-epic:
This issue was found thanks to `error C4668: '__MSVC_LANG' is not defined as a preprocessor macro, replacing with '0' for '#if/#elif'` compiler diagnostic. Our codebase happens to enable the relevant warning.

henryiii:
I'd like to know how to enable that warning. :)

yuriy-odonnell-epic:
`#pragma warning (error : 4668)` this could be one way ([compiler explorer](https://godbolt.org/z/618Y3154n)).

PR #708, opened 2022-02-23
chore: enable all MSVC level 4 warnings

henryiii:
The warning in https://github.com/CLIUtils/CLI11/pull/706 is level 4, but off by default. Let's see what else might show up.


henryiii:
This was an interesting experiment, but not something we can enable completely.

PR #709, opened 2022-02-24
Counting flags

phlptp:
This should address #694.  

There was a simple way to address #694 by adding increasingly complex enable_if discriminations with the type, but that didn't really address what I consider to be the heart of the issue, which is that flags are treated different depending on the type and that can result in unexpected behavior as was observed in this issue.  

This PR starts to address it, by adding `MultiOptionPolicy::Sum`    and a set of flag option defaults that is type dependent.  

The Sum option allows any type to act as a sum operator like was done before but doing it this way allows it to act on regular options, floating point values, or strings.

By adding the capability in that way we can simplify the conditionals on the add_flag and reduce one of the overloads.  (it may be possible to get rid of more now but I don't want to try that here).


phlptp:
this needs to be rebased when #702 is merged

README.md
```diff
@@ -363,7 +363,7 @@ Before parsing, you can set the following options:
...
-* `->multi_option_policy(CLI::MultiOptionPolicy::Throw)`: Set the multi-option policy. Shortcuts available: `->take_last()`, `->take_first()`,`->take_all()`, and `->join()`. This will only affect options expecting 1 argument or bool flags (which do not inherit their default but always start with a specific policy). `->join(delim)` can also be used to join with a specific delimiter. This equivalent to calling `->delimiter(delim)` and `->join()`
+* `->multi_option_policy(CLI::MultiOptionPolicy::Throw)`: Set the multi-option policy. Shortcuts available: `->take_last()`, `->take_first()`,`->take_all()`, and `->join()`. This will only affect options expecting 1 argument or bool flags (which do not inherit their default but always start with a specific policy). `->join(delim)` can also be used to join with a specific delimiter. This equivalent to calling `->delimiter(delim)` and `->join()`.  Valid values are `CLI::MultiOptionPolicy::Throw`, `CLI::MultiOptionPolicy::Throw`,`CLI::MultiOptionPolicy::TakeLast`,`CLI::MultiOptionPolicy::TakeFirst`,`CLI::MultiOptionPolicy::Join`,`CLI::MultiOptionPolicy::TakeAll`, `CLI::MultiOptionPolicy::Sum`🚧.
```
```suggestion
* `->multi_option_policy(CLI::MultiOptionPolicy::Throw)`: Set the multi-option policy. Shortcuts available: `->take_last()`, `->take_first()`,`->take_all()`, and `->join()`. This will only affect options expecting 1 argument or bool flags (which do not inherit their default but always start with a specific policy). `->join(delim)` can also be used to join with a specific delimiter. This equivalent to calling `->delimiter(delim)` and `->join()`.  Valid values are `CLI::MultiOptionPolicy::Throw`, `CLI::MultiOptionPolicy::Throw`, `CLI::MultiOptionPolicy::TakeLast`, `CLI::MultiOptionPolicy::TakeFirst`, `CLI::MultiOptionPolicy::Join`, `CLI::MultiOptionPolicy::TakeAll`, and `CLI::MultiOptionPolicy::Sum` 🚧.
```

book/chapters/options.md
```diff
@@ -171,7 +171,7 @@ When you call `add_option`, you get a pointer to the added option. You can use t
 | `->allow_extra_args()` | Allow extra argument values to be included when an option is passed. Enabled by default for vector options. |
 | `->disable_flag_override()` | specify that flag options cannot be overridden on the command line use `=<newval>` |
 | `->delimiter('<CH>')` | specify a character that can be used to separate elements in a command line argument, default is <none>, common values are ',', and ';' |
-| `->multi_option_policy( CLI::MultiOptionPolicy::Throw)` | Sets the policy for handling multiple arguments if the option was received on the command line several times. `Throw`ing an error is the default, but `TakeLast`, `TakeFirst`, `TakeAll`, and `Join` are also available. See the next four lines for shortcuts to set this more easily. |
+| `->multi_option_policy( CLI::MultiOptionPolicy::Throw)` | Sets the policy for handling multiple arguments if the option was received on the command line several times. `Throw`ing an error is the default, but `TakeLast`, `TakeFirst`, `TakeAll`, `Join`, `Sum` are also available. See the next four lines for shortcuts to set this more easily. |
```
henryiii:
```suggestion
| `->multi_option_policy( CLI::MultiOptionPolicy::Throw)` | Sets the policy for handling multiple arguments if the option was received on the command line several times. `Throw`ing an error is the default, but `TakeLast`, `TakeFirst`, `TakeAll`, `Join`, and `Sum` are also available. See the next four lines for shortcuts to set this more easily. |
```

henryiii (approved):

PR #726, opened 2022-05-02
Starting an experiment Yaml support

dfleury2:
Hi,
I started to write an experimental Yaml support.
I have added a ConfigYaml class with almost all tests from ConfigFileTest (TOML or INI).
Most are ok but, due to the lack of my understanding of how the result of parse from file should be, I failed for secton tests and some parse count tests. The ones still in comment in the ConfigYamlFileTest.cpp

I have started the optional part for conan, and CMakeLists.txt. not parfect yet. Need help on this part (the create package test does not work). I have used yaml-cpp and it is enable by default.

If anyone with a better insight of CLI11 internal can help. It's welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ a.out*
 /*.kdev4
 /html/*
 !/meson.build
+.idea
 
 /node_modules/*
 /package.json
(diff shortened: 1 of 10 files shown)


dfleury2:
Back on this subject.
I will try another way (external only) on propose it as a contrib file.

PR #727, opened 2022-05-06
Include <functional> in FormatterFwd

shameekganguly:
Without this header, the project might not compile is some environments.

Signed-off-by: Shameek Ganguly <shameekarcanesphinx@gmail.com>

### DIFF
--- a/include/CLI/FormatterFwd.hpp
+++ b/include/CLI/FormatterFwd.hpp
@@ -7,6 +7,7 @@
 #pragma once
 
 // [CLI11:public_includes:set]
+#include <functional>
 #include <map>
 #include <string>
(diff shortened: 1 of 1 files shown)


phlptp:
Do you know any particular environments?  Something we can add to the CI builds?

henryiii:
Is this on the new GCC? I saw several headers now often have to listed separately in https://gcc.gnu.org/gcc-12/porting_to.html, but not `<functional>`. It would be nice to have some way to test this so more things don't show up too.

jzakrzewski:
Maybe some tool, like "include-what-you-use" would suffice?

phlptp (approved):

PR #734, opened 2022-05-24
fix: rename variable to avoid clash with optarg

PeteAudinate:
https://linux.die.net/man/3/optarg

I realise it's unusual to include both CLI11 and getopt.h, but it can happen in large diverse codebases.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -2844,9 +2844,9 @@ class App {
                     break;
                 }
                 if(validate_optional_arguments_) {
-                    std::string optarg = args.back();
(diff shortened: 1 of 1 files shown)


henryiii:
I think the new name is fine, but a local `optarg` shouldn't conflict with the extern? Generally only macros (well, one being a macro) collide.

PeteAudinate:
Thanks for the quick reply.

You're right, it's not really a conflict but MSVC complains:
`include\CLI\App.hpp(2847,1): warning C4459: declaration of 'optarg' hides global declaration`

PR #741, opened 2022-06-07
wrap min and max 

phlptp:
in parenthesis for Visual studio 2017 and earlier compatibility.
This fixes some issues from a max/min macro on some earlier Visual studio builds

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,6 +71,14 @@ repos:
         entry: PyBind|Numpy|Cmake|CCache|PyTest|Github
         exclude: .pre-commit-config.yaml
 
+  - repo: local
+    hooks:
+      - id: avoid-msvc-macro
+        name: Avoid MSVC <=2017 min/max macro (use extra parens)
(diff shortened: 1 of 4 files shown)


henryiii:
Any way to test/enforce this not being reintroduced? Maybe a pre-commit pygrep looking for `min()` or `max()`?

henryiii:
Maybe:

```yaml
     - id: avoid-msvc-macro
       name: Avoid MSVC <=2017 min/max macro (use extra parens)
       entry: \b(min|max)\(\)
       language: pygrep
       exclude: .pre-commit-config.yaml
```

phlptp:
I think the pre-commit is working,  caught a few more in the tests though I don't think those would ever have been an issue.  

Also looks like we need to update the windows CI builds on azure as one of the images currently in use is being deprecated

henryiii:
We'll be out of the brownout later / tomorrow and then we can get #742 in while we still have MSVC 2015. After that, we'll probably have to drop MSVC 2015 support.

PR #742, opened 2022-06-07
chore: clang tidy updates

henryiii:
```bash
docker run --rm -v $PWD:/cli11 -it silkeh/clang:14 bash
cmake -Scli11 -Bbuild -DCMAKE_CXX_CLANG_TIDY="$(which clang-tidy);--use-color" -DCMAKE_CXX_STANDARD=17
cmake --build build -j 4
```

This removes the custom clang-tidy option for standard CMake idioms. Clang warnings adjusted, and new tidy checks added.

### DIFF
--- a/include/CLI/Config.hpp
+++ b/include/CLI/Config.hpp
@@ -82,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(std::isspace(sepChar) == 0) {
+            if(std::isspace<char>(sepChar, std::local()) == 0) {
                 joined.push_back(' ');
             }
         }
--- a/include/CLI/Config.hpp
+++ b/include/CLI/Config.hpp
@@ -8,6 +8,7 @@
 
 // [CLI11:public_includes:set]
 #include <algorithm>
+#include <cctype>
 #include <fstream>
 #include <iostream>
 #include <string>
@@ -81,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(isspace(sepChar) == 0) {
+            if(std::isspace(sepChar) == 0) {
                 joined.push_back(' ');
             }
         }
--- a/include/CLI/Config.hpp
+++ b/include/CLI/Config.hpp
@@ -25,7 +25,7 @@ namespace detail {
 
 inline std::string convert_arg_for_ini(const std::string &arg, char stringQuote = '"', char characterQuote = '\'') {
     if(arg.empty()) {
-        return {2, stringQuote};
+        return std::string(2, stringQuote);
     }
     // some specifically supported strings
     if(arg == "true" || arg == "false" || arg == "nan" || arg == "inf") {
--- a/include/CLI/Config.hpp
+++ b/include/CLI/Config.hpp
@@ -25,7 +25,7 @@ namespace detail {
 
 inline std::string convert_arg_for_ini(const std::string &arg, char stringQuote = '"', char characterQuote = '\'') {
     if(arg.empty()) {
-        return std::string(2, stringQuote);
+        return {2, stringQuote};
     }
     // some specifically supported strings
     if(arg == "true" || arg == "false" || arg == "nan" || arg == "inf") {
@@ -33,7 +33,7 @@ inline std::string convert_arg_for_ini(const std::string &arg, char stringQuote 
     }
     // floating point conversion can convert some hex codes, but don't try that here
     if(arg.compare(0, 2, "0x") != 0 && arg.compare(0, 2, "0X") != 0) {
-        double val;
+        double val = 0.0;
         if(detail::lexical_cast(arg, val)) {
             return arg;
         }
@@ -62,9 +62,8 @@ inline std::string convert_arg_for_ini(const std::string &arg, char stringQuote 
     }
     if(arg.find_first_of(stringQuote) == std::string::npos) {
         return std::string(1, stringQuote) + arg + stringQuote;
-    } else {
-        return characterQuote + arg + characterQuote;
     }
+    return characterQuote + arg + characterQuote;
 }
 
 /// Comma separated join, adds quotes if needed
@@ -230,7 +229,7 @@ inline std::vector<ConfigItem> ConfigBase::from_config(std::istream &input) cons
             std::string item = detail::trim_copy(line.substr(pos + 1));
             auto cloc = item.find(commentChar);
             if(cloc != std::string::npos) {
-                item.erase(cloc, std::string::npos);
+                item.erase(cloc, std::string::npos);  // NOLINT(readability-suspicious-call-argument)
                 detail::trim(item);
             }
             if(item.size() > 1 && item.front() == aStart) {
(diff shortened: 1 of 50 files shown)


include/CLI/Option.hpp
```diff
@@ -682,41 +682,41 @@ class Option : public OptionBase<Option> {
...
 
     /// The minimum number of arguments the option expects
-    int get_type_size_min() const { return type_size_min_; }
+    CLI11_NODISCARD int get_type_size_min() const { return type_size_min_; }
     /// The maximum number of arguments the option expects
-    int get_type_size_max() const { return type_size_max_; }
+    CLI11_NODISCARD int get_type_size_max() const { return type_size_max_; }
 
     /// Return the inject_separator flag
-    int get_inject_separator() const { return inject_separator_; }
```
phlptp, line 693: should the type of this be bool,  I noticed a few places this was used makes me think the error is in the return type here rather than in all the other spots?

henryiii: It does seem like it should be bool.

include/CLI/StringTools.hpp
```diff
@@ -347,7 +347,7 @@ inline std::vector<std::string> split_up(std::string str, char delimiter = '\0')
 
     const std::string delims("\'\"`");
     auto find_ws = [delimiter](char ch) {
-        return (delimiter == '\0') ? (std::isspace<char>(ch, std::locale()) != 0) : (ch == delimiter);
+        return (delimiter == '\0') ? (static_cast<int>(std::isspace<char>(ch, std::locale())) != 0) : (ch == delimiter);
```
phlptp:
```suggestion
        return (delimiter == '\0') ? std::isspace<char>(ch, std::locale()) : (ch == delimiter);
```
I guess std::isspace<TYPE> returns a bool directly,  no reason to convert it to int only to convert to a bool again.

include/CLI/Validators.hpp
```diff
@@ -212,11 +215,10 @@ class Validator {
             std::string s2 = f2(input);
             if(!s1.empty() && !s2.empty())
                 return std::string("(") + s1 + ") AND (" + s2 + ")";
-            else
-                return s1 + s2;
+            return s1 + s2;
         };
 
-        newval.active_ = (active_ & other.active_);
+        newval.active_ = ((static_cast<int>(active_) & static_cast<int>(other.active_)) != 0);
```
phlptp: this change seems odd?  maybe should be active_ && other.active_   instead of the int casts

include/CLI/Validators.hpp
```diff
@@ -240,7 +242,7 @@ class Validator {
 
             return std::string("(") + s1 + ") OR (" + s2 + ")";
         };
-        newval.active_ = (active_ & other.active_);
+        newval.active_ = ((static_cast<int>(active_) & static_cast<int>(other.active_)) != 0);
```
phlptp: see above comment

tests/AppTest.cpp
```diff
@@ -788,7 +790,7 @@ TEST_CASE_METHOD(TApp, "SumOpt", "[app]") {
 
 TEST_CASE_METHOD(TApp, "SumOptFloat", "[app]") {
 
-    double val;
+    double val = NAN;
```
phlptp, line 793: Should this use `std::nan("")` instead of the C Macro?

tests/OptionGroupTest.cpp
```diff
@@ -421,7 +421,7 @@ struct ManyGroups : public TApp {
         g3->add_option("--val3", val3);
     }
 
-    void remove_required() {
+    void remove_required() const {
```
phlptp: I suspect it doesn't matter but it is kind of misleading for this to be const.

tests/OptionTypeTest.cpp
```diff
@@ -44,13 +49,13 @@ TEST_CASE_METHOD(TApp, "doubleFunction", "[optiontype]") {
...
-    CHECK(300.356 == res);
+    CHECK(300.356_a == res);
     // get the original value as entered as an integer
-    CHECK(-354.356f == app["--val"]->as<float>());
+    CHECK(-354.356_a == app["--val"]->as<float>());
 }
 
 TEST_CASE_METHOD(TApp, "doubleFunctionFail", "[optiontype]") {
-    double res;
+    double res = NAN;
```
phlptp, line 58: same comment on NAN as earlier

include/CLI/StringTools.hpp
henryiii: The regular one is int, but the locale one is bool. 👍

tests/AppTest.cpp
henryiii: Seems to be in C++: https://en.cppreference.com/w/cpp/numeric/math/NAN

tests/OptionGroupTest.cpp
henryiii: Okay, will do that.

phlptp (approved):
Looks like some errors with std::isspace on windows systems but otherwise looks good to me.

include/CLI/Config.hpp
```diff
@@ -82,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(isspace(sepChar) == 0) {
+            if(std::isspace(sepChar) == 0) {
```
henryiii:
```suggestion
            if(std::isspace<char>(sepChar, std::local()) == 0) {
```

include/CLI/Config.hpp
```diff
@@ -82,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(isspace(sepChar) == 0) {
+            if(std::isspace<char>(sepChar, std::local()) == 0) {
```
henryiii:
```suggestion
            if(std::isspace<char>(sepChar, std::locale()) == 0) {
```

include/CLI/Config.hpp
```diff
@@ -82,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(isspace(sepChar) == 0) {
+            if(std::isspace<char>(sepChar, std::locale()) == 0) {
```
henryiii:
```suggestion
            if(std::isspace<char>(sepChar, std::locale()) {
```

include/CLI/Config.hpp
```diff
@@ -82,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(isspace(sepChar) == 0) {
+            if(std::isspace<char>(sepChar, std::locale()) {
```
henryiii:
```suggestion
            if(std::isspace<char>(sepChar, std::locale())) {
```

include/CLI/Config.hpp
```diff
@@ -82,7 +82,7 @@ inline std::string ini_join(const std::vector<std::string> &args,
     for(const auto &arg : args) {
         if(start++ > 0) {
             joined.push_back(sepChar);
-            if(isspace(sepChar) == 0) {
+            if(std::isspace<char>(sepChar, std::locale())) {
```
henryiii:
```suggestion
            if(!std::isspace<char>(sepChar, std::locale())) {
```

PR #748, opened 2022-06-27
ci: include windows 2022

henryiii:
This seems to be broken.


phlptp:
I think I fixed the VS2022 issue,  seemed to be something with C++20 changes to allowed constructors for std::array that messed up some of the template type detections.

phlptp:
@henryiii  not sure what is going on with the pre-commit error.

henryiii:
Curious. Pre-commit updated a few days ago, but no node changes afaict, that’s all I see that’s changed. Can check soonish, currently at SciPy.

Issue #338, opened 2019-10-27
Plan for precompiled mode


henryiii:
Current idea:

* All non-templated function/method bodies move to new files, such as `option_fwd.inl`.
* `inline` gets replaced by `CLI11_INLINE_DEF`
* If `CLI11_COMPILE` is not defined, `CLI11_INLINE_DEF` is `inline` and the `.inl` files get included at the end of `option.hpp`
* If `CLI11_COMPILE` is defined, `CLI11_INLINE_DEF` is nothing. 0 or 1 selects whether this is the compiled file or a user.
* The generation scripts need to be adjusted to handle this correctly; the .icc part will be wrapped in `#if`s.

The CMake files will automatically compile each header with an icc file, and link to that. A user not using CMake can add a file like this:

```cpp
#def CLI11_COMPILE 1
#include "CLI11.hpp"
```

Then use CLI11 like this:

```cpp
#def CLI11_COMPILE 0
#include "CLI11.hpp"
```

And get the benefits of compilation. If nothing is done, nothing changes and everything is header only.

---

```cpp
// option.hpp

#include "option_fwd.hpp"
#ifndef CLI11_COMPILE || CLI11_COMPILE==1
#include "option.inl"
#endif
```

```cpp
// option_fwd.hpp

// Minimal required includes 
// [cli11:preinc:includes]
#include <fwd>
// [cli11:preinc:end]

// All files need this (has inline, etc)
#include "macros.hpp"

// Includes for other files in CLI11 (fwd only)
#include "app_fwd.hpp"

namespace CLI {
// [cli11:app_fwd:code]

class Option {
   // Definitions and templated functions only
};
CLI11_INLINE void f();

// [cli11:app_fwd:end]
}  // namespace CLI
```

```cpp
// option.inl

// Currently not planning to include fwd here, since it is an inline file (depending on IDEs, possibly)

// Compile only includes
// [cli11:inc:includes]
#include <algorithm>
// [cli11:inc:end]

namespace CLI {
// [cli11:option_inl:code]
Option::Option() {} // Definitions
CLI11_INLINE void f() {} 
// [cli11:option_inl:end]
}  // namespace CLI
```

```cpp
// option.cpp
#define CLI11_COMPILE 1 // Might as well force it here)
#include <option.hpp>
```

phlptp:
We would need to make sure that the library worked well as a subproject in cmake as part of a bigger project, that requires a bit more care and flexibility when it has the option of a non-header only build.  

[...]

PR #759, opened 2022-08-08
[precompiled] Initial proof-of-concept with App.cpp

dherrera-fb:
This is a proof-of-concept PR to show a potential implementation of https://github.com/CLIUtils/CLI11/issues/338

- Add C11_COMPILE cmake option that creates a static lib instead of header-only
- Add C11_INLINE macro that depends on C11_COMPILE
- Split App.hpp into App.hpp and impl/App_inl.hpp
- Add App.cpp that compiles App_inl.hpp into an object file
- CMake modifications to handle impl headers differently for sinlge-header, headers-only, and compiled versions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,7 @@ endif()
 
 option(CLI11_WARNINGS_AS_ERRORS "Turn all warnings into errors (for CI)")
 option(CLI11_SINGLE_FILE "Generate a single header file")
+option(CLI11_PRECOMPILED "Generate a precompiled static library instead of a header-only")
 cmake_dependent_option(CLI11_SANITIZERS "Download the sanitizers CMake config" OFF
                        "NOT CMAKE_VERSION VERSION_LESS 3.11" OFF)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -160,13 +166,35 @@ if(NOT CMAKE_VERSION VERSION_LESS 3.13)
...
+  list(INSERT MAYBE_CONFIGURE_DEPENDS 0 CONFIGURE_DEPENDS)
+endif()
+
+file(GLOB CLI11_headers ${MAYBE_CONFIGURE_DEPENDS} "${PROJECT_SOURCE_DIR}/include/CLI/*.hpp")
+file(GLOB CLI11_impl_headers ${MAYBE_CONFIGURE_DEPENDS}
+     "${PROJECT_SOURCE_DIR}/include/CLI/impl/*.hpp")
+
+if(CLI11_PRECOMPILED)
+  # Create static lib
+  file(GLOB CLI11_precompile_sources "${PROJECT_SOURCE_DIR}/src/*.cpp")
```
jzakrzewski, line 181: Do you really think that globbing source files is a good idea?

dherrera-fb: I have no opinion on globbing. Master branch is using glob so this PR is following the status-quo. I'm happy to update it if glob is removed from the main branch.

jzakrzewski: Yeah - it's already bad in master but this makes it even worse when the library will get build with outdated source list during development it's time wasted for chasing non-existent problems.
I know some people do not want to list all the files in the build scripts but I'd urge the maintainers to consider that.

henryiii: `CONFIGURE_DEPENDS` makes it quite a bit better, and we add that if CMake supports it.

henryiii: (though I think it's missing here, actually)

PR #762, opened 2022-08-15
feat: precompiled all

### LINKED ISSUE #338: Plan for precompiled mode

dherrera-fb:
This is the full implementation precompiled option. It splits all header files into x.hpp and x_inl.hpp, where the first has only declarations and trivial inlines and the latter has the implementations. 

If the cmake variable CLI11_PRECOMPILED is false:
- All implementations are tagged as inline
- One can still choose multi-header or single-header.
Else
- Implementations are not tagges as inline
- There is a single Precomplie.cpp file that compiles all implementations
- The cmake target becomes a static library
- The inl files are not included in the headers

This considerably improves compile-time but may hurt binary size.

Closes #338


CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ endif()
 
 option(CLI11_WARNINGS_AS_ERRORS "Turn all warnings into errors (for CI)")
 option(CLI11_SINGLE_FILE "Generate a single header file")
+option(CLI11_PRECOMPILED "Generate a precompiled static library instead of a header-only")
```
henryiii: I think I'd recommend setting the default to OFF at first. That way it could be opt-in for a while, and then maybe the default could swap in 3.0.

henryiii (commented):
It's hard to review the diff, since everything moved.  The procedure outlined sounds correct, though. I think we should block out merging any other PRs until this goes in. Does this include any/all recent changes?

CC @phlptp

PS: Don't worry about coverage. We lost 100% at some point when we lost coverage reporting for a while. We'll bring it back up eventually.

CMakeLists.txt
dherrera-fb: Agree. Pushed a commit to make it off by default

henryiii:
Ahh, `readability-avoid-const-params-in-decls` is something I hadn't thought of. `const T x` and `T x` are the same thing if T is a pass-by-copy type, but adding the const helps the author of the function not change `x`. So those can be removed from the declarations but not the definitions. (Well, technically, the definitions too since they only help authors working on the definition).

Perfectly fine to just disable the check for now, and re-add it in a separate PR. Diff should stay small for this sort of moving code!

dherrera-fb:
> It's hard to review the diff, since everything moved. The procedure outlined sounds correct, though. I think we should block out merging any other PRs until this goes in. Does this include any/all recent changes?
> 
> CC @phlptp
> 
> PS: Don't worry about coverage. We lost 100% at some point when we lost coverage reporting for a while. We'll bring it back up eventually.

I think this contains all recent changes. It was based off main. Last commit is c57000e582fae45b01cd8074c79c81897379cbe5 from Jul 5th. I think that is still main as of now.

Review is indeed hard for such massive changes. I tried to split it into commits to tackle each file separately. But I don't think there is much more I can do.

dherrera-fb:
CI doesn't yet have checks for building precompiled mode. Maybe we should add that before merging?

henryiii:
@phlptp Okay with this? I'm okay to merge then see if anything needs polishing. If we are lucky, maybe we'll have ppl on main that will help catch anything our tests do not.

phlptp:
It looks ok to me,  It is a big change and I suspect will need a bit of polish but that is going to be difficult to check at present, so I think  I am good with a merge then we can figure out what cleanup is necessary after a bit more testing.

phlptp (approved):

PR #768, opened 2022-08-24
CMake Update

phlptp:
Change the CMake to use subdirectory and move the library generation …and warnings to separate files.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -110,7 +110,7 @@ jobs:
           containerImage: gcc:8
           cli11.std: 17
         gcc4.8:
-          containerImage: gcc:4.8
+          containerImage: helics/buildenv:gcc4-8-builder
           cli11.std: 11
           cli11.options:
         clang3.4:
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,5 +1,5 @@
 function(add_cli_exe T)
-  add_executable(${T} ${ARGN} ${CLI11_headers})
+  add_executable(${T} ${ARGN})
   target_link_libraries(${T} PUBLIC CLI11)
   set_property(TARGET ${T} PROPERTY FOLDER "Examples")
   if(CLI11_FORCE_LIBCXX)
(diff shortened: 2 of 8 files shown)


phlptp:
pre-commit.ci autofix

azure-pipelines.yml
```diff
@@ -109,10 +109,6 @@ jobs:
         gcc8:
           containerImage: gcc:8
           cli11.std: 17
-        gcc4.8:
-          containerImage: gcc:4.8
```
henryiii, line 113: Is there a reason this can't just change to `helics/buildenv:gcc4-8-builder`? Does `.ci/azure-cmake.yml` break?

phlptp: My guess would be no, but CMake is already installed so just seemed a waste of time, I will put it back to make sure.

examples/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 function(add_cli_exe T)
-  add_executable(${T} ${ARGN} ${CLI11_headers})
+  add_executable(${T} ${ARGN})
```
henryiii, line 2: By removing the headers, doesn't this mean you can't access these in an IDE anymore? (Though if you are using the build-ahead mode, you can - so maybe that's why?)

henryiii (approved):

examples/CMakeLists.txt
phlptp: In an IDE, the files are now added to the CLI11 Project, which is new.  They are added as build in precompile mode or with newer CMakes they are added as private sources in the interface library.

azure-pipelines.yml
phlptp: I put it back in the Docker section,  installing CMake takes about 30 seconds on the CI test, so we can reduce the build time slightly by leaving it in its own section, or just leave the extra CMake install in place.

henryiii: We've got enough jobs where I'm not too worried about it. I'd rather like to move to GHA eventually & then we can probably use the CMake action. Keeping this simple makes it easier to move. ;)

PR #774, opened 2022-09-14
Add vcpkg installation instructions

Adela0814:
`cli11` is available as a port in vcpkg, a C++ library manager that simplifies installation for cli11 and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build cli11, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/cli11/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,10 @@ them are unknown yet.
 </br>
 
 ## Installing cli11 using vcpkg
-You can download and install cli11 using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+You can download and install cli11 using the
+[vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
 ```bash
 git clone https://github.com/Microsoft/vcpkg.git
 cd vcpkg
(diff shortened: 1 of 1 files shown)


jzakrzewski:
Sorry for the negativity, but I'm not convinced that the project should document something it does not control. Such things tend to get outdated pretty easily.

henryiii:
I'd be okay to help make PRs to keep this up to date if it becomes a problem. I'm a little more worried about making the README grow too much. I think I'd rather add this to the gitbook, and then link to it from the readme. I also need to move the gitbook over to a non-dead renderer. Let's keep it open but on hold until we work on the docs a bit.

siilky:
in the meanwhile, could we update vcpkg release version? It is still 2.2 there, which is a year old

phlptp:
changes are included in #901

Issue #773, opened 2022-09-13
config_to_str outputs wrong values for !--no-flag on v2.2.0


Krpa:
Having only `!--no-flag` specified when adding an option or having it as the first name will result in wrong (reversed) config file outputs. 
Steps to reproduce:

1. Run with `--no-flag`. This results with setting the variable `flag=false` (correct), but it outputs `no-flag=false`.
2. Save the output (`no-flag=false`) to `config.toml`
3. Run with` --config "path/to/config.toml"`. This results with `flag=true` and the output is `no-flag=true`.
```
#include <CLI11/CLI11.hpp>

#include <iostream>

int main()
{
  CLI::App app;
  bool flag;
  app.add_flag("!--no-flag", flag);
  app.set_config("--config,--ini");
  try {
    app.parse(argc, argv);
  } catch (const CLI::Error& e) {
    return app.exit(e);
  }

  std::cout << app.config_to_str(false, false) << std::endl;
  return 0;
}
```

phlptp:
Well that seems like a bug,  I don't think that scenario was tested.  I will look into it.

PR #775, opened 2022-09-14
Negated flag config files

phlptp:
Fix Issue #773
add some coverage tests

### DIFF
--- a/include/CLI/impl/Config_inl.hpp
+++ b/include/CLI/impl/Config_inl.hpp
@@ -344,8 +344,8 @@ ConfigBase::to_config(const App *app, bool default_also, bool write_description,
                 }
 
                 if(!value.empty()) {
-                    if (!opt->get_fnames().empty()) {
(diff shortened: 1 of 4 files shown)


henryiii:
> add some coverage tests

Yeah, we really need to get back to 100% - coverage much less useful when it's not 100%, because it can produce false positives.

henryiii (approved):

phlptp:
I think the switch to a compiled library(possibility) changed the code coverage.  Looks like there are sections of missing coverage that are pretty old.  Anyway getting back to 100% is going to be a long process.

henryiii:
I think it mostly happened about a year ago when we lost Travis, new PRs stopped being checked for coverage so it started slipping. When I recently renabled it (before the compiled change), that's when I discovered the coverage had slipped. There might have been even more coverage losses due to the change, as well. :/

No rush, probably something we can work on once in a while.

PR #776, opened 2022-09-14
chore: update changelog

henryiii:
Updating changelog a bit. Are we getting close to 2.3?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,36 @@
 # Changelog
 
+## Version 2.3: Precompilation Support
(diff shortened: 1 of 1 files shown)


henryiii:
@phlptp anything left before a 2.3?

phlptp:
I don't have anything else I am working on and I think getting the precompiled stuff into a version would be good.

phlptp:
A few tweaks with the new and under construction markers in the readme need to be changed before a release, but otherwise I think I am fine with the release.

PR #785, opened 2022-10-06
fix: static analysis problems

### LINKED ISSUE #784: Static analysis issues
We are using cli11 v2.2.0 in one of the projects I am working on. Running Klocwork for static analysis of the project, we observe some issues reported by the tool. Most of the issues are false positives, but a few are easy to fix.:

1)
>\# 438: 'num' is used uninitialized in this function. Also there is one similar error on line 997.
>CLI11-2.2.0/include/CLI/Validators.hpp:979 | operator()()
>Code: UNINIT.STACK.MUST | Severity: Critical (1) | State: Existing | Status: Analyze | Taxonomy: C and C++ | Owner: >unowned

Fix (added curly brackets):
```
...
func_ = [mapping, opts](std::string &input) -> std::string {
Number num{};
...
```

2)
>\# 268: Class 'Formatter' defines a copy constructor, but no assignment operator
>CLI11-2.2.0/include/CLI/FormatterFwd.hpp:116 | FormatterFwd.hpp()
>Code: CWARN.COPY.NOASSIGN | Severity: Review (4) | State: Existing | Status: Analyze | Taxonomy: C and C++ | Owner: >unowned

Fix:
```
FormatterBase& operator=(const FormatterBase&) = default;
```

3)
>\# 280: Class 'FormatterBase' defines a copy constructor, but no assignment operator
>CLI11-2.2.0/include/CLI/FormatterFwd.hpp:39 | FormatterFwd.hpp()
>Code: CWARN.COPY.NOASSIGN | Severity: Review (4) | State: Existing | Status: Analyze | Taxonomy: C and C++ | Owner: >unowned

Fix:
```
Formatter& operator=(const Formatter&) = default;
```

Another problem concerns the `stat()` function, which is recognized by the tool as a "time-of-check-time-of-use" vulnerability.:

>\# 410: function 'stat' operates on file names and is vulnerable to race conditions. Files can be manipulated by attackers >between creation and access time.
>CLI11-2.2.0/include/CLI/Validators.hpp:335 |check_path()
>Code: SV.TOCTOU.FILE_ACCESS | Severity: Review (4) | State: Existing | Status: Analyze | Taxonomy: C and C++ | Owner: >unowned

cetius:
fixes #784 - warnings reported by Klocwork in CLI11 v2.2.0

### DIFF
--- a/include/CLI/FormatterFwd.hpp
+++ b/include/CLI/FormatterFwd.hpp
@@ -121,6 +121,7 @@ class Formatter : public FormatterBase {
     Formatter(const Formatter &) = default;
     Formatter(Formatter &&) = default;
     Formatter &operator=(const Formatter &) = default;
+    Formatter &operator=(Formatter &&) = default;
 
     /// @name Overridables
     ///@{
--- a/include/CLI/FormatterFwd.hpp
+++ b/include/CLI/FormatterFwd.hpp
@@ -57,6 +57,7 @@ class FormatterBase {
     FormatterBase() = default;
     FormatterBase(const FormatterBase &) = default;
     FormatterBase(FormatterBase &&) = default;
+    FormatterBase& operator=(const FormatterBase&) = default;
 
     /// Adding a destructor in this form to work around bug in GCC 4.7
     virtual ~FormatterBase() noexcept {}  // NOLINT(modernize-use-equals-default)
@@ -118,6 +119,7 @@ class Formatter : public FormatterBase {
     Formatter() = default;
     Formatter(const Formatter &) = default;
     Formatter(Formatter &&) = default;
+    Formatter& operator=(const Formatter&) = default;
 
     /// @name Overridables
     ///@{
--- a/include/CLI/FormatterFwd.hpp
+++ b/include/CLI/FormatterFwd.hpp
@@ -57,7 +57,7 @@ class FormatterBase {
     FormatterBase() = default;
     FormatterBase(const FormatterBase &) = default;
     FormatterBase(FormatterBase &&) = default;
-    FormatterBase& operator=(const FormatterBase&) = default;
+    FormatterBase &operator=(const FormatterBase &) = default;
 
     /// Adding a destructor in this form to work around bug in GCC 4.7
     virtual ~FormatterBase() noexcept {}  // NOLINT(modernize-use-equals-default)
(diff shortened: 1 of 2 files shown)


include/CLI/FormatterFwd.hpp
```diff
@@ -57,6 +57,7 @@ class FormatterBase {
     FormatterBase() = default;
     FormatterBase(const FormatterBase &) = default;
     FormatterBase(FormatterBase &&) = default;
+    FormatterBase &operator=(const FormatterBase &) = default;
```
henryiii, line 60: What about the move assignment operator?

include/CLI/Validators.hpp
```diff
@@ -750,7 +750,7 @@ class AsNumberWithUnit : public Validator {
 
         // transform function
         func_ = [mapping, opts](std::string &input) -> std::string {
-            Number num;
+            Number num{};
```
henryiii, line 753: Not sure this is necessary, but it's okay.

include/CLI/FormatterFwd.hpp
cetius: Maybe it's a good idea, but I only focused on what the tool is complaining about.

include/CLI/Validators.hpp
cetius: I understand that this is just a cosmetic change. I just wanted to hold the tool back from moaning.

henryiii: If it's just a cosmetic change, I'd have no problem with it. I think it might zero initialize some types, which is technically wasteful, the will only be written to. But in the scope of parsing CLI code, it's totally acceptable.

include/CLI/FormatterFwd.hpp
henryiii: The rule of 5 states you should have none or all of these. Guessing the tool hasn't been updated for moves, perhaps? (It seems to be applying the rule of 3).

henryiii:
```suggestion
    FormatterBase &operator=(const FormatterBase &) = default;
    FormatterBase &operator=(FormatterBase &&) = default;
```

include/CLI/FormatterFwd.hpp
```diff
@@ -118,6 +119,7 @@ class Formatter : public FormatterBase {
     Formatter() = default;
     Formatter(const Formatter &) = default;
     Formatter(Formatter &&) = default;
+    Formatter &operator=(const Formatter &) = default;
```
henryiii, line 123:
```suggestion
    Formatter &operator=(const Formatter &) = default;
    Formatter &operator=(Formatter &&) = default;
```

include/CLI/Validators.hpp
jzakrzewski: Yep. That was my take also.

cetius: I completely understand your point of view

PR #786, opened 2022-10-07
feat: added usage message replacement feature

polistern:
## Usage message replacement

Now you able to set custom usage message or pass callback function that will generate it for you.  
Example:

```c++
CLI::App app{"app v1"};
app.usage("use: [-h|--help] | --option UINT");
```

Output will be:

```text
app v1
use: [-h|--help] | --option UINT

Options:
  -h,--help                             Show this message
```
To void usage message replacement in subcommand and be able to get generated usage message just set with empty string:

```c++
auto sub = app.add_subcommand("sub");
sub->usage("");
```
Help message for `app -h`:

```text
app v1
use: [-h|--help] | --option UINT

Options:
  -h,--help                             Show this message

Subcommands:
sub

```

Help message for `app sub -h`:

```text
Usage: app sub [OPTIONS]

Options:
  -h,--help                             Show this message
```

## Footer newline behavior

Moved to https://github.com/CLIUtils/CLI11/pull/796

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -147,6 +147,12 @@ class App {
     /// @name Help
     ///@{
 
+    /// Usage to put after program/subcommand description in the help output INHERITABLE
+    std::string usage_{};
+
+    /// This is a function that generates a usage to put after program/subcommand description in help output
+    std::function<std::string()> usage_callback_{};
+
     /// Footer to put after all options in the help output INHERITABLE
     std::string footer_{};
 
@@ -942,6 +948,16 @@ class App {
     /// @name Help
     ///@{
 
+    /// Set usage.
+    App *usage(std::string usage_string) {
+        usage_ = std::move(usage_string);
+        return this;
+    }
+    /// Set usage.
+    App *usage(std::function<std::string()> usage_function) {
+        usage_callback_ = std::move(usage_function);
+        return this;
+    }
     /// Set footer.
     App *footer(std::string footer_string) {
         footer_ = std::move(footer_string);
@@ -1049,6 +1065,11 @@ class App {
 
     /// Get the group of this subcommand
     CLI11_NODISCARD const std::string &get_group() const { return group_; }
+
+    /// Generate and return the usage.
+    CLI11_NODISCARD std::string get_usage() const {
+        return (usage_callback_) ? usage_callback_() + '\n' + usage_ : usage_;
+    }
 
     /// Generate and return the footer.
     CLI11_NODISCARD std::string get_footer() const {
--- a/include/CLI/impl/Formatter_inl.hpp
+++ b/include/CLI/impl/Formatter_inl.hpp
@@ -137,7 +137,7 @@ CLI11_INLINE std::string Formatter::make_footer(const App *app) const {
     if(footer.empty()) {
         return std::string{};
     }
-    return footer + "\n";
+    return "\n" + footer + "\n";
 }
 
 CLI11_INLINE std::string Formatter::make_help(const App *app, std::string name, AppFormatMode mode) const {
@@ -159,7 +159,7 @@ CLI11_INLINE std::string Formatter::make_help(const App *app, std::string name, 
     out << make_positionals(app);
     out << make_groups(app, mode);
     out << make_subcommands(app, mode);
-    out << '\n' << make_footer(app);
+    out << make_footer(app);
 
     return out.str();
 }
--- a/include/CLI/impl/Formatter_inl.hpp
+++ b/include/CLI/impl/Formatter_inl.hpp
@@ -91,6 +91,11 @@ CLI11_INLINE std::string Formatter::make_description(const App *app) const {
 }
 
 CLI11_INLINE std::string Formatter::make_usage(const App *app, std::string name) const {
+    std::string usage = app->get_usage();
+    if(!usage.empty()) {
+        return usage + "\n";
+    }
+
     std::stringstream out;
 
     out << get_label("Usage") << ":" << (name.empty() ? "" : " ") << name;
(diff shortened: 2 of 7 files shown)


include/CLI/App.hpp
```diff
@@ -147,6 +147,12 @@ class App {
     /// @name Help
     ///@{
 
+    /// Usage to put after program/subcommand description in the help output INHERITABLE
+    std::string usage_{};
```
henryiii, line 151: I find it confusing that there are two different things here, both set from the same function, that will both potentially show up. I think it would be better if there was just a `usage_callback_`, and `usage(std::string usage_string)` would just call usage again withset `[](){return usage_string;}`. Simpler internals, and I think a lot more predictable usage.

polistern: I just followed `footer` implementation as example when implementing similar behavior for `usage`.
So, if you think it's better to use only `callback` for `usage`, then I'll make the appropriate changes.

Could you clarify, should I leave only `callback` as you described?
Should appropriate changes be made for `footer`?

phlptp: The footer has two use cases.  One is to supply additional information via the callback.  I use this to trigger help in cascading apps.  The callback is not inherited.  The other case is to supply branding or copyright information which does need to be inherited(or at least can be) so it is printed on all subcommands, and both are used when generating the help.  

I am not sure the same situation applies to the usage, also not clear that usage should be inherited like the footer.  So it might make sense to not exactly follow the pattern for the footer.

henryiii: See more discussion in, too: https://github.com/CLIUtils/CLI11/pull/309#issuecomment-522364430.

include/CLI/impl/Formatter_inl.hpp
```diff
@@ -137,7 +142,7 @@ CLI11_INLINE std::string Formatter::make_footer(const App *app) const {
     if(footer.empty()) {
         return std::string{};
     }
-    return footer + "\n";
+    return "\n" + footer + "\n";
```
henryiii: Do you think you could pop out this part of the change into a quick PR? This can go into a patch release if you do. The usage one needs to target the next minor release instead.

include/CLI/impl/Formatter_inl.hpp
```diff
@@ -159,7 +164,7 @@ CLI11_INLINE std::string Formatter::make_help(const App *app, std::string name,
     out << make_positionals(app);
     out << make_groups(app, mode);
     out << make_subcommands(app, mode);
-    out << '\n' << make_footer(app);
+    out << make_footer(app);
```
henryiii: And here.

tests/FormatterTest.cpp
```diff
@@ -63,7 +63,7 @@ TEST_CASE("Formatter: OptCustomize", "[formatter]") {
                   "Usage: [OPTIONS]\n\n"
                   "Options:\n"
                   "  -h,--help              Print this help message and exit\n"
-                  "  --opt INT (MUST HAVE)  Something\n\n");
```
henryiii: And the changes to this file.

tests/HelpTest.cpp
```diff
@@ -813,7 +850,7 @@ TEST_CASE_METHOD(CapturedHelp, "CallForAllHelpOutput", "[help]") {
                        "  One description\n\n"
                        "two\n"
                        "  Options:\n"
-                       "    --three                     \n\n\n");
```
henryiii: and this one.

include/CLI/App.hpp
polistern: @phlptp, thanks for the clarification, now I get it.
Yes, I'll fix that behavior.

include/CLI/impl/Formatter_inl.hpp
polistern: I think I can make all fixes over the weekend.
Will it be quickly enough? :)
Or changes in `usage` not getting through anyway?

include/CLI/App.hpp
polistern: @henryiii, @phlptp,  little clarification to see if I understood correctly.

You suggest using only callback, something like this:

```c++
    /// Set usage.
    App *usage(std::string usage_string) {
        usage_callback_ = std::move([usage_string](){return usage_string;});
        return this;
    }
    /// Set usage.
    App *usage(std::function<std::string()> usage_function) {
        usage_callback_ = std::move(usage_function);
        return this;
    }
    /// Generate and return the usage.
    CLI11_NODISCARD std::string get_usage() const {
        return (usage_callback_) ? usage_callback_() : "";
    }
```

Without using separate string variable for plain text.
So, in this case, `usage` setters (with a string parameter and with a callback function parameter) will be mutually exclusive and non-inherited.

If so, I can make changes within a few hours.

include/CLI/impl/Formatter_inl.hpp
henryiii: I'd like to make a quick bug fix release with no new features, but the spacing on the footer is close enough to a bug I think it could be included. The usage is a new features, and needs to sit in main a little bit, be added to the readme with a new feature emoji, etc.

include/CLI/App.hpp
henryiii: Yes, I think that's the idea.

include/CLI/impl/Formatter_inl.hpp
henryiii: So the footer change can go into 2.3.1, while the usage will go into 2.4.0.

polistern: Well, this will be a lesson for me to cut such things in the future :)
In that case, I close this PR and create two separate ones?

henryiii: You can just make the PR for the fix, this one can easily be rebased. I think you can just make a new branch and checkout the related changes from this one.

polistern: Done (https://github.com/CLIUtils/CLI11/pull/796)

henryiii: Thanks! Merged. It is usually best to try to keep PRs very focused. :)

henryiii:
Thanks! Moving on with 2.4.

PR #789, opened 2022-10-14
feat: dot notation

phlptp:
Add the ability to use subcommand options using . notation  
`--sub.option`  or nested subcommands using dot notation.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -235,16 +235,29 @@ jobs:
           cmake-version: "3.20"
         if: success() || failure()
 
-      - name: Check CMake 3.21 (full)
+      - name: Check CMake 3.21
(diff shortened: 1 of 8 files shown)


phlptp:
Still likely need to add more tests and some documentation updates

phlptp:
from discussion on [gitter](https://gitter.im/CLI11gitter/Lobby?at=633edd799a0ad66a76f788eb)

phlptp:
@henryiii  I think this is ready now

henryiii:
Is there a test for toml config with this? That was one of the reasons for it IIRC?

phlptp:
The toml processing of dot notation worked before and is tested in a few locations, so this just adds to the command line parsing and doesn't change anything in the file parsing.

PR #793, opened 2022-10-27
missing function definition

phlptp:
the get_option_group definition was missing from the splitting of the definitions. 

I tried to compile the single header into our code and discovered a missing function call.

### DIFF
--- a/include/CLI/impl/App_inl.hpp
+++ b/include/CLI/impl/App_inl.hpp
@@ -433,6 +433,15 @@ CLI11_NODISCARD CLI11_INLINE CLI::App_p App::get_subcommand_ptr(int index) const
     throw OptionNotFound(std::to_string(index));
 }
 
+CLI11_NODISCARD CLI11_INLINE CLI::App *App::get_option_group(std::string group_name) const {
+for(const App_p &app : subcommands_) {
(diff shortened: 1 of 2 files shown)


phlptp:
Need to merge #794 first, that should fix the tests

henryiii:
Ouch, is this a regression? Do we know why it was passing tests?

phlptp:
I think the recent pre-commit update changed something and introduced a cyclic change in some files that pre-commit couldn't resolve on its own.  And I didn't want to make a bunch of unrelated doc changes in this PR.  As to the original problem yes I think it was a regression but I am still trying to figure out why this wasn't triggering some issues with the single header tests or the other tests for that matter.   What I know is tried updating the single header CLI11 used in other applications and it generated a linking error which led to a missing definition.

henryiii:
It does complain that it's completely untested.

henryiii:
And I'm not seeing a test for it.

phlptp:
I will add some tests

henryiii:
Looks like we can do a .1 release after this, no new features have gone in yet AFAICT.

phlptp:
as long as we do it before merging #789 and #786, that should be fine.

PR #794, opened 2022-10-27
add a few more coverage tests

phlptp:
try to add a few more coverage tests and fix some markdown issues


henryiii (approved):

henryiii:
Thanks for working on this! #793 shows that we need it.

PR #808, opened 2022-11-30
fix: make_description min option output

SherlockInSpace:
Include spacing between the number of minimum options required and the rest of the description.


phlptp (approved):

phlptp:
Thanks

Issue #805, opened 2022-11-21
Warnings gcc11 C++17


ferdymercury:
When including CLI11 in my program (C++17 standard), I get the following warnings. Is there anything I am doing incorrectly, or are these known?
Thanks.

```
/opt/CLI11/include/CLI/App.hpp:1358: warning: redundant redeclaration of ‘std::string CLI::FailureMessage::simple(const CLI::App*, const CLI::Error&)’ in same scope [-Wredundant-decls]
In file included from bth.cpp:8:
/opt/CLI11/include/CLI/App.hpp:1358:26: warning: redundant redeclaration of ‘std::string CLI::FailureMessage::simple(const CLI::App*, const CLI::Error&)’ in same scope [-Wredundant-decls]
 1358 | CLI11_INLINE std::string simple(const App *app, const Error &e);
      |                          ^~~~~~
/opt/CLI11/include/CLI/App.hpp:52:13: note: previous declaration of ‘std::string CLI::FailureMessage::simple(const CLI::App*, const CLI::Error&)’
   52 | std::string simple(const App *app, const Error &e);
      |             ^~~~~~
/opt/CLI11/include/CLI/App.hpp:1361: warning: redundant redeclaration of ‘std::string CLI::FailureMessage::help(const CLI::App*, const CLI::Error&)’ in same scope [-Wredundant-decls]
/opt/CLI11/include/CLI/App.hpp:1361:26: warning: redundant redeclaration of ‘std::string CLI::FailureMessage::help(const CLI::App*, const CLI::Error&)’ in same scope [-Wredundant-decls]
 1361 | CLI11_INLINE std::string help(const App *app, const Error &e);
      |                          ^~~~
/opt/CLI11/include/CLI/App.hpp:53:13: note: previous declaration of ‘std::string CLI::FailureMessage::help(const CLI::App*, const CLI::Error&)’
   53 | std::string help(const App *app, const Error &e);
      |             ^~~~
/opt/CLI11/include/CLI/Validators.hpp:84: warning: noexcept-expression evaluates to ‘false’ because of a call to ‘CLI::Validator::<lambda()>’ [-Wnoexcept]
In file included from /usr/include/c++/11/bits/move.h:57,
                 from /usr/include/c++/11/bits/exception_ptr.h:43,
                 from /usr/include/c++/11/exception:153,
                 from /usr/include/c++/11/ios:39,
                 from /usr/include/c++/11/ostream:38,
[...]

PR #813, opened 2022-12-07
fix some GCC warnings and update codecov

phlptp:
Try to address some of the warnings that have been reported from GCC.


phlptp:
Fixes the redundant declaration warnings from #805

henryiii (approved):

PR #820, opened 2022-12-25
Using ADL everywhere for lexical_cast

captainurist:
Right now it's pretty much impossible to extend the library for custom template types because we can't have partial function template specializations in C++ (and thus cannot partially specialize `CLI::detail::lexical_cast`). Note that just throwing an overload into `CLI::detail` doesn't work because the overload should be visible at the call site, and to have that you'd have to first declare the overload, and only then include the CLI11 headers. Not very practical.

This PR fixes it by making sure all calls to `lexical_cast` are done via argument-dependent lookup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1435,11 +1435,10 @@ provide a custom `operator>>` with an `istream` (inside the CLI namespace is
 fine if you don't want to interfere with an existing `operator>>`).
 
 If you wanted to extend this to support a completely new type, use a lambda or
-add an overload of the `lexical_cast` function in the namespace of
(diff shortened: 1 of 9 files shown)


phlptp:
Looks like one minor issue with clang tidy, then could be merged.  
It looks good, and will hopefully simplify adding custom parsers.
Thank you!

phlptp (approved):

phlptp:
I don't think the CMAKE test issue is related to this PR.  I will look into it more.

PR #823, opened 2023-01-02
add new CMAKE version support and fix CI test failure

phlptp:
add support for CMake 3.23 and 3.24 and fix an issue with the CMake tests

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -236,16 +236,28 @@ jobs:
           cmake-version: "3.20"
         if: success() || failure()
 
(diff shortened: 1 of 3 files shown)


henryiii:
Any special reason this is max 3.24 and not 3.25?

phlptp:
none, other than my local testing environment hasn't moved to 3.25 yet, so didn't have a local means of testing yet.  We will probably update in next couple weeks.

henryiii:
Okay, that's fine, just wanted to make sure there wasn't something (known) that needed fixing.

PR #830, opened 2023-01-13
format the CLI11_PARSE formart

aengusjiang:
clean the code

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -35,13 +35,13 @@ namespace CLI {
 // [CLI11:app_hpp:verbatim]
 
 #ifndef CLI11_PARSE
-#define CLI11_PARSE(app, argc, argv)                        \
-    do                                                      \
-        try {                                               \
-            (app).parse((argc), (argv));                    \
-        } catch(const CLI::ParseError &e) {                 \
-            return (app).exit(e);                           \
-        }                                                   \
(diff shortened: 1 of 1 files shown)


phlptp:
This would be a pretty big breaking change.   
Going from a `;` optional to now required.  
and only accepting the 2 argument form of parse for the macro.  

Arguably it should have been wrapped in a do/while loop from the beginning and all the examples use a `;` but I don't know that all user code is currently, and I have used the single argument parse with the macro in the past.

aengusjiang:
This pr main consider that when we write a code such such as function call it will be like fun();
Bug the in macro CLI11_PARSE define with a ';',  it is strange the write a code like:CLI11_PARSE() without ';', 
it is pretty that CLI11_PARSE(); but it has double ';'. 

so this pr can make the code more standard. the change consider the compatiable it good. but I think remote the ';' in  the macro is more more better.

phlptp:
While I agree with you, I don't think we can make this change until we would move to a 3.0 release as it would be a breaking change.

henryiii:
The two argument change was due to the recent merge, I think. This didn't start by forcing the semicolon, so we shouldn't force it now, it would break a lot of existing code. I've also always thought that form was weird - and IIRC, it tripped up a compiler (nvcc, perhaps?) when I was developing it, so I left it off.

Issue #821, opened 2022-12-28
Getting map-like data from config


siilky:
I have a map-like structure in config

```
[map]
a = 1
b = 2
```

Is it possible to get its contents in some readable form?

I have added 
```
auto map = cfg.add_subcommand("map");
map->allow_config_extras();
```
but `remaining()` gives me only field names like `"a", "b"` without values. It appears also I cannot provide some 'global' subcommand callback for more options.

Interesting that at the same time when I provide `map.a=1 map.b=2` from command line, app's `remaining()` returns the full form like `"map.a=1"`

phlptp:
This is likely an inconsistency in the way config files extras are handled.  I don't know of a good workaround at present, but this is probably symptomatic of a bug/oversight that should be fixed.

PR #839, opened 2023-02-05
fix: parse config file `remaining`

phlptp:
Adds the output values from a config file to the remaining stream.  
This fixes issue #821 and allows the remaining in the case subcommands to be reparsed if the options are added or in a second CLI11 app.

### DIFF
--- a/book/chapters/config.md
+++ b/book/chapters/config.md
@@ -87,6 +87,8 @@ app.allow_config_extras(CLI::config_extras_mode::ignore_all);
 will completely ignore any mismatches, extras, or other issues with the config
 file
 
+Config file extras are stored in the remaining output as two components. The first is the name of the field including subcommands using dot notation the second (or more) are the argument fields.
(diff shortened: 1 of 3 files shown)


phlptp:
The remaining can be reparsed if it is in a subcommand.    Which addresses the issue in #821.   This works since the dot notation handling accommodates that.    The same thing is not true for unknown fields not in a subcommand.  They will show up in the remaining as "a", "1",  which is not going to be recognized as a valid option should it be reprocessed into another CLI app with a named option.     

This is going to be a pretty unusual situation where it is needed, so not entirely clear it needs to be addressed but it is an inconsistency.    The only option I can think of is prepending a "-" or "--" when  generating the remaining arguments if not in a subcommand for config file parsing.  But I can also think of situations where that is not a desired behavior either, so planning on leaving it alone at present.

henryiii (approved):

PR #846, opened 2023-02-11
CLI11 Fuzzing

phlptp:
This PR adds the start of a fuzzing system for CLI11.  It currently includes 1 issue fix found by fuzzing so I want to get it merged so we can continue to expand on it.  

At some point I think it would worthwhile to try to get into OSS-Fuzz for CLI11 since it lies at the front end of an increasingly large number of projects given by the # of stars.


phlptp:
May have opened a bit of black hole here,  this could take a while to resolve.

phlptp:
I think this is ready for initial review, squashed all the commits together to a single one.  The fuzz program is by no means complete but this is a start, and it already found a couple bugs.  I hope to keep expanding it but I think getting it into the main branch will be good, and a place to build on.  With the eventual aim to get it into OSS-Fuzz.

henryiii:
Okay. The copyright lines at the top are a bit inconsistent; some of them are the original header, some are LL, and some don't have one at all.

henryiii (approved):

phlptp:
I will clean up the copyright, next time I work on this, probably sometime this weekend.

Issue #856, opened 2023-03-09
Adding a space between option aliases in help text [FEATURE REQUEST]


LostInCompilation:
Currently the generated help text looks like this:
```
Options:
  -h,--help                             Print this help message and exit
  -v,--version                          Print App version
```

I suggest adding a space between the aliases for better readability:
```
Options:
  -h, --help                             Print this help message and exit
  -v, --version                          Print App version
```

I didn't find an option for that in the docs. Writing a custom formatter just for that seems a bit overkill.

LostInCompilation:
**EDIT: Below is not correct. I made a PR which indirectly solves this issue by completely reformatting the help output #866**

~~I just digged through the code. It seems like this isn't even possible with a custom formatter. The only way would be to directly edit `std::string Option::get_name(bool positional, bool all_options)` and change the delimiter for the join function: `detail::join(name_list, ", ");`~~

~~I can make a PR on this. Maybe I'll also look at #353.~~

Issue #353, opened 2019-11-28
Question: Options formatting


sergeyklay:
Is there any simple way to get formatted help output for options like this:
```
Options:
      --option1 TEXT Some description for option1
  -h, --help         Print this help message and quit
      --vernum       Print the version of the application as integer and quit
  -V, --version      Print application version information and quit
```

Instead of (current behavior):
```
Options:
  --option1 TEXT Some description for option1
  -h,--help      Print this help message and quit
  --vernum       Print the version of the application as integer and quit
  -V,--version   Print application version information and quit
```

Thank you

henryiii:
You can write any formatter you like, there's a system in place for that (not very well documented, though). However, such a change looks quite useful for the main formatter; I'd take a PR to fix that (or may eventually manage to write one myself).

PR #858, opened 2023-03-10
[DEPRECATED] Better Help-Text formatting - Align single and double dash flags and options

LostInCompilation:
# Deprecated
### ⚠️ Please ignore everything below. Only the new PR is relevant now.

----------------------------------------------------------------------------------------------------------------------
**EDIT: I found a way to implement this PR nicely into the default `Formatter`, instead of the ugly way like it is now. I'm currently working on that and some additional features for better help formatting, like paragraphs with correct line wrapping for long descriptions.
Please ignore the text below, do not merge this PR, and please leave it open for now until I sorted things out. I'll take care of everything.**

**EDIT 2: Still on it, almost done. Should be completed this weekend.**

This PR is mainly based on #353 and indirectly on #856. It adds a new and better help text formatting by aligning single and double dash flags and options nicely.

A bit about the background: ~~As far as I understood the source code (and wrote in #856) there is no good way to reliable format all combinations for flags and options using a custom formatter (or by modifying the default formatter). This PR therefore changes the `std::ostream& format_help(...)` function directly.~~ **_EDIT: This is wrong. I'm working on a PR which changes the default formatter and leaves the `format_help` function in it's original state._**
A goal was not needing to modify any function signatures or otherwise spread-out changes to the codebase. Because of that, the code for this PR seems a bit more verbose since, for example, we need to split up the parameter `std::string name` inside `format_help` after it got assembled while actually calling `format_help`. But as I said, I wanted to avoid changing function signatures or spread the changes across more files than necessary. I added a lot of comments to explain what is going on. The code also supports setting a custom column size. Single dash flags occupy 25% of the column width, double dash flags and options (like REQUIRED) 75%.

I thoroughly testes this code and so far it worked without a problem with all examples and custom tests made by me. ~~I'm currently testing with CLI11 tests.~~ **_EDIT: The failed tests are failing because they expect the old help text format. So all Tests seem to "pass"._** However there is no guarantee (yet) that it is fully bug free. The code is only manipulating singe and double dash flags and options (-f or --file). Subcommands and positionals are formatted like before, no changes are made to them.

Here's an example of the new help text formatting:
```
Description
Usage: myApp [OPTIONS]

Options:
  -h      --help                        Print this help message and exit
  -v      --version                     Display program version information and exit
  -p      --path STRING REQUIRED        Path to start in
  -a, -b  --multiple                    Multiple
          --only-long                   Long
          --hidden, --hide              Show or hide hidden files
  -r      REQUIRED                      Short req.
  -s                                    Short

This is a footer!
```

The alignment is like suggested in #353 and also looks more UNIX standard.

I think I'll make some improvements tomorrow or simplify some code, fix some spelling and code style issues. But there shouldn't be any major changes. I'll appreciate some feedback and am looking forward to possibly see the new formatting in CLI11. If this is completely the wrong way and such changes are only allowed using custom formatters, let me know.

### DIFF
--- a/include/CLI/impl/StringTools_inl.hpp
+++ b/include/CLI/impl/StringTools_inl.hpp
@@ -10,10 +10,10 @@
 #include <CLI/StringTools.hpp>
 
 // [CLI11:public_includes:set]
-#include <string>
-#include <vector>
 #include <algorithm>
 #include <cmath>
+#include <string>
+#include <vector>
 // [CLI11:public_includes:end]
 
 namespace CLI {
@@ -122,7 +122,7 @@ format_help(std::ostream &out, std::string name, const std::string &description,
         // Calculate setw sizes
         const int leftSideWidth = static_cast<int>(wid) / 4;  // 25% left for short names
         const int rightSideWidth = static_cast<int>(
-                                                    std::ceil(static_cast<float>(wid) / 4.0f * 3.0f));  // 75% right for long names and options, ceil result
+            std::ceil(static_cast<float>(wid) / 4.0f * 3.0f));  // 75% right for long names and options, ceil result
 
         //*****************************************************************************
         // Assemble short (single dash) names if any and print them on left side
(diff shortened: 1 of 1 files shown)


LostInCompilation:
This PR has been deprecated. A new and better PR has been made. See #866.

PR #861, opened 2023-03-12
Polish configuration file format

VolkerChristian:
- Polish empty lines and disable description printing for apps and subcommands with empty name, empty group and with no options and subcommands.
- Add commentDefaults() method to trigger to comment default value options in case default_also is true, so that the result set of an option is not pollute with default option values.
- In case dafault_also is true mark required but not yet configured options as "<REQUIRED>" and not yet configured default options as "" in a commented line in the config file.


phlptp:
Going to work on this next after #1073 is merged.

phlptp:
closing merged in #1075,  thanks!!

PR #862, opened 2023-03-12
Add a method to trigger reading of all config files

VolkerChristian:
This pull request did not change the behavior of application code if the newly introduced method
`App *all_config_files(bool all = true)` is not used.

- Add a method `App *all_config_files(bool all = true)` to trigger reading of all config files in reverse order so that the last read config file takes precedence.
- This is at least interesting when subcommand aliases are used and the same option names exist in different configuration files for these different subcommand names. This allows a combined option set to be constructed from different subcommand names and configuration files.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -288,6 +288,10 @@ class App {
     /// This is the formatter for help printing. Default provided. INHERITABLE (same pointer)
     std::shared_ptr<Config> config_formatter_{new ConfigTOML()};
 
+    /// Take all options from all config files into account. Results are in the order config files are parsed
+    /// Last parsed config file takes precedence
+    bool all_config_files_{false};
+
     ///@}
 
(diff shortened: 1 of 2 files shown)


VolkerChristian:
I made three pull requests today. Perhaps they will be accepted to be integrated into the main branch of CLI11.

I use a CLI11 patched like this heavily in [SNode.C](https://github.com/VolkerChristian/snode.c) to manage the whole configuration system. The changes have proven to be stable and functional.

Please contact me in case you have questions about that pull requests.

Kind regards
Volker Christian

phlptp:
Is the main purpose of this to allow reading of the config files in reverse order? If that is the case there probably is an argument to be made that that should be the proper way of doing it.     In which case would the new method be necessary?

phlptp:
Going to close this unless @VolkerChristian has some other comment

PR #864, opened 2023-03-13
Test update

phlptp:
update checkout to v3, update cmake tested version to 3.25, and add some additional CUDA tests

Update the CUDA 10 test, to CUDA 11 and added a CUDA 12 test.
Update the boost test to use the runner directly and install boost on ubuntu.  
upgrade all checkout actions to v3
add CMAKE 3.25 test and allow use in CLI11


phlptp:
this will supercede #863

henryiii (approved):

PR #872, opened 2023-04-10
Fix escaped arguments

### LINKED ISSUE #871: Allow resolution of escaped arguments.
Given a program `exe` with the following help:

```
Usage: exe [OPTIONS] flags

Positionals:
  flags TEXT_REQUIRED         Compiler flags.

Options:
  -O       output name
```

The following command line, as expected, fails:
`exe -O2`

The following commands line, which should work, fails:
`exe \\-O2`

The end user could remove the escapes themselves; however, it seems appropriate for CLI11 to handle this.

ScottBailey:
Fixes https://github.com/CLIUtils/CLI11/issues/871.

Added tests to exercise code and then added code to handle escaped characters.


ScottBailey:
There is a better way.

Issue #875, opened 2023-05-04
Unicode support: automatically widen on windows when option is assigned a std::filesystem::path


jmarrec:
I've seen #804, and it does widen in check_path for the CLI:ExistingXXXX validator. https://github.com/CLIUtils/CLI11/pull/804/files#diff-212772a3245e02e1fb8bf60ed69b5c89eb1b4f109e459b3294ff1edbd6c6aa27R131

Assuming this MCVE

```c++
#include <CLI/CLI.hpp>
#include <fmt/format.h>
#include <fmt/std.h>     // Formatting std::filesystem::path

#include <filesystem>

namespace fs = std::filesystem;

int main(int argc, char** argv) {

  CLI::App app("test path");
  fs::path dir;
  app.add_option("DIR", dir, "Directory")->check(CLI::ExistingDirectory);

  CLI11_PARSE(app);

  fmt::print("{} - {}\n", dir, fs::exists(dir));
}
```

In the resulting executable directory:

```shell
mkdir voilà
./test voilà
```

Ideally that would work on both unix and windows.

On windows, the CLI:ExistingDirectory is perfectly fine since it does widden when checking, but the resulting `fs::path dir` doesn't exist and it prints `"tÃ©rmica" - false`

jmarrec:
Note: the workaround is this

```c++
app.add_option_function<std::string>(
      "DIR", [&dir](const std::string& s) { dir = CLI::to_path(s); }, "Directory")
    ->check(CLI::ExistingDirectory);
```

phlptp:
looks like we might have to double check the direct support for std::filesystem paths in a flag.

jmarrec:
so, Template MetaProgramming (TMP) is not my strong suit, but I think I understand why this happens.

Turns out TypeTools does try to do it around here:

1.

https://github.com/CLIUtils/CLI11/blob/4a14b78ff44ee4e25105d0ec7d6fc1fbd7bbca5e/include/CLI/TypeTools.hpp#L1027-L1032

2.

 https://github.com/CLIUtils/CLI11/blob/4a14b78ff44ee4e25105d0ec7d6fc1fbd7bbca5e/include/CLI/TypeTools.hpp#L1043-L1050

The thing is, for std::filesystem::path, it's both string_assignable and wstring_assignable, on all platforms. But it resolves to object_category::string_assignable anyways.

https://gcc.godbolt.org/z/YWErffnjn

[...]

PR #876, opened 2023-05-05
prioritize wide strings over strings on MSVC

phlptp:
to handle `std::filesystem::path` better widestring operations should be preferred over regular strings.  

Fix Issue #875

### DIFF
--- a/tests/AppTest.cpp
+++ b/tests/AppTest.cpp
@@ -1724,7 +1724,7 @@ TEST_CASE_METHOD(TApp, "filesystemWideName", "[app]") {
 
     CHECK_THROWS_AS(app.parse(L"--file voil\u20ac.txt"), CLI::ValidationError);
 
-    bool ok = static_cast<bool>(std::ofstream(L"voil\u20ac.txt").put('a'));  // create file
+    bool ok = static_cast<bool>(std::ofstream(myfile).put('a'));  // create file
     CHECK(ok);
 
     // deactivate the check, so it should run now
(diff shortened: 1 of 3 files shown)


tests/AppTest.cpp
```diff
@@ -1715,6 +1715,30 @@ TEST_CASE_METHOD(TApp, "FileExists", "[app]") {
...
+
+    CHECK_NOTHROW(app.parse(L"--file voil\u20ac.txt"));
+
+    CHECK(fpath == myfile);
+
+    CHECK(std::filesystem::exists(fpath));
+    std::filesystem::remove(myfile);
+    CHECK(!std::filesystem::exists(fpath));
+}
+#endif
```
jmarrec: On Ubuntu 20.04, if you do run this test it fails

**That means necessarily passing `-DCMAKE_CXX_STANDARD:STRING=20` (or 17)**

```
filesystemWideName
CLI11/tests/AppTest.cpp:1719

CLI11/tests/AppTest.cpp:1719: FAILED:
  filesystem error: Cannot convert character sequence: Invalid or incomplete
  multibyte or wide character
```

phlptp: A couple of the CI builds are failing with the same error, I am going to take a closer look today.  Seems there is some ambiguity in the standards about how filesystem::path is treated with respect to strings and some differences between compilers.

phlptp:
I swapped the prioritize wstring to only be active on MSVC,  not sure if that should be all windows or just visual studio?

henryiii (approved):

phlptp:
@jmarrec  I am going to merge this and close #878,  feel free to resubmit if there are situations this doesn't resolve.

jmarrec:
> I swapped the prioritize wstring to only be active on MSVC, not sure if that should be all windows or just visual studio?

Great that you limited it to not affect Unix, I think that's the right decision!

Ultimately, it should be all of windows and not just MSVC. std::filesystem::path is wide on Windows as a whole (I have a doubt for cygwin though)

PR #878, opened 2023-05-05
Fix #875 - Unicode support: automatically widen on windows when option is assigned a std::filesystem::path

### LINKED ISSUE #875: Unicode support: automatically widen on windows when option is assigned a std::filesystem::path

jmarrec:
- Fix #875

Tested locally on Ubuntu 20.04 and Windows 10.

### DIFF
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -661,8 +661,8 @@ template <typename T>
 struct classify_object<
     T,
     typename std::enable_if<!std::is_floating_point<T>::value && !std::is_integral<T>::value &&
-                            (!std::is_assignable<T &, std::string>::value || force_widen<T>::value) &&
-                            !std::is_constructible<T, std::string>::value &&
+                            (force_widen<T>::value || (!std::is_assignable<T &, std::string>::value &&
+                                                       !std::is_constructible<T, std::string>::value)) &&
                             std::is_assignable<T &, std::wstring>::value>::type> {
     static constexpr object_category value{object_category::wstring_assignable};
 };
@@ -671,8 +671,8 @@ template <typename T>
 struct classify_object<
     T,
     typename std::enable_if<!std::is_floating_point<T>::value && !std::is_integral<T>::value &&
-                            (!std::is_assignable<T &, std::string>::value || force_widen<T>::value) &&
-                            !std::is_constructible<T, std::string>::value &&
+                            (force_widen<T>::value || (!std::is_assignable<T &, std::string>::value &&
+                                                       !std::is_constructible<T, std::string>::value)) &&
                             !std::is_assignable<T &, std::wstring>::value && (type_count<T>::value == 1) &&
                             std::is_constructible<T, std::wstring>::value>::type> {
     static constexpr object_category value{object_category::wstring_constructible};
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -18,6 +18,10 @@
 #include <vector>
 // [CLI11:public_includes:end]
 
+#if defined CLI11_HAS_FILESYSTEM && CLI11_HAS_FILESYSTEM > 0
+#include <filesystem>
+#endif  // CLI11_HAS_FILESYSTEM
+
 #include "Encoding.hpp"
 #include "StringTools.hpp"
 
@@ -73,6 +77,18 @@ template <typename T> struct is_shared_ptr<const std::shared_ptr<T>> : std::true
 template <typename T> struct is_copyable_ptr {
     static bool const value = is_shared_ptr<T>::value || std::is_pointer<T>::value;
 };
+
+template <typename T> struct is_path : std::false_type {};
+template <typename T> struct force_widen : std::false_type {};
+
+#if defined CLI11_HAS_FILESYSTEM && CLI11_HAS_FILESYSTEM > 0
+template <> struct is_path<std::filesystem::path> : std::true_type {};
+#if _WIN32
+template <> struct force_widen<std::filesystem::path> : std::true_type {};
+#endif
+// struct is_path : std::is_same<std::filesystem::path,
+//                              typename std::remove_cv<T>::type>::value {};
+#endif
 
 /// This can be specialized to override the type deduction for IsMember.
 template <typename T> struct IsMemberType {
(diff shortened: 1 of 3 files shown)


include/CLI/TypeTools.hpp
```diff
@@ -43,6 +47,17 @@ constexpr enabler dummy = {};
...
+// C++14
+template <class T> using remove_cv_t = typename std::remove_cv<T>::type;
+
+template <class T> using remove_reference_t = typename std::remove_reference<T>::type;
+
+// C++20
+template <class T> struct remove_cvref {
+    using type = remove_cv_t<remove_reference_t<T>>;
+};
+template <class T> using remove_cvref_t = typename remove_cvref<T>::type;
```
jmarrec, line 59: Backport some helpers to be in C++11 for readability

include/CLI/TypeTools.hpp
```diff
@@ -18,6 +18,10 @@
 #include <vector>
 // [CLI11:public_includes:end]
 
+#if defined CLI11_HAS_FILESYSTEM && CLI11_HAS_FILESYSTEM > 0
+#include <filesystem>
+#endif  // CLI11_HAS_FILESYSTEM
```
jmarrec, line 23: conditional import

include/CLI/TypeTools.hpp
```diff
@@ -74,6 +89,20 @@ template <typename T> struct is_copyable_ptr {
...
+#if defined CLI11_HAS_FILESYSTEM && CLI11_HAS_FILESYSTEM > 0
+template <typename T>
+struct is_path<T, typename std::enable_if<std::is_same<std::filesystem::path, remove_cvref_t<T>>::value>::type>
+    : std::true_type {};
+#if _WIN32
+template <typename T> struct force_widen<T, enable_if_t<is_path<T>::value>> : std::true_type {};
+#endif
+#else
+template <typename T> struct is_path : std::false_type {};
+#endif
```
jmarrec, line 104: Detect if T is a path, exists even if <filesystem> isn't used. I'm using remove_cvref, but that may not be necessary (but it doesn't hurt much)

if WIN32 => force_widen

include/CLI/TypeTools.hpp
```diff
@@ -610,9 +639,10 @@ template <typename T> struct classify_object<T, typename std::enable_if<std::is_
 
 /// String and similar direct assignment
 template <typename T>
-struct classify_object<T,
-                       typename std::enable_if<!std::is_floating_point<T>::value && !std::is_integral<T>::value &&
-                                               std::is_assignable<T &, std::string>::value>::type> {
+struct classify_object<
+    T,
+    typename std::enable_if<!std::is_floating_point<T>::value && !std::is_integral<T>::value &&
+                            std::is_assignable<T &, std::string>::value && !force_widen<T>::value>::type> {
```
jmarrec, line 645: Here basically, I make it so that an object, in the event that it is both assignable and/or constructible from `std::string `AND from `std::wstring`, will be classified as `wstring_assignable` / `wstring_constructible` (the wide one) if it's `force_widen`

include/CLI/TypeTools.hpp
jmarrec: I wrote it this way so force_widen could be extended to other types later if needed.

tests/CMakeLists.txt
```diff
@@ -69,7 +69,7 @@ set(CLI11_MULTIONLY_TESTS TimerTest)
 find_package(Catch2 CONFIG)
 
 if(Catch2_FOUND)
-  if(NOT TARGET Catch2::Catch2)
+  if(NOT (TARGET Catch2::Catch2 OR TARGET Catch2::Catch2WithMain))
```
jmarrec, line 72: Unrelated: I tried to build with Catch2 v3, hit this error. Then I hit another one and gave up (later it's trying to add_sources to catch_main whcih is just an ALIAS in the case it's v3).

tests/HelpersTest.cpp
```diff
@@ -21,6 +21,10 @@
 #include <unordered_map>
 #include <utility>
 
+#if defined CLI11_HAS_FILESYSTEM && CLI11_HAS_FILESYSTEM > 0
+#include <filesystem>
+#endif  // CLI11_HAS_FILESYSTEM
```
jmarrec, line 26: Conditional import in test

tests/HelpersTest.cpp
```diff
@@ -1308,6 +1312,44 @@ TEST_CASE("Types: LexicalConversionComplex", "[helpers]") {
...
+TEST_CASE("Types: LexicalConversionPath", "[helpers]") {
+    CLI::results_t input = {"TéstFileNotUsed.txt"};
+    std::filesystem::path x;
+    bool res = CLI::detail::lexical_conversion<decltype(x), decltype(x)>(input, x);
+    CHECK(res);
+    CHECK(CLI::to_path("TéstFileNotUsed.txt").native() == x.native());
+}
+
+#endif  // CLI11_HAS_FILESYSTEM
+
```
jmarrec, line 1352: conditional testing: lots of static_assert to make sure it's behaving correctly, and an actual runtime test that fails before, works now.

phlptp:
this issue was resolved by #876, if there are unresolved issues from that please feel free to resubmit.  
Thanks

PR #896, opened 2023-06-28
fix: tests were broken on Catch3

henryiii:
These were invalid on an imported target (and was incorrect anyway).


henryiii:
FYI, I think this was introduced in #804.

PR #898, opened 2023-06-28
fix: increase min CMake to 3.5

henryiii:
CMake 3.27 will start warning if this is older than 3.5.

### DIFF
--- a/.github/actions/quick_cmake/action.yml
+++ b/.github/actions/quick_cmake/action.yml
@@ -1,5 +1,5 @@
 name: Quick CMake config
-description: "Runs CMake 3.4+ (if already setup)"
+description: "Runs CMake 3.5+ (if already setup)"
 inputs:
   args:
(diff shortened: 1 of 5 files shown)


phlptp (approved):
I think 3.5 keep this running on a default Ubuntu 18.04 and RHEL 6 so should be good

henryiii:
Checking https://cliutils.gitlab.io/modern-cmake/chapters/intro/installing.html, 18.04 used 3.10. 16.04 used 3.5. Though both are EoL, both should be fine. :) 

RHEL 6 is weird, as "cmake" continued to point at 2.8, as if 3 was a breaking change (it wasn't). Guessing this is referring to EPEL 6, which is .

phlptp:
If I remember right RHEL 6 had a cmake3 package that had cmake 3.6, and it was a bit weird but at least it was available as a package.    Been a while since I used that version though.

PR #900, opened 2023-06-29
add optional to cpack include in cmake

Jo5ta:
If cpack was already include, including it again without the `optional` argument creates an error. With the following message:
```
[cmake] CMake Warning at /usr/share/cmake/Modules/CPack.cmake:507 (message):
[cmake]   CPack.cmake has already been included!!
```

`optional` fixes this based on this documentation. https://cmake.org/cmake/help/v3.5/command/include.html And it is compatible with the current minimal cmake version 3.5


phlptp:
this does not appear to be valid cmake.   What we have done in other projects is wrap the entire CPACK section in a conditional variable, so it can be completely turned off if not needed.  guessing packaging isn't done that often for CLI11 so it wouldn't affect much to have it disabled by default.

henryiii:
The OPTIONAL is backward. And I also think it doesn't do what you want it to do - it avoids an error if it's missing, not an error if it's already been included. You probably could protect it with something that CPack defines, though. (Or disable by default is also fine).

henryiii:
Also, is that an error or a warning?

henryiii:
Actually, we should probably only do the CPack stuff if this is the main project. That would fix this.

PR #901, opened 2023-06-29
Doc installation

phlptp:
This PR includes updates from #774 and #613.  
It updates the readme and docbook with some additional installation instructions.

### DIFF
--- a/book/chapters/installation.md
+++ b/book/chapters/installation.md
@@ -9,7 +9,7 @@
 This example uses the single file edition of CLI11. You can download `CLI11.hpp`
 from the latest release and put it into the same folder as your source code,
 then compile this with C++ enabled. For a larger project, you can just put this
-in an include folder and you are set.
+in an include folder and you are set.  This is the simplest and most straightforward means of including CLI11 with a project.
 
 ## Full edition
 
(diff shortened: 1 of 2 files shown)


phlptp:
@henryiii  did we want to reduce the installation section of the readme and include some of the stuff that is in that section in the docbook?

henryiii:
We can, we could just show the main way + list the others shown in the gitbook.

FYI, I'm going to need to migrate off of GitBook in the near future, it's pretty well past EoL. Maybe a JupyterBook, or a Hugo/jekell/sphinx/mkdocs solution would work, not sure yet what would be best.

phlptp:
we use sphinx and readthedocs in most of my other projects.    That handles markdown, though I would imagine any changes would require some level of restructuring.

henryiii (approved):

book/chapters/installation.md
```diff
@@ -46,6 +47,83 @@ necessary `lib/cmake/CLI11/CLI11Config.cmake` files, so
...
+    QUIET
+    GIT_REPOSITORY https://github.com/CLIUtils/CLI11.git
+    GIT_TAG v2.3.2
+    SOURCE_DIR     cli11_proj
+)
+
+FetchContent_MakeAvailable(cli11)
+
+# And now you can use it
+add_subdirectory(${cli11_proj_SOURCE_DIR} ${cli11_proj_SOURCE_DIR}/build)
```
wirew0rm, line 111: Is there any reason the additional `add_subdirectory` call and the `SOURCE_DIR` parameter have been added here compared to the more idiomatic version that was in the readme before?

The second one just bit me on a CI system with caching for the default FetchContent directories set up.

I can prepare a small PR but I wanted to ask first since it looks like it has been deliberately changed.

PR #902, opened 2023-06-29
fix: only use CPack if this is the main project

henryiii:
Alternative fix to, and closes #900.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,39 +198,41 @@ if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND EXISTS "${CMAKE_CURRENT_SOURCE_D
 endif()
 
 # Packaging support
-set(CPACK_PACKAGE_VENDOR "github.com/CLIUtils/CLI11")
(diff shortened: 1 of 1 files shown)


Jo5ta:
it would also be nice to have also a option to enable this.

henryiii:
Why? If you want to pack CLI11 into a parent project, I don't think this works, does it? I don't think CPack is aware of subprojects. Though I'm not sure about that.

henryiii:
Most of the variables don't seem to be useful in the context of a superproject, IMO. The parent project would want to set its own name - CPACK_DEBIAN_PACKAGE_NAME isn't correct if you aren't CLI11! Etc. I don't think anything would be reusable if you aren't running CLI11 directly.

phlptp:
I was thinking I might want a variable to disable this normally.  Since most of the time I don't want any packaging regardless if it was  the main project or not.    Then if it was enabled by default if it is the main project, disabled by default if it was not, either way in the unusual use cases where you want something different, there is a variable setting that can change it.

henryiii:
It doesn't package by default - you have to run CPack. The only thing this does is set some variables and add two targets (that you don't have to build - they are not in ALL).

PR #913, opened 2023-08-04
fix: build on Windows ARM64

peterh:
Regression introduced in commit 3cd730b18b563fdd48cd95dfb0174551aa18c3b3

### DIFF
--- a/include/CLI/impl/Argv_inl.hpp
+++ b/include/CLI/impl/Argv_inl.hpp
@@ -29,6 +29,10 @@
 #define _X86_
 #elif defined(__arm__) || defined(_M_ARM) || defined(_M_ARMT)
 #define _ARM_
+#elif defined(__aarch64__) || defined(_M_ARM64)
+#define _ARM64_
+#elif defined(_M_ARM64EC)
+#define _ARM64EC_
 #endif
 #endif
 


phlptp:
What was the error your were seeing?

peterh:
The compile error is:
```
C:\Program Files (x86)\Windows Kits\10\\include\10.0.22621.0\\um\winnt.h(169): fatal error C1189: #error:  "No Target Architecture"
```

You can reproduce it when building for X86_64 by removing the \_AMD64_ lines just above where I added the \_ARM64_ lines.

phlptp (approved):

phlptp:
Ok,  would having an ARM build on visual studio in a CI test have caught this error?

peterh:
> Ok, would having an ARM build on visual studio in a CI test have caught this error?

The question I think you were asking: Yes: Having a 64-bit ARM build on Visual Studio in a CI would have caught this error.

The question you actually asked: No: In Windows nomenclature, ARM without qualifications is a 32-bit ARM build, and this configuration was already supported. Only 64-bit ARM was broken. 😁

phlptp:
That explains that then,  I was pretty sure I had built this many times on Visual studio ARM but quite possibly not the ARM64 build.

Issue #911, opened 2023-07-22
Exported pkg-config ignores CLI11_PRECOMPILED


kiwixz:
When configuring CLI11 with `CLI11_PRECOMPILED`, the CMake config will import the static library as expected; but the pkg-config will stay the same and just include the header-only version.

phlptp:
that sounds like an oversight.  We will take a look

PR #916, opened 2023-08-15
fix: CMake package and pkgconfig rework

phlptp:
add tests that exercise the CMake module and pkg-config files


phlptp:
FIxes #911

.github/workflows/tests.yml
```diff
@@ -35,7 +35,7 @@ jobs:
           cmake -S . -B build \
             -DCMAKE_CXX_STANDARD=${{matrix.std}} \
             -DCLI11_SINGLE_FILE_TESTS=OFF \
-            -DCLI11_EXAMPLES=OFF \
```
phlptp, line 38: I think this change is what caused the coverage change, but I don't think the examples should be built or tested during the single file tests,  not entirely sure what is going on.

henryiii: Coverage upload is finicky, it's possible it just had too many jobs not able to report. Going to try a rebuild to see. Otherwise, not too worked about it.

phlptp: I should take another crack at getting back to 100% sometime soon.  Probably not in this PR though

henryiii: The failing lines are

https://github.com/phlptp/CLI11/blob/1cb75471863bb364038de3a5ef8b596694945990/include/CLI/impl/App_inl.hpp#L608-L609

https://github.com/phlptp/CLI11/blob/1cb75471863bb364038de3a5ef8b596694945990/include/CLI/impl/Config_inl.hpp#L250-L251

https://github.com/phlptp/CLI11/blob/1cb75471863bb364038de3a5ef8b596694945990/include/CLI/impl/Config_inl.hpp#L342-L343

I think they were uncovered before, but for some reason they are now counting as a slightly higher percentage? Coverage is weird unless it's 100%. :/

Not worried about it for this PR.

PR #918, opened 2023-08-22
feat: add a reverse multi option policy

phlptp:
use it for the default in `set_config` and simplify and add more flexibility to the the config processing, and potentially in other options as well.

The reverse policy returns a vector but in reversed order from normal.  This is what we want in the config processing 

Inspired by #862, and updated with recent code changes.


phlptp:
@VolkerChristian  I think this does what you were thinking in #862?  I borrowed some code from that PR and added you as a co-author.

henryiii (approved):

PR #923, opened 2023-09-18
Unicode API rework

### LINKED ISSUE #845: `CLI::argv()` crashes on linux when `/proc` is missing
When `/proc` is missing on a Linux system, calling `CLI::argv()` fails with `could not open /proc/self/cmdline for reading`.

I was aware of this while implementing Unicode, but thought that this is an extraordinary scenario of a broken Linux filesystem. However apparently this is a common occurrence when a user calls [`chroot`](https://man7.org/linux/man-pages/man1/chroot.1.html) before running the application.

There is another way of obtaining argc and argv, but I'm a little scared of it. It does seem to work though.
```c++
#include <unistd.h>
#include <utility>

std::pair<int, char**> findargs() {
	int i;
	char** p = __environ - 2;

	for (i = 1; i != *reinterpret_cast<int*>(p - 1); i++) {
		p--;
	}

	return {i, p};
}
```
The idea here is to obtain a stack address nearby from `__environ`, then walk back, counting steps. When dereferencing a pointer as-if an int* reveals that it is equal to the amount of steps, it means we have reached a pointer to argc and walked through the entire array of argv.

One problem is that this has to be done before `main`, otherwise user might call `setenv` and change the pointer to environ (I presume). And if this is prepared in advance we might as well discard the `/proc/self/cmdline` approach altogether. At least no files are read so the setup would be pretty fast.

bindreams:
Fixes #845 as discussed.

Comparing the two approaches of getting `argv`:
1. The "old" way, through `CLI::argv()`:
    ✔️ Works automatically and almost everywhere
    ❌ Small abstraction overhead on macOS
    ❌ Does not work in weird edge-cases such as missing `/proc`
2. This PR, through `app.ensure_utf8`:
    ✔️ True zero-overhead abstraction: you don't pay for what you don't use
    ✔️ Less moving parts than the "old" approach, probably can't be broken
    ❌ Requires extra code to be written by the user (which is sad because ideally everyone should use this by default)

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,7 +207,9 @@ int main(int argc, char** argv) {
 }
 ```
 
-For more information about 🚧`ensure_utf8` the section on [Unicode support](#unicode-support) below. The 🚧`ensure_utf8` function is only available in main currently and not in a release.
+For more information about 🚧`ensure_utf8` the section on
+[Unicode support](#unicode-support) below. The 🚧`ensure_utf8` function is only
+available in main currently and not in a release.
 
 <details><summary>Note: If you don't like macros, this is what that macro expands to: (click to expand)</summary><p>
 
@@ -1405,9 +1407,15 @@ CLI11 supports Unicode and wide strings as defined in the
 
 When using the command line on Windows with unicode arguments, your `main`
 function may already receive broken Unicode. Parsing `argv` at that point will
-not give you a correct string. To fix this, you have three good options and two bad ones:
+not give you a correct string. To fix this, you have three good options and two
+bad ones:
 
-1. Replace `argv` with `app.ensure_utf8(argv)` before any arguments are parsed. `ensure_utf8` will do nothing on systems where `argv` is already in UTF-8 (Such as Linux or macOS) and return `argv` unmodified. On Windows, it will discard `argv` and replace it with a correctly decoded array or arguments from win32 API.
+1. Replace `argv` with `app.ensure_utf8(argv)` before any arguments are parsed.
+   `ensure_utf8` will do nothing on systems where `argv` is already in UTF-8
+   (Such as Linux or macOS) and return `argv` unmodified. On Windows, it will
+   discard `argv` and replace it with a correctly decoded array or arguments
+   from win32 API.
+
    ```cpp
    int main(int argc, char** argv) {
        CLI::App app;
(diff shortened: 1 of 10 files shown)


bindreams:
@henryiii PTAL. Implemented this before your comment about axing `CLI::argv`. The way it works here is `CLI::argv` is available but completely optional. Do I still cut it out? I'll be honest, I'm sad to see it go 😢.

henryiii:
I was thinking this:

```cpp
     argv = app.ensure_utf8(argc, argv);
     CLI11_PARSE(app);
     return 0;
}
```

That is, calling `ensure_utf8` stores the (new on windows) argv in the app on all platforms. Then it's legal to call `CLI11_PARSE` with just app, no additional magic needed. That's a _little_ bit more complicated than before, but slightly better than having to list argc/argv again. You don't even need the `argv =` part if you aren't going to use it later. If it hasn't been stored, I was originally thinking we could make `CLI11_PARSE(app)` an error, though if you still wanted to have the `CLI::argv()` work, then it could use that.

The downside is it's not as obvious how inject modification of argv in this form, of course.

These are the possible usages:

```cpp
// 1
app.ensure_utf8(argc, argv);
CLI11_PARSE(app);  // Uses stored argv vector on all platforms

// 2
argv = app.ensure_utf8(argc, argv);
CLI11_PARSE(app, argc, argv);  // Uses produced argv vector

// 3
CLI11_PARSE(app, argc, argv); // Not unicode

// 4
CLI11_PARSE(app); // Not portable, but unicode
```

My favorite probably is 2, as it's explicit, but 1 is pretty simple looking for most users who don't manipulate argv first, and can be written as 2 if a user does need to. I'm fine to drop 4 as too magical and not portable, but if we keep `CLI::argv()`, we could keep that too. Some people might like the brevity and not care about portability. But 1 is pretty simple.

henryiii:
This looks like an improvement, my main question is, would it make sense to support (1) above and maybe drop (4)? (In later PRs)

henryiii:
@phlptp thoughts about API in follow-ups? I’ll approve this PR tomorrow.

henryiii (approved):

phlptp:
I don't have any strong opinions on the unicode stuff, so this seems reasonable to me.

bindreams:
Thanks for the merge!

Would you mind putting in a word with the all-contributors bot about my contributions? Sorry for being indiscreet.

henryiii:
@all-contributors, please add @andreasxp for code

henryiii:
(Not totally sure the bot is still working, we might need to update manually, feel free to ping (a lot) if we forget!)

henryiii:
Done in #977.

PR #926, opened 2023-09-23
Env validation

### LINKED ISSUE #925: Displaying the help should take precedence over checking env vars & other possible checks
This is an issue which was only slightly mentioned in #903, so I will describe this specific case here in more details.

Basic example:

```cpp
auto CliApp = new CLI::App("myapp");
auto sub1 = CliApp->add_subcommand("sub1");
std::filesystem::path someFile = "";
sub1->add_option("-f,--file", someFile)->envname("SOME_FILE")->required()->expected(1)->check(CLI::ExistingFile);

try
{
	CliApp->parse(argsCount, args);
}
catch (const CLI::ParseError& e)
{
	return CliApp->exit(e);
}
```

Make sure `%SOME_FILE%` env var is defined as invalid/non-existing file path.
Then call the app like that: `myapp.exe sub1 -h`:
-Normally it should only display the help about `sub1` expectations & requirements and exit, as that's a completely separate control flow path for the app, just like displaying the app version with `-v`.
-Actually it fails `CLI::ExistingFile` validator check, suggesting to `Run with --help for more information.`. But it's impossible to call for help.

phlptp:
Fixes #925


phlptp:
The environmental variable results must now pass any existing validators to be included.  This resolves an issue where help could not be reached.

henryiii (approved):

PR #928, opened 2023-10-05
add additional tests for coverage,

phlptp:
Add some additional tests to try to get test coverage back to 100%

 refactor the positional parsing to  reduce code duplication

### DIFF
--- a/cmake/CodeCoverage.cmake
+++ b/cmake/CodeCoverage.cmake
@@ -88,7 +88,7 @@ elseif(NOT CMAKE_COMPILER_IS_GNUCXX)
 endif()
 
 set(COVERAGE_COMPILER_FLAGS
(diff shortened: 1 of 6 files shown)


henryiii (approved):

phlptp:
@henryiii I am going to merge this in one day unless you object.  (2023-10-7)
There are no API changes only some code refactoring, new tests, and a fix for an edge case on specifically formatted inline comments on flags.  then I will try to tackle the coverage in some of the other files, but this fixes app_inl and config_inl to 100% coverage.

henryiii:
I've already approved so you can merge any time (you don't have to have my approval unless it's potentially controversial or something). And these sorts of cleanups & added tests don't need much review, IMO.

PR #935, opened 2023-10-14
Multiline string config

phlptp:
Support for TOML style multiline string values and comments


phlptp:
This is necessary to fix the issue exposed by #930

phlptp:
The basic issue is that it is possible to get multiline strings as a value into CLI11, and if those values are made into a config string they generate config files that can't be read, and cause some very strange crossover effects and errors, so the solution is that CLI11 needs to be able to have some way to generate and support strings containing newlines.   The obvious solution is to support TOML multiline strings in the config files.  Which is what this PR does.

Issue #958, opened 2023-12-14
'usage()' method missing from single header file


tmcayden:
As referenced in the [API docs](https://cliutils.github.io/CLI11/class_c_l_i_1_1_app.html#a6cb159f2ebdfeb4468f93a6f873f3263), `usage()` is a member of `CLI::App` that takes a string or function as an input to set a custom "usage" description for your menu. This is defined and declared in in CLI11/include/CLI/App.hpp:978. I found that this method is missing from the single header file. Thus, when attempting to call `app.header("header text");` in my program it results in the error: "'class CLI::App' has no member named 'usage'".

The neighboring method in `App.hpp` to `usage()` is `footer()`, which is found on line 6485 in the single header file.`usage()` should be defined above this, but is missing.

phlptp:
I will take a look,  thanks

phlptp:
I verified the usage feature is in the single header file.   It looks like that feature was added recently and we forgot to add a marker on the docs indicating it is only available on the latest code.   So if you want it update to main.  It just isn't in the 2.3.2 release.  And the readme was missing an indicator, which I will remedy shortly.  Hopefully we can also get a release out soon as well that will include it.

PR #959, opened 2023-12-16
update docs on usage to indicate it is only available on main

phlptp:
fix missing doc indicator on usage method


phlptp:
Close issue #958

PR #964, opened 2023-12-25
regular and literal strings

phlptp:
Add escaping to quoted strings, differentiate between literal and regular strings.

The goal is to make string processing in config files as close as possible to toml standards.  This means handing escape sequences including unicode, and differentiating between literal strings and regular strings in files and when splitting the command line.  Also allowing variable names in the files to be quoted.   

This PR gets partway there.    Removes some hacks from the previous PR to deal with unusual option names and replaces with the quoted names.


phlptp:
@henryiii this PR handles string literals and escaped string differently including unicode escape sequences.   Basically working towards more complete TOML compliance.  The next one will add a few more capabilities like a transformer that can handle escaped strings and some additional documentation on the topic.  This changes the way a few edge cases are handled, and fixes a few more edge cases the fuzzer caught.  I will merge this on 2023-12-30 unless you have some comments or issues.

PR #967, opened 2023-12-30
add support for quotes in the config naming to match TOML standard

phlptp:
This PR is to further support for TOML. To allow and generate quoted names in config files including those separated by the parent separator.  

like 
```toml
"sub"."sub2".value=1
'sub'.'sub.sub'.value=2
```


phlptp:
2 more steps after this one.   Update the value handlers to match TOML specifications and then update documentation.

PR #970, opened 2024-01-04
Escape transform and docs

phlptp:
Update some documentation and add a string escape transformer so escaped strings can be handled on the command line as well as in the config files.


README.md
```diff
@@ -452,7 +452,7 @@ Before parsing, you can set the following options:
   are `CLI::MultiOptionPolicy::Throw`, `CLI::MultiOptionPolicy::Throw`,
   `CLI::MultiOptionPolicy::TakeLast`, `CLI::MultiOptionPolicy::TakeFirst`,
   `CLI::MultiOptionPolicy::Join`, `CLI::MultiOptionPolicy::TakeAll`, and
```
henryiii: Does the `and` need to move?

henryiii (commented):
FYI, I've been lightly looking over these but not in-depth reviewing them - let me know if there's something specific that needs a second set of eyes.

phlptp:
@henryiii  after this I will update the copyright dates to 2024, and then work on a release update (changelog, readme, etc).   Is there anything else you wanted to change before a 2.4 release?

PR #972, opened 2024-01-06
update supported cmake versions and add some tests with newer compilers

phlptp:
Update and test with some newer compilers and cmake versions


phlptp:
the new clang-tidy recommended getting rid of all std::endl  also new fuzzing failure.

phlptp:
@henryiii  most changes are getting rid of `std::endl`.  The others are updating a few CI builds and adding a few more.  And a few other clang-tidy fixes.

PR #975, opened 2024-01-08
Full help/usage display and modification example

tansy:
Example of help modification as documentation supplement.


phlptp:
Thanks this looks good, I will integrate after we get the release out so it gets tested and built with all the other examples, but I think this will be helpful.

tansy:
That 'no parameter option' was my guess, and I just noticed, that it has \`level' parameter. I think that needs correction.

Shall it be
`encode->add_option("-R, --remove", nullptr, "remove input file");`
or
`encode->add_flag("-R, --remove", "remove input file");`?

Also I world add an option to make space between shirt and long option description (`-h,--help` -> `-h, --help`), but don't know how.

examples/help_usage.cpp
```diff
@@ -0,0 +1,72 @@
...
+    encode->add_option("input", input_file_name, "input file")
+        ->option_text(" ")
+        ->required()
+        ->check(CLI::ExistingFile);                                                               // file must exist
+    encode->add_option("output", output_file_name, "output file")->option_text(" ")->required();  // required option
+    encode->add_option("-l, --level", level, "encoding level")
+        ->option_text("[1..9]")
+        ->check(CLI::Range(1, 9))
+        ->default_val(5);                                            // limit parameter range
+    encode->add_option("-R, --remove", level, "remove input file");  // no parameter option
```
phlptp: level here is used in another option, should probably be add_flag so it doesn't default to requiring an additional argument.   this matches your other comment

phlptp:
```suggestion
    encode->add_flag("-R, --remove", "remove input file");  // no parameter option
```

phlptp:
Going to merge this, as the failures are unrelated (an update in test runners).   Will clean it up soon in a subsequent PR.

PR #979, opened 2024-01-11
rename the `parse()` method and adjust documentation

phlptp:
Based on discussions in #974

### DIFF
--- a/README.md
+++ b/README.md
@@ -1409,7 +1409,7 @@ app.add_option("--fancy-count", [](std::vector<std::string> val){
     });
 ```
 
-### Unicode support 
+### Unicode support
 
 🆕 CLI11 supports Unicode and wide strings as defined in the
 [UTF-8 Everywhere](http://utf8everywhere.org/) manifesto. In particular:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -196,7 +196,6 @@ this:
 ```cpp
-    argv = app.ensure_utf8(argv);
```
henryiii, line 199: Why is this removed? This is by far the best way to do this - `parse()` / `parse_from_cli_args()` does black magic on some operating systems, while this reliably works everywhere.

README.md
```diff
@@ -1460,7 +1457,11 @@ itself.
 > int main() {
 >     CLI::App app;
 >     // ...
->     CLI11_PARSE(app);
```
henryiii, line 1463: The point of `CLI11_PARSE(one arg)` was to make this simpler to type, encouraging people to support unicode. This makes this "short" but magical method four lines long instead! It would better to just use the more reliable and safe method then (`argv = app.ensure_utf8(argv);`), which is only one line longer.

README.md
phlptp: In this section of the documentation it is covering the most basic application,  in a vast majority of applications this is not necessary.  I can add it back if you think it valuable, but to me it creates a lot of potentially unnecessary questions when just getting started.

phlptp:
Superseded by #987

PR #980, opened 2024-01-12
tests: fix catch2 3 support

henryiii:
Is there an easy way to pre-install this in CI or download it (perhaps in meson)? It's bad that we aren't running in CI with version 3 (but don't want to drop 2, since we can't drop it until we drop C++11).

### DIFF
--- a/tests/OptionalTest.cpp
+++ b/tests/OptionalTest.cpp
@@ -53,6 +53,10 @@
 #endif
 // [CLI11:verbatim]
 
+TEST_CASE("OptionalNoEmpty") {
+    CHECK(1 == 1);
+}
+
 #if CLI11_STD_OPTIONAL
 
 #ifdef _MSC_VER
--- a/tests/OptionalTest.cpp
+++ b/tests/OptionalTest.cpp
@@ -53,9 +53,7 @@
 #endif
 // [CLI11:verbatim]
 
-TEST_CASE("OptionalNoEmpty") {
-    CHECK(1 == 1);
-}
+TEST_CASE("OptionalNoEmpty") { CHECK(1 == 1); }
 
 #if CLI11_STD_OPTIONAL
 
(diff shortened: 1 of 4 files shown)


phlptp:
Definitely don't want to do that this Release, but at what point do we consider upping the minimum compiler support to C++14?

henryiii:
That's tricky, since "11" is in our name. :) I suppose it needs to be done at some point, though. Don't have a strong opinion, 14+ won't actually affect us too much, 17+ would be larger. 14+ would actually affect pybind11 more (which also has 11 in the name, and also hasn't increased it yet...)

henryiii:
If we could go 23+ I'd use `std::result` instead of exception handling. :)

phlptp:
Most of my other code is running C++17 as min now,   Might move some to C++20, this summer after Ubuntu 24.04 comes out.   Not going to be able to move to 23, likely until at least late '26

tests/OptionalTest.cpp
```diff
@@ -53,6 +53,8 @@
 #endif
 // [CLI11:verbatim]
 
+TEST_CASE("OptionalNoEmpty") { CHECK(1 == 1); }
```
henryiii, line 56: Not sure if there's a nicer way to make Catch2 (maybe only 3+?) not error on no tests run in a file.

henryiii: I found `--allow-running-no-tests` but that's probably only in 3+, and we still support 2, and don't want to have to figure out how to make it optimally get added in meson (or bother with optionally adding it in CMake). So I think this is a fine workaround for now. (Plus it only affects this file, while the flag is tricky to add just for one file)

henryiii:
Okay, maybe we can target moving to 17+ this year. Might see what pybind11 does, too, we could match that (even though they are completely unrelated, but it's simpler if the projects I maintain are similar ;) )

tests/OptionalTest.cpp
phlptp: we could either figure out how to get CMake to do the same detection and not generate that executable,  or we could just add that whole file to the optionTypeTests and delete the optionalTest file,  then it wouldn't be an issue.

PR #996, opened 2024-02-05
fix: transiative include issue

### LINKED ISSUE #993: Tests fail to build due to missing <cstdint> with 2.4.0
Assume my currently used toolchain/libraries (on Gentoo Linux) removes a transitively-used header (haven't dug into exact cause) if CI here hasn't caught it.

`include/CLI/impl/StringTools_inl.hpp` has several use of `uint32_t` but does not include cstdint that I can see, including it fixes it for me.

```
/tmp/portage/dev-cpp/cli11-2.4.0/work/CLI11-2.4.0/include/CLI/impl/StringTools_inl.hpp:218:19: error: 'uint32_t' in namespace 'std' does not name a type; did you mean 'wint_t'?
  218 | CLI11_INLINE std::uint32_t hexConvert(char hc) {
      |                   ^~~~~~~~
      |                   wint_t
```

henryiii:
Fix #993. I tried include-what-you-use, but it reports too much stuff to be able to pick out what's missing very effectively.

### DIFF
--- a/include/CLI/impl/StringTools_inl.hpp
+++ b/include/CLI/impl/StringTools_inl.hpp
@@ -10,6 +10,7 @@
 #include <CLI/StringTools.hpp>
 
 // [CLI11:public_includes:set]
+#include <cstdint>
 #include <string>
 #include <utility>
 #include <vector>


Spixmaster:
@henryiii With clang v17, `clang-tidy` has [misc-include-cleaner](https://clang.llvm.org/extra/clang-tidy/checks/misc/include-cleaner.html) which seems to be better than `include-what-you-use` from two days of testing.

PR #1001, opened 2024-02-06
chore: prepare for 2.4.1

henryiii:
Quick release fixing the missing header.


phlptp:
If you can wait a day I found an issue with some config file setup that should be in a patch as well.  I should be able to get that PR in the next day.

henryiii:
That’s fine, though you’ll need to release it, we are having a baby. :)

PR #1008, opened 2024-02-18
Add IWYU pragmas

### LINKED ISSUE #816: clangd warnings on transitive includes
In the README, it is recommended to include the following three headers in your application:

#include "CLI/App.hpp"
#include "CLI/Formatter.hpp"
#include "CLI/Config.hpp"

This works well, with the exception that clangd raises warning:

image

If I remove the two headers below, then I get a compilation error:

```
/usr/bin/ld: CMakeFiles/bth.dir/bth.cpp.o: warning: relocation against `_ZTVN3CLI9FormatterE' in read-only section `.text._ZN3CLI9FormatterC2Ev[_ZN3CLI9FormatterC5Ev]'
/usr/bin/ld: CMakeFiles/bth.dir/bth.cpp.o: in function `CLI::ConfigBase::ConfigBase()':
/opt/CLI11/include/CLI/ConfigFwd.hpp:81: undefined reference to `vtable for CLI::ConfigBase'
/usr/bin/ld: CMakeFiles/bth.dir/bth.cpp.o: in function `CLI::Formatter::Formatter()':
/opt/CLI11/include/CLI/FormatterFwd.hpp:120: undefined reference to `vtable for CLI::Formatter'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
collect2: error: ld returned 1 exit status
```

It would be nice if the warnings raised by clangd were suppressed via IWYU pragma comments on the CLI11 header files:
https://clangd.llvm.org/guides/include-cleaner#scenarios-and-solutions

If I add 
`#include "CLI/CLI.hpp"`
instead of the three above, I get a similar warning, because of the transitive includes.

gostefan:
Added include-what-you-use pragmas to:
* let IWYU point users to the main include file CLI/CLI.hpp
* tell IWYU that CLI/CLI.hpp is the main exporting header.

This should fix #816

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -1439,5 +1439,5 @@ struct AppFriend {
 }  // namespace CLI
 
 #ifndef CLI11_COMPILE
-#include "impl/App_inl.hpp" // IWYU pragma: export
+#include "impl/App_inl.hpp"  // IWYU pragma: export
 #endif
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -6,6 +6,8 @@
 
 #pragma once
 
+// IWYU pragma: private, include "CLI/CLI.hpp"
+
 // [CLI11:public_includes:set]
 #include <algorithm>
 #include <cstdint>
@@ -1437,5 +1439,5 @@ struct AppFriend {
 }  // namespace CLI
 
 #ifndef CLI11_COMPILE
-#include "impl/App_inl.hpp"
+#include "impl/App_inl.hpp" // IWYU pragma: export
 #endif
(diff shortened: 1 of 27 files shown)


phlptp:
This looks OK to me.  I am going to give @henryiii  a few days to comment, then will merge.

phlptp (approved):

henryiii (approved):
Not very familiar with IWYU pragmas, but seems fine.

henryiii:
Oh, it is valid though to include individual headers (it's in the readme). It sounds like this will just point users to always including the conglomerate header? Shouldn't these (Config and formatter) have `// IWYU pragma: keep` instead?

gostefan:
I looked through the Readme and didn't find any references to individually including headers except for the utilities (which I promptly forgot to correctly handle in my changes).

Is it ok to include the "regular" headers (e.g. `Argv.hpp`) individually?
I'll adapt the PR accordingly.

PR #1011, opened 2024-02-21
Fix internal header include paths.

samhocevar:
The extra leading `"CLI/"` part of include directives prevents the inclusion of `CLI.hpp` from a relative directory without an extra `-I` or `/I` compiler directive, and makes it harder to make CLI11 part of a larger codebase.

This is a regression of #475.

### DIFF
--- a/include/CLI/Argv.hpp
+++ b/include/CLI/Argv.hpp
@@ -11,7 +11,7 @@
 #include <vector>
 // [CLI11:public_includes:end]
 
-#include <CLI/Macros.hpp>
+#include "Macros.hpp"
 
 namespace CLI {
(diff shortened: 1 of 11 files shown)


henryiii:
Is there any way to add this to tests / checks to avoid future regressions?

henryiii (approved):

samhocevar:
I think a simple enough test would be, for a specific target:
 - remove `${CMAKE_SOURCE_DIR}/include` from `target_include_directories`
 - replace `#include "CLI/CLI.hpp"` with `#include "../include/CLI/CLI.hpp"`
 - just check that everything compiles…

This makes assumptions about the source tree hierarchy, so I’m afraid it’s not a perfect solution.

PR #1012, opened 2024-02-26
Restore non-arch dependant path for the pkgconfig file

j-rivero:
Originally setup at https://github.com/CLIUtils/CLI11/pull/635 but reverted in https://github.com/CLIUtils/CLI11/pull/916 without any specific reason that I can find.


henryiii:
@phlptp, looks like you moved it, was there a reason?

phlptp:
I was going to by the pkg-config [docs]( https://linux.die.net/man/1/pkg-config#:~:text=By%20default%2C%20pkg%2Dconfig%20looks,by%20the%20PKG_CONFIG_PATH%20environment%20variable.)  Which has the default location in the lib directory.  I don't use pkg-config much so if we are sure the new location is cross platform and the tests pass,  other than making a comment in the cmake file I have no issue changing it.

j-rivero:
> the new location is cross platform and the tests pass, other than making a comment in the cmake file I have no issue changing it.

Using [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) as the wrapper to install files, we have: 

> DATADIR
>    read-only architecture-independent data (DATAROOTDIR)

Which on my experience has worked well to define the paths of the pkgconfig files for all kind of projects. Let me know if you see any use case that is failing due to this change.

phlptp:
On some projects the datadir is used to store model files and other "data" necessary for program operation or examples.   Mixing library information into that seems a bit counterintuitive to me,  but if that is what you regularly use, and pkgconfig works normally with that I would suggest adding a comment in the cmake file, and go forward with it.

cmake/CLI11GeneratePkgConfig.cmake
```diff
@@ -4,4 +4,4 @@ else()
   configure_file("cmake/CLI11.pc.in" "CLI11.pc" @ONLY)
 endif()
 
-install(FILES "${PROJECT_BINARY_DIR}/CLI11.pc" DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
+install(FILES "${PROJECT_BINARY_DIR}/CLI11.pc" DESTINATION "${CMAKE_INSTALL_DATADIR}/pkgconfig")
```
dg0yt, line 7: This change is wrong at least when using cli11 in compiled mode. The binary artifact (`libCLI11`) is very much arch dependant.
Noticed while updating the vcpkg package.

PR #1024, opened 2024-03-19
Meson: fixes, cleanups, and modernizations

dcbaker:
This series cleans up and modernizes the Meson code a bit:

 - It adds the license SPDX name to the proper field
 - Sets a minimum version, the version is fairly old, but without setting a version then Meson will never warn about issues, including potentially important ones like "this happens to work in some cases but wasn't intended and doesn't work in all cases"
 - uses more modern Meson features to make things easier for consumers
 - Fixes the tests using the ensure_utf8 helpers by ensuring that they're actually built, and placing them in the correct path.
 - Adds the Meson test to the CI

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -166,6 +166,9 @@ jobs:
       - name: Build
         run: meson compile -C build-meson
 
+      - name: Test
+        run: meson test -C build-meson
+
   install:
     name: install tests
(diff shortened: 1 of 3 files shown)


henryiii:
It would be nice to have a bit of documentation on the modern way to use it, but this seems like a nice cleanup!

henryiii (approved):

PR #1025, opened 2024-03-20
More Meson work

dcbaker:
This is follow up work to my previous series. I've tried to make the Meson build mirror the CMake build more closely. I've also made an attempt at adding some instructions to the documents on using Meson.

### DIFF
--- a/include/CLI/meson.build
+++ b/include/CLI/meson.build
@@ -37,10 +37,18 @@ cli11_impl_headers = files(
 
 single_main_file = files('../../CLI11.hpp.in')
 
-single_header = custom_target(
-  'CLI11.hpp',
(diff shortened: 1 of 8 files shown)


include/CLI/meson.build
henryiii, line 1: Pretty sure this is the wrong directory - I think you meant `include/meson.build`, since `CLI11.hpp` sits inside `include` in the current CMake build.

Even so, I'm still not very fond of the idea of playing anything besides headers inside `/include`. It's valid to `-I` include it, and it seems weird to have meson files there too. Maybe it would be better to make a new dir, `single-include`, and place meson.build and the single file output there. We could also move the CMake output there, and even put the CMake single header file code there too.

henryiii:
I've moved the single file to it's own folder, so now it should be possible to avoid the meson.build file inside the include dir.

PR #1029, opened 2024-04-03
Improved clarity in RequiredError

sifferman:
Hello. Thank you for this tool.

I found a few small typos in the RequiredError class.

This is an error message I received:

```
Requires at most 2 options be used and 3were given from [filename,--get-size,--get-size2]
```
This PR changes the error message to
```
Requires at most 2 options be used but 3 were given from [filename,--get-size,--get-size2]
```

Happy to make changes as needed!

## Source

```cpp
    std::string filename;
    bool get_size = false;

    CLI::App app{"Program to read a file specified as command-line argument"};

    app.add_option("filename", filename, "File to read")->required();
    app.add_flag("--get-size", get_size, "Print the size of the file");
    app.add_flag("--get-size2", get_size, "Print the size of the file2");

    app.require_option(1, 2); // Enforce only one flag can be input

    CLI11_PARSE(app, argc, argv);
```

## Further Idea

Also, another idea I had was to remove required options from this error message. In this example, "filename" is required, so a better error message would be
```
Requires at most 1 options be used but 2 were given from [--get-size,--get-size2]
```
I'm happy to look into this if you feel it would be valuable

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -99,7 +99,7 @@ add_test(NAME option_groups_missing COMMAND option_groups)
 set_property(TEST option_groups_missing PROPERTY PASS_REGULAR_EXPRESSION "Exactly 1 option from"
                                                  "is required")
 add_test(NAME option_groups_extra COMMAND option_groups --csv --binary)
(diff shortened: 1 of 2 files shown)


henryiii:
Sure, sounds like a reasonable further cleanup; thanks for this!

PR #1033, opened 2024-04-16
Support building with Bazel

calebzulawski:
Adds support for building with Bazel.  If merged, I can push this to https://registry.bazel.build/ when a new release is cut :)


tests/HelpersTest.cpp
```diff
@@ -522,7 +522,11 @@ TEST_CASE("Validators: FileIsDir", "[helpers]") {
 }
 
 TEST_CASE("Validators: DirectoryExists", "[helpers]") {
+#ifdef BAZEL
```
henryiii: Is it possible to avoid having different paths here for Bazel? Bazel is running from the parent dir, while the others are running from the tests dir?

henryiii: Maybe we could make a `tests` directory in tests, then put a .gitkeep in it? Then running in either directory would be supported. It seems bazel cannot set the working directory when running tests.

calebzulawski: That sounds like an elegant solution to me. I'm not quite sure why the CMake builds are failing now (I'm certainly no CMake expert)

calebzulawski: I made a small change to your last commit that seems to fix it

henryiii: Great, thanks!

henryiii:
Thanks!

henryiii:
Do you think you could add a little section on Bazel to the docs, similar to https://github.com/CLIUtils/CLI11/pull/1025/files ?

calebzulawski:
Yep, no problem. Do you think you'll be tagging a new patch release sometime soon? Once you do that, I'll add this project to the bazel central registry, and update the docs with a reference to how to pull it from there.

Issue #1034, opened 2024-04-19
Question: Option requires one of two other options


ptc-tgamper:
Hi,

I am in the situation where I have options `-D`, `-E`, and `-F`. What I need is that if `-D` is provided, then exactly one of `-E` and `-F` has to be specified, too.
Providing `-D` with `-E` is valid. Providing `-D` with `-F` is valid.
Providing only `-D` is invalid. Providing `-D`, `-E`, and `-F` is invalid. Providing only `-E` is invalid. Providing only `-F` is invalid.

I haven't found a way to do this yet.

Otherwise I am really happy with CLI11 and the functionality it provides. Kudos to the authors.

Thanks,
Thomas

phlptp:
I think the easiest way would be to put the D option in 1 option group and the E and F option in another group.  Make the E and F options exclusive or use the option_group required count so the min and max required is 1.   Then for the D group use it to trigger the E and F option group on.   using `CLI::TriggerOn` helper.      To make the help more useful you may have to catch some specific errors related to using the E and F without D to make the help message more useful

ptc-tgamper:
I had to do both. Mark E and F to exclude each other, and to tell their group that only a single option is allowed. Then I am getting reasonable error messages.

Still this leaves me a bit unhappy, because I have now groups in the help message, where I would prefer not to have them. Especially because I have options that belong together, but are in different groups.

phlptp:
in the main branch, if you add an option group with a name that begins with a '+'  it should now display as part of the parent group/app.

PR #1039, opened 2024-05-10
Group merge

phlptp:
This PR adds a mechanism to hide groups but have their options visible as part of the parent.
This works for option group names starting with a '+'

for example 
```
 CLI::App app;

    bool flag = false;
    std::optional<bool> optional_flag = std::nullopt;

    app.add_option("--tester");
    auto *m1=app.add_option_group("+tester");

    m1->add_option("--flag", flag, "description");
    m1->add_option("--optional_flag", optional_flag, "description");

    CLI11_PARSE(app,argc, argv);
```
will produce help as 
```txt
Options:
  -h,--help                   Print this help message and exit
  --tester
  --flag BOOLEAN              description
  --optional_flag BOOLEAN     description
```

instead of 
```
Options:
  -h,--help                   Print this help message and exit
  --tester
[Option Group: tester]
  Options:
    --flag BOOLEAN              description
    --optional_flag BOOLEAN     description

```

Fixes issue #1034 and a few other past issues or questions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- add mechanism to allow option groups to be hidden and all options be considered part of the parent for help display
+
 ## Version 2.4: Unicode and TOML support
 
 This version adds Unicode support, support for TOML standard including multiline
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## Unreleased
 
-- add mechanism to allow option groups to be hidden and all options be considered part of the parent for help display
+- add mechanism to allow option groups to be hidden and all options be
+  considered part of the parent for help display
 
 ## Version 2.4: Unicode and TOML support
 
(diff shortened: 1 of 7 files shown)


phlptp:
@henryiii unless I hear some objections I am going to merge this on Monday.

henryiii:
Sounds good. Let me know if there’s anything specific to review. At PyCon currently.

phlptp:
Mostly if you have any concerns about the notation itself and having slightly different behavior depending on the option group name.    I know there have been several requests for this capability and I am a little reticent to add more modifiers and options, so this seemed like a nice solution.

henryiii:
> Especially because I have options that belong together, but are in different groups.

That wouldn't be covered by this?

phlptp:
> > Especially because I have options that belong together, but are in different groups.
> 
> That wouldn't be covered by this?

It should be.   The way the mechanic works is that the option would get associated with a parent group, that could be the main app or a sub_group.   So you could tie any number of sub_groups together for help display as part of the main app or in another sub group.

Issue #1048, opened 2024-05-30
New error after updating to 2.4.2: add_option throws error when adding option "--h"


DSMitten:
Existing built-in options include `-h` and `--help`

I have code which adds the following:
```cpp
parser->add_option("--h", myArgs.headerPath, "Path to header file");
```

The intent is that you would call `myprogram.exe <other-options> --h <path-to-header>`

This worked in 2.4.0, but breaks in 2.4.2.  After updating to the latest vcpkg port, I get the following error:
`error: added option matched existing option name: h is already added`

I would not expect the parser to treat `-h` and `--h` the same, since they are different switches.
After commenting out my line, when running `myprogram.exe --h`, it invokes help, so it looks like the parser isn't distinguishing between single and double dashes at all.

Testing this out with other letters, if I try `add_option` for both `-t` and `--t`, it errors out, also saying that it is a duplicate argument name, but when parsing it doesn't treat them as the same.

phlptp:
One reason for not allowing that is that when parsing config files they would conflict.   We could possibly allow it if the option was not able to present in a config file which is the case for the help option.   This was changed in a recent release to fix the ambiguity in config file writing and reading,  naming positional the same as a short or long option will also cause an error.

henryiii:
I’m not sure it makes sense to put short options in a config file.

phlptp:
If an option only has a short option defined, it still needs to be able to be defined in a config file.   If it defines both, the long option name is used for generation.  Also it would be rather confusing if a short option name used in a config file didn't match up with the option that defined the short name.

DSMitten:
This worked in 2.4.0. There are no notes in 2.4.2 indicating that it includes a breaking change to prevent the use of single and double-dash options with the same name. Was this an intentional breaking change?

DSMitten:
From a library consumer's standpoint, it was confusing to treat `-t` and `--t` as the same when validating the parser but not when actually parsing.

If you want to avoid conflicts, it might make sense to just ban the use of two-dash single-letter arguments. At least then the errors will be consistent and make sense.  If I print out a usage statement and it shows that there is a `-t` option but no `--t` option, it won't make sense that I can't add the `--t` option.

It would just be nice for the old behavior to still work.

phlptp:
[...]

PR #1049, opened 2024-05-31
conflicting option names

phlptp:
Take the configurability of an option into account when determining ambiguous names and conflicts.


phlptp:
This will fix #1047 
This will resolve #1048

phlptp:
I made the changes in  PR #1049.   There were a few edge cases to work through that the fuzzer caught.  
There is probably a bit more to do to take into account whether the subcommand or app itself is not configurable.   That might require some additional thought.  But for now this will resolve the `--h` issue and any other case where one of the options is not configurable, and hence wouldn't present an ambiguity in the config files.

.pre-commit-config.yaml
```diff
@@ -87,7 +87,7 @@ repos:
         exclude: .pre-commit-config.yaml
 
   - repo: https://github.com/codespell-project/codespell
-    rev: v2.2.6
+    rev: v2.3.0
     hooks:
       - id: codespell
-        args: ["-L", "atleast,ans,doub,inout"]
+        args: ["-L", "atleast,ans,doub,inout,AtMost"]
```
henryiii, line 93: Oh, did this bug get fixed in 2.3.0? It used to be you had to put all lower case versions here.

tests/fuzzFail/fuzz_app_file_fail40
```diff
@@ -0,0 +1 @@
+     config                      ' '
```
henryiii, line 1: Haven't followed the fuzzer that closely, what does this do?

henryiii (approved):

tests/fuzzFail/fuzz_app_file_fail40
phlptp: Right now the fuzzer has a large number of options,  and some limited mechanics to generate new ones.  It processes the options making sure there is nothing that generates a seg fault or other bad condition.   Then generates a config file from the results and reads that config file back in.      This led to all kinds of inconsistencies that got fixed a few months ago.   At some point I will make it so it checks to make sure the values read from the config file are actually the same as the original values, then I might consider submitting it to [OSS-Fuzz](https://github.com/google/oss-fuzz).   The current test just runs a quick_fuzz ~6 minutes and can catch some issues especially when changing some internal mechanics or other things.    I will keep adding to it as I have time.  

This particular test caught an issue with creating a positional option named config, which was newly allowed.  This then triggered an error as it was trying to match with the config option which was not configurable, so had to make it so it checked for other options if some were not configurable.

.pre-commit-config.yaml
phlptp: Didn't know there was a bug, but this resolved the pre-commit issue

PR #1066, opened 2024-08-22
Static linking with MSVC and PRECOMPILED=ON

timkruse:
Solves the following issue:

Using CLI11 with the option `CLI11_PRECOMPILED=ON` results in the following linking error (Windows10 + MSVC set up to link statically)

`CLI11.lib(Precompile.cpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MT_StaticRelease' in srcfile.cpp.obj`

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,7 +4,8 @@ if(CLI11_PRECOMPILED)
   add_library(CLI11 STATIC ${CLI11_headers} ${CLI11_library_headers} ${CLI11_impl_headers}
                            ${CLI11_precompile_sources})
   if(MSVC)
-    set_property(TARGET CLI11 PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    set_property(TARGET CLI11 PROPERTY MSVC_RUNTIME_LIBRARY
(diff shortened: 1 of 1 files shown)


phlptp:
what would a CI build look like that would have caught this error?

timkruse:
Changing my application that includes CLI11 back to dynamic linking brings back the linking error from above but with swapped  `MD_DynamicRelease` and `MT_StaticRelease`. So this PR is pointless, sorry. 

For documentation purposes:
I want a statically linked executable on Windows with MSVC and use `CLI11_PRECOMPILED=ON`. Therefore, I have to put 

`set_property(TARGET myApp PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")`

in my `CMakeLists.txt`. That line alone raises the linking error shown above. To fix it CLI11 needs to be compiled with the same `MSVC_RUNTIME_LIBRARY` flags, but it can be placed in my `CMakeLists.txt`, too:

`set_property(TARGET CLI11 PROPERTY MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")`

phlptp:
I have vague recollections of having something like this in another library, I will poke around and see if anything should be done differently to make it a little smoother.

Issue #1067, opened 2024-09-03
Specifying container of containers in configuration file


pauldmccarthy:
Howdy,

I am trying to build a CLI application with an option that both expects multiple values, and can be given multiple times, e.g.:

```
myapp --option 1 2 3 --option 4 5 6
```

I want the values to be aggregated into a vector of vectors, so am using the following
code:

```c++
std::vector<std::vector<int>> option;
app.add_option(
  "--option",
  option,
  "Vector of vectors");
```

This works just fine when the `--option` values are specified at the command-line - I end up with a vector containing two vectors, which contain `1 2 3` and `4 5 6` respectively.

However there appears to be an issue when I try to specify the options in a configuration file, enabled via `app.set_config("--config")`. If I create a `config.ini` file with the following:

```
option = 1 2 3
option = 4 5 6
```

and run `myapp --config config.ini`, I end up with a vector which contains just one vector, containing `1 2 3 4 5 6`.

Here is a minimal example which reproduces the issue:

```c++
#include <iostream>
#include <vector>

#include "CLI/CLI.hpp"

int main(int argc, char * argv[]) {

  CLI::App app{"MyApp"};

  std::vector<std::vector<int>> option;
  app.add_option(
    "--option",
    option,
    "Container of containers");

  app.set_config("--config");

  CLI11_PARSE(app, argc, argv);

  std::cout << "CLI11 version "
            << CLI11_VERSION_MAJOR << "."
            << CLI11_VERSION_MINOR << "."
            << CLI11_VERSION_PATCH << std::endl;

  std::cout << "Value for --option" << std::endl;
  for (auto &vec : option) {
    for (auto &val : vec) {
      std::cout << val << " ";
    }
    std::cout << std::endl;
  }
}
```

When specifying the values at the command-line:

```
./myapp --option 1 2 3 --option 4 5 6
```

I get the expected result:

```
CLI11 version 2.4.2
Value for --option
1 2 3
4 5 6
```

However, when I use a config file:

```
echo "option = 1 2 3"  > config.ini
echo "option = 4 5 6" >> config.ini
./myapp --config config.ini
```

I get:

```
CLI11 version 2.4.2
Value for --option
1 2 3 4 5 6
```

Thanks!

phlptp:
Are you subvectors all the same length?  If so you can specify that directly and it should split them via typesize.

You can also try `->inject_separator()` on the option.   I don't have time to build a test case for that right now, but that was designed to handle some similar situations.

[...]

PR #1069, opened 2024-09-06
Vector input to config file

phlptp:
From #1067, there is a bit of ambiguity in the handling of config file with vector input and multiple consecutive parameters.   For example

```toml
option1=[3,4,5]
option1=[4,5,6]
```

Currently this is handled as if it were 
```toml
option1=[3,4,5,4,5,6]
```
But this could be confusing in the case where the input was referring to a vector of vectors.  
This PR adds a separator in the sequence to separate the vector so they are two vectors of 3 elements each.  
Will need to verify if this change has other side effects.   It is a pretty unusual situation.


phlptp:
@henryiii I will merge this Tuesday, unless you have some issues with it.

pauldmccarthy:
@phlptp This appears to have resolved the issue, thanks very much for looking into it!

_edit_ my comment was a bit premature, as I am still getting inconsistent behaviour when I adjust my example a little bit - will continue the discussion at #1067

PR #1073, opened 2024-10-07
Subcommand fallthrough

### LINKED ISSUE #1022: Positional is treated as a sub-command
CLI11: Version 2.4.1

I have this setup:

```c++
int main() {
    CLI::App cli{};
    cli.set_help_flag();
    cli.add_subcommand("foo", "Does foo")->parse_complete_callback([&](){std::cout << "=== Foo doing foo" << std::endl;});
    cli.add_subcommand("bar", "Does bar")->parse_complete_callback([&](){std::cout << "=== Bar doing bar" << std::endl;});
    auto *help_cmd = cli.add_subcommand("help", "Does help");
    std::string help_for_command;
    auto *help_opt = help_cmd->add_option("command", help_for_command, "Command to print help for")->expected(0, 1);
    help_cmd->parse_complete_callback([&]() {
            if (*help_opt) {
                std::cout << "Print help for " << help_for_command << std::endl;
            } else {
                CLI::Formatter fmt{};
                std::cout << fmt.make_help(&cli, "", CLI::AppFormatMode::Normal) << std::endl;
            }
    });

    std::cout << "=> " << std::flush;
    for (std::string line; std::getline(std::cin, line);) {
        try {
            cli.parse(line);
        } catch (const CLI::ParseError &e) {
            std::cout << e.what() << std::endl;
            std::cout << cli.help();
        }
        std::cout << "=> " << std::flush;
    }
}
```

Given the above setup, here are some behaviors that I expect:
```
Scenario 1
When help is called without arguments
Then the help string of all commands is printed

Scenario 2
When help is called with one argument
And   the argument is a registered (sub)command
Then the help string of this command is printed

Scenario 3
When help is called with one argument
And   the argument is not a registered (sub)command
Then the help string of all commands is printed
```

However, I get the below output for each scenario:

Scenario 1 🟢 
```shell
=>help
 help
Usage: [SUBCOMMAND]

Subcommands:
  foo                         Does foo
  bar                         Does bar
  help                        Does help

=>
```

Scenario 2 🔴 
Not only the positional is not recognised, it is executed as a sucommand.
```shell
=>help foo
 help foo
Usage: [SUBCOMMAND]

Subcommands:
  foo                         Does foo
  bar                         Does bar
  help                        Does help

=== Foo doing foo
=>
```

```shell
=>help help
 help help
Usage: [SUBCOMMAND]

Subcommands:
  foo                         Does foo
  bar                         Does bar
  help                        Does help

Usage: [SUBCOMMAND]

Subcommands:
  foo                         Does foo
  bar                         Does bar
  help                        Does help

=>
```

Scenario 3 🔴 
```shell
=>help other
 help other
Usage: [SUBCOMMAND]

Subcommands:
  foo                         Does foo
  bar                         Does bar
  help                        Does help

The following argument was not expected: other
Does help
Usage: help [command]

Positionals:
  command TEXT                Command to print help for
```

I've tried using `fallthrough(false)` for `help` command, but with no luck.
How could I implement my use case?

phlptp:
Add modifier for subcommands to restrict subcommands falling through to parent.  
This will resolve #1022

### DIFF
--- a/.github/actions/quick_cmake/action.yml
+++ b/.github/actions/quick_cmake/action.yml
@@ -13,7 +13,7 @@ runs:
   using: composite
   steps:
     - name: CMake ${{ inputs.cmake-version }}
-      uses: jwlawson/actions-setup-cmake@v1.14
+      uses: jwlawson/actions-setup-cmake@v2.0.2
       with:
         cmake-version: "${{ inputs.cmake-version }}"
     - run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -292,10 +292,10 @@ jobs:
           cmake-version: "3.13"
         if: success() || failure()
 
-      - name: Check CMake 3.14
+      - name: Check CMake 3.14.7
         uses: ./.github/actions/quick_cmake
         with:
-          cmake-version: "3.14"
+          cmake-version: "3.14.7"
           args: -DCLI11_SANITIZERS=ON -DCLI11_BUILD_EXAMPLES_JSON=ON
         if: success() || failure()
 
@@ -387,6 +387,6 @@ jobs:
       - name: Check CMake 3.28 (full)
         uses: ./.github/actions/quick_cmake
         with:
-          cmake-version: "3.28"
+          cmake-version: "3.28.X"
           args: -DCLI11_SANITIZERS=ON -DCLI11_BUILD_EXAMPLES_JSON=ON
         if: success() || failure()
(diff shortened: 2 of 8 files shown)


phlptp:
This also fixes some of the cmake build jobs that were failing from a recent update in that github action.

phlptp:
Unless I hear some objections/suggestions I will merge this tomorrow morning.

henryiii (approved):

PR #1076, opened 2024-10-12
Use pymod.find_installation() to find python in meson.build

nshaheed:
Since Monterey 12.3, macOS removed python2.7 from it's base installation. This means that `python` is not longer a command by default (just `python3`), and the build for CLI11 fails. This PR moves to using `pymod.find_installation()` to handle this in a more general way. See https://mesonbuild.com/Python-module.html

### DIFF
--- a/single-include/meson.build
+++ b/single-include/meson.build
@@ -2,7 +2,8 @@
 # meson.build here when generating the single file header so that it is placced
 # in the correct location.
 
-prog_python = find_program('python')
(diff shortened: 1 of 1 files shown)


phlptp:
I have no experience with meson, but seems reasonable.  
@henryiii  unless you have objections I will approve and merge on Monday.

henryiii:
This looks correct to me.

PR #1077, opened 2024-10-14
Custom fuzz options

phlptp:
add mechanics for the fuzzer to add custom options.


phlptp:
The aim is to eventually get cli11 running in ossfuzz.  This is another step in that direction.

Issue #474, opened 2020-06-10
How strong is the dislike for non-standard syntax?


samhocevar:
Hello. I am not (yet) a CLI11 user but I am growing frustrated with the bugs in [clipp](https://github.com/muellan/clipp) and CLI11 seems like a very mature alternative. The README states that this is not supported:

> Non-standard variations on syntax, like `-long` options. This is non-standard and should be avoided, so that is enforced by this library.

I do avoid these, too. However I am writing a replacement for some non-free software which uses such flags. If I ever want to provide a drop-in replacement that does not break existing user scripts, I am a bit stuck for reasons that are not really technical.

What are my options? Should I just fork? Or would you accept a PR under certain conditions, like an opt-in macro that users should explicitly provide?

phlptp:
See #462 for some other discussion on this.  

I think it would be fairly challenging to add support without breaking a bunch of other stuff, but as long as the first character of the `-xxxx` option doesn't match another single character option the option gets placed in an extra arguments location.  So then after regular processing you can have a second step which deals with unhandled options more directly or have a `final_callback` which deals with them like the example given in #462.

samhocevar:
Thanks for the suggestion. Unfortunately I do need to support conflicting stuff such as `-o something` and `-out something_else`. Also using `final_callback` would prevent me from using all the nice type-safe argument parsing.

I guess I will try to implement the feature and keep it updated in a custom branch.

**Edit**: by the way, I am now a CLI11 user and already very satisfied.

samhocevar:
So, I wrote a proof of concept in ca183479b7c11f2ec59fcade9efc95e302ef7b96. My strategy is to store these special flags in the short options, but try to parse them as if they were long options first. If option lookup fails, try again with `Classifier::SHORT`.

This allows to define options `-o`, `-p` and `-poop` and they all work as expected; `-pop` will behave like `-p -o -p` but `-poop` will be something else.

This is not very elegant and not complete, but it suits my needs and is non-intrusive enough that I can maintain the patch without much effort.

henryiii:
I wouldn't be against adding it if it can be added safely, and if there is a little warning about using it in the docs, saying something like this should be for legacy support only or something like that. I was targeting being able to be used in ROOT at some point, and it does have 1-2 legacy "medium" options. It shouldn't match long options. I think defining "-medium" should be the only way to trigger a medium option match.

[...]

PR #1078, opened 2024-10-19
Allow non standard option names like `-option`

phlptp:
This has been bounced around for a couple years now 

#474 and a few others have expressed desire to work with non-standard option names.  We have been somewhat resistant to that but I think it can be done now.   This PR adds a modifier  `allow_non_standard_option_names()`  It is purposely long, it is purposely off by default.   But what it does is allow option names with a single `-`  to act like a short option name.     With this modifier enabled no single letter short option names are allowed to start with the same letter as a non-standard names.  For example `-s` and `-single`  would not be allowed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,10 @@ installation fuss.
 There are some other possible "features" that are intentionally not supported by
 this library:
 
-- Non-standard variations on syntax, like `-long` options are not supported by default.  🚧 There is a new option to allow this syntax but in general our opinion is these should be avoided except for backwards compatibility with existing applications.
+- Non-standard variations on syntax, like `-long` options are not supported by
+  default. 🚧 There is a new option to allow this syntax but in general our
+  opinion is these should be avoided except for backwards compatibility with
+  existing applications.
 - Completion of partial options, such as Python's `argparse` supplies for
   incomplete arguments. It's better not to guess. Most third party command line
   parsers for python actually reimplement command line parsing rather than using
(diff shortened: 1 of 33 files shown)


phlptp:
@henryiii  Been thinking about this one for a while.   played with a bunch of different options for this, still not sure how I feel about it, but I do see the point that if CLI11 is to be a replacement in some legacy applications, we do need some mechanic to support single dash long options.  This way seems like the least intrusive and least complex on the option side.    Not going to merge this until you approve though.

azure-pipelines.yml
```diff
@@ -22,7 +22,7 @@ jobs:
   - job: CppLint
     pool:
       vmImage: "ubuntu-latest"
-    container: sharaku/cpplint:latest
+    container: helics/buildenv:cpplint
```
phlptp, line 25: the container we had been using hasn't been updated in 5 years.  Figured we needed something a little newer.

README.md
```diff
@@ -158,8 +158,10 @@ installation fuss.
 There are some other possible "features" that are intentionally not supported by
 this library:
 
-- Non-standard variations on syntax, like `-long` options. This is non-standard
-  and should be avoided, so that is enforced by this library.
+- Non-standard variations on syntax, like `-long` options are not supported by
+  default. 🚧 There is a new option to allow this syntax but in general our
+  opinion is these should be avoided except for backwards compatibility with
+  existing applications.
```
henryiii: This could be removed, as it's not a limitation anymore.

examples/custom_parse.cpp
```diff
@@ -9,6 +9,7 @@
 #include <CLI/CLI.hpp>
 #include <iostream>
 #include <sstream>
+#include <string>
```
henryiii, line 12: Why are these adding string with no changes? IWYU?

tests/BoostOptionTypeTest.cpp
```diff
@@ -13,17 +13,17 @@
 #include <boost/container/static_vector.hpp>
 #include <boost/container/vector.hpp>
 #include <string>
+#include <tuple>
+#include <utility>
 #include <vector>
 
-using namespace boost::container;
```
henryiii, line 18: :heart: (change not related to this PR's topic, but like the change)

README.md
```diff
@@ -904,6 +906,14 @@ option_groups. These are:
   the form of `/s /long /file:file_name.ext` This option does not change how
   options are specified in the `add_option` calls or the ability to process
   options in the form of `-s --long --file=file_name.ext`.
+- `.allow_non_standard_option_names()`:🚧 Allow specification of single `-` long
+  form option names. This is not recommended but is available to enable
+  reworking of existing interfaces. If this modifier is enabled on an app or
+  subcommand, options or flags can be specified like normal but instead of
+  throwing an exception long form single dash option names will be allowed. It
```
```suggestion
  throwing an exception, long form single dash option names will be allowed. It
```

README.md
```diff
@@ -904,6 +906,14 @@ option_groups. These are:
...
+  is not allowed to have a single character short option starting with the same
+  character as a single dash long form name for example `-s` and `-single` are
```
```suggestion
  character as a single dash long form name; for example, `-s` and `-single` are
```

tests/BoostOptionTypeTest.cpp
phlptp: IWYU was confusing slist from boost with the one from gcc, obviously couldn't include that one, so that just triggered getting rid of the boost container namespace and using it more directly with the full name of the types

README.md
phlptp:
```suggestion

```

examples/custom_parse.cpp
phlptp: the update to cpplint includes a lot more IWYU checks so this came up.

henryiii (approved):

PR #1081, opened 2024-11-05
pair/tuple defaults

### LINKED ISSUE #711: default value for std::pair
Following code does not work.

```
    // (...)
    float2 pre_translate;
    app.add_option<float2, std::pair<float, float>>(
        "--pre_translate", pre_translate, "translate before scaling (ie. to center the svg file)")
        ->default_val(std::pair<float, float>{0,0});
    // (...)
```

It compiles, but when ran fails with:

```
terminate called after throwing an instance of 'CLI::ArgumentMismatch'
  what():  --pre_translate: At least 2 required but received 1
zsh: IOT instruction (core dumped)  ./a.out --pre_translate 0 0
```

`float2` is interestingly not constructable from `std::pair` but it works nevertheless.
It works correctly if I comment out the `default_val`.

I'm pretty sure I'm doing something wrong. What's the correct method of specifying the default value?

phlptp:
add capability to accept pair/tuple default values and a little cleaner parsing in some cases

Fixes #711

### DIFF
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -356,35 +356,21 @@ std::string to_string(T &&value) {
     return stream.str();
 }
 
-/// Convert a tuple like object to a string
-
-/// forward declarations for tuple_value_strings
-template <typename T, std::size_t I>
-inline typename std::enable_if<I == type_count_base<T>::value, std::string>::type tuple_value_string(T && /*value*/);
-
-/// Recursively generate the tuple value string
-template <typename T, std::size_t I>
-inline typename std::enable_if<(I < type_count_base<T>::value), std::string>::type tuple_value_string(T &&value);
+//additional forward declarations
 
 /// Print tuple value string for tuples of size ==1
 template <typename T,
-          enable_if_t<!std::is_convertible<std::string, T>::value && !std::is_constructible<std::string, T>::value &&
-                          !is_ostreamable<T>::value && is_tuple_like<T>::value && type_count_base<T>::value == 1,
-                      detail::enabler> = detail::dummy>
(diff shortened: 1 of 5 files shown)


include/CLI/TypeTools.hpp
```diff
@@ -307,6 +307,30 @@ template <typename S> class is_tuple_like {
     static constexpr bool value = decltype(test<S>(0))::value;
 };
 
+/// This will only trigger for actual void type
```
phlptp, line 310: code moved to get thing in the right order

include/CLI/impl/Option_inl.hpp
```diff
@@ -662,11 +662,11 @@ CLI11_INLINE std::string Option::_validate(std::string &result, int index) const
 
 CLI11_INLINE int Option::_add_result(std::string &&result, std::vector<std::string> &res) const {
     int result_count = 0;
-    if(allow_extra_args_ && !result.empty() && result.front() == '[' &&
+    if((allow_extra_args_ || get_expected_max() > 1) && !result.empty() && result.front() == '[' &&
        result.back() == ']') {  // this is now a vector string likely from the default or user entry
         result.pop_back();
 
-        for(auto &var : CLI::detail::split(result.substr(1), ',')) {
+        for(auto &var : CLI::detail::split_up(result.substr(1), ',')) {
```
phlptp, line 669: wasn't handling sub vectors properly

phlptp:
@henryiii   FYI,  no api changes here,  Mainly getting the default_val on parity with the types the options support themselves.   Also cleaned up a few methods.   Was a bug that came up a while ago and hadn't got around to fixing until now.    I will merge in a few days

PR #1084, opened 2024-11-09
feat: increase min CMake to 3.10

### LINKED ISSUE #1082: cmake version deprecation warning
I am using cmake 3.30

and just started getting this:

```
CMake Deprecation Warning at ext/CLI11/CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

henryiii:
Close https://github.com/CLIUtils/CLI11/issues/1082.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,16 +1,6 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.10...3.31)
 # Note: this is a header only library. If you have an older CMake than 3.5,
 # just add the CLI11/include directory and that's all you need to do.
-
-# Make sure users don't get warnings on a tested (3.5 to 3.30) version
-# of CMake. For most of the policies, the new version is better (hence the change).
-# We don't use the 3.5...3.30 syntax because of a bug in an older MSVC's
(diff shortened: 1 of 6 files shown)


phlptp (approved):

CMakeLists.txt
```diff
@@ -1,17 +1,7 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.10...3.31)
 # Note: this is a header only library. If you have an older CMake than 3.5,
 # just add the CLI11/include directory and that's all you need to do.
 
-# Make sure users don't get warnings on a tested (3.5 to 3.30) version
-# of CMake. For most of the policies, the new version is better (hence the change).
-# We don't use the 3.5...3.30 syntax because of a bug in an older MSVC's
-# built-in and modified CMake 3.11
```
phlptp, line 8: Do we care about this bug anymore?

henryiii: I don't think so. Not sure we even support this old version of MSVC anymore.

PR #1101, opened 2024-12-06
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling) (which is an evolution of the script I used ages ago when I first made a PR here...).

The misspellings have been reported at https://github.com/jsoref/CLI11/actions/runs/12194174338#summary-34017587518

The action reports that the changes in this PR would make it happy: https://github.com/jsoref/CLI11/actions/runs/12194174680#summary-34017588281

### DIFF
--- a/README.md
+++ b/README.md
@@ -1665,7 +1665,7 @@ brief description of each is included here
 
 To contribute, open an [issue][github issues] or [pull
 request][github pull requests] on GitHub, or ask a question on [gitter][]. There
-is also a short note to contributors [here](./.github/CONTRIBUTING.md). This
+is also a [short note to contributors](./.github/CONTRIBUTING.md). This
 readme roughly follows the [Standard Readme Style][] and includes a mention of
 almost every feature of the library. More complex features are documented in
 more detail in the [CLI11 tutorial GitBook][gitbook].
--- a/README.md
+++ b/README.md
@@ -1254,7 +1254,7 @@ formatter. An example of a TOML file:
 
 ```toml
 # Comments are supported, using a #
-# The default section is [default], case insensitive
+# The default section is [default], case-insensitive
 
 value = 1
 value2 = 123_456 # a string with separators
@@ -1275,7 +1275,7 @@ or equivalently in INI format
 
 ```ini
 ; Comments are supported, using a ;
-; The default section is [default], case insensitive
+; The default section is [default], case-insensitive
 
 value = 1
 str = "A string"
@@ -1291,7 +1291,7 @@ sub.subcommand = true
 Spaces before and after the name and argument are ignored. Multiple arguments
 are separated by spaces. One set of quotes will be removed, preserving spaces
 (the same way the command line works). Boolean options can be `true`, `on`, `1`,
-`yes`, `enable`; or `false`, `off`, `0`, `no`, `disable` (case insensitive).
+`yes`, `enable`; or `false`, `off`, `0`, `no`, `disable` (case-insensitive).
 Sections (and `.` separated names) are treated as subcommands (note: this does
 not necessarily mean that subcommand was passed, it just sets the "defaults").
 You cannot set positional-only arguments. Subcommands can be triggered from
--- a/README.md
+++ b/README.md
@@ -514,7 +514,7 @@ On the command line, options can be given as:
 - `-ffilename` (no space required)
 - `-abcf filename` (flags and option can be combined)
 - `--long` (long flag)
-- `--long_flag=true` (long flag with equals to override default value)
+- `--long_flag=true` (long flag with equals -- to override default value)
 - `--file filename` (space)
 - `--file=filename` (equals)
 
--- a/README.md
+++ b/README.md
@@ -1665,10 +1665,10 @@ brief description of each is included here
 
 To contribute, open an [issue][github issues] or [pull
 request][github pull requests] on GitHub, or ask a question on [gitter][]. There
-is also a [short note to contributors](./.github/CONTRIBUTING.md). This
-readme roughly follows the [Standard Readme Style][] and includes a mention of
-almost every feature of the library. More complex features are documented in
-more detail in the [CLI11 tutorial GitBook][gitbook].
+is also a [short note to contributors](./.github/CONTRIBUTING.md). This readme
+roughly follows the [Standard Readme Style][] and includes a mention of almost
+every feature of the library. More complex features are documented in more
+detail in the [CLI11 tutorial GitBook][gitbook].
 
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -514,7 +514,7 @@ On the command line, options can be given as:
 - `-ffilename` (no space required)
 - `-abcf filename` (flags and option can be combined)
 - `--long` (long flag)
-- `--long_flag=true` (long flag with equals to override default value)
+- `--long_flag=true` (long flag with equals -- to override default value)
```
jsoref, line 517: this is a little different from most changes and is very manual (all changes are manual, but this is more so). I'm happy to drop it or anything else, but I think it kinda helps to clarify that the second part describes the first.

cmake/CodeCoverage.cmake
```diff
@@ -71,7 +71,7 @@
 # - Added the option for users to set the GCOVR_ADDITIONAL_ARGS variable to supply additional
 #   flags to the gcovr command
 #
-# 2020-05-04, Mihchael Davis
+# 2020-05-04, Michael Davis
```
jsoref, line 74: I'm fairly certain this is the correct person's name, but I'm not really sure where this code comes from -- I'd love to fix the source material.

include/CLI/TypeTools.hpp
```diff
@@ -163,7 +163,7 @@ struct pair_adaptor<
     }
 };
 
-// Warning is suppressed due to "bug" in gcc<5.0 and gcc 7.0 with c++17 enabled that generates a Wnarrowing warning
+// Warning is suppressed due to "bug" in gcc<5.0 and gcc 7.0 with c++17 enabled that generates a -Wnarrowing warning
```
jsoref, line 166: I think adding the `-` isn't unreasonable...

tests/HelpersTest.cpp
```diff
@@ -1309,7 +1309,7 @@ TEST_CASE("Types: TypeNameStrings", "[helpers]") {
     auto wsclass = CLI::detail::classify_object<std::wstring>::value;
     CHECK(CLI::detail::object_category::wstring_assignable == wsclass);
 
-#if defined CLI11_HAS_FILEYSTEM && CLI11_HAS_FILESYSTEM > 0 && defined(_MSC_VER)
+#if defined CLI11_HAS_FILESYSTEM && CLI11_HAS_FILESYSTEM > 0 && defined(_MSC_VER)
```
jsoref, line 1312: Notably, this appears to be a broken guard

cmake/CodeCoverage.cmake
phlptp: The original is here https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake

tests/HelpersTest.cpp
phlptp: hmmm

phlptp: this is probably triggering the coverage issue, meaning some bits of code were never run

examples/CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ add_test(NAME shapes_all COMMAND shapes circle 4.4 circle 10.7 rectangle 4 4 cir
                                  4.5 ++ rectangle 2.1 ++ circle 234.675)
 set_property(
   TEST shapes_all PROPERTY PASS_REGULAR_EXPRESSION "circle2" "circle4"
-                           "rectangle2 with edges [2.1,2.1]" "triangel1 with sides [4.5]")
+                           "rectangle2 with edges [2.1,2.1]" "triangle1 with sides [4.5]")
```
phlptp, line 134: I have wondered that before but never looking into it,  I don't think these tests are doing what we think they are doing.  This should have been a test failure.

phlptp: obviously not an issue to fix in this PR

tests/HelpersTest.cpp
jsoref: I'm happy to drop things and leave them for you to address separately.

cmake/CodeCoverage.cmake
jsoref: Thanks. I posted https://github.com/bilke/cmake-modules/pull/90/files#r1878500171 to upstream this fix.

tests/HelpersTest.cpp
jsoref: Yeah, the extra coverage in https://app.codecov.io/gh/CLIUtils/CLI11/pull/1101/blob/include/CLI/TypeTools.hpp definitely would make sense based on being able to run this additional line:
https://github.com/CLIUtils/CLI11/blob/bdd12bf308e90e911da0e58c83ce72179c3ce3ec/tests/HelpersTest.cpp#L1313

phlptp: I will see about getting the coverage back in a different PR

phlptp:
Is that action generating the commit(s) automatically?   Is it something that could be run regularly on a CI job?

jsoref:
The action only complains about spelling. Commits require a human.

There are some tools that automatically suggest corrections (I'm not particularly enamored by any of them).

I have a spreadsheet that I've used in the past to get Google to make suggestions which can generate commands that I use with another script to make uniform commits. Recently I've just written the commands out by hand since I haven't had enough corrections relative to unknown words to justify using the spreadsheet.

phlptp:
This check seems to be catching more than was caught with the codespell run as part of pre-commit.   Wondering if it is something we should be running regularly?

jsoref:
That's my general experience -- I regularly find bugs in repositories that use codespell. It's easy enough to configure. If you're interested, I can make a PR to enable it after this is merged. There are a couple of tunables worth discussing. (The sample workflow here is using a prerelease version, so I'd switch it to the current shipping release...)

phlptp:
@jsoref I will probably dig deeper in the spelling for another repo, but I am going to merge this one.  Thank You!

jsoref:
@phlptp fwiw, I've rebased my checks and converted them from my prerelease to the latest release, (there's one spelling error that appears to have snuck in between when I prepared this PR and when it was merged), if you're interested in playing with the code, it's available here: https://github.com/jsoref/CLI11/tree/refs/heads/spell-check-with-spelling.

Or, if you want to play with check-spelling, you can try dropping https://github.com/check-spelling/spell-check-this/blob/prerelease/.github/workflows/spelling.yml into the workflows directory of another repository and let check-spelling walk you through adapting check-spelling to a repository.

PR #1106, opened 2024-12-23
Help priority adjustment

### LINKED ISSUE #1099: configurable subcommand becomes --help's specific output to screen
Hi,

When a subcommand is set to configurable
```
        CLI::App &security = *app.add_subcommand("security", "Security properties in config file");
        security.add_option("ssl-certification-location", sslCertificationLocation, "ssl certification location");
        security.configurable();

```
and has a section in the config toml
```
[security]
ssl-certification-location = "/etc/ssl/certs"
```
the help acts as if the user typed
```
cnpm --help security

#instead of
cnpm --help
```
when the ```App::_parse_single_config``` is called and runs thru the toml and finds the ```[security]``` section it push_back's the subcommand into ```parent_->parsed_subcommands_``` at line https://github.com/CLIUtils/CLI11/blob/main/include/CLI/impl/App_inl.hpp#L1502.  Then ```App::_process_help_flags``` is called and it checks that  ```parsed_subcommands_``` is not empty at line https://github.com/CLIUtils/CLI11/blob/main/include/CLI/impl/App_inl.hpp#L1229.  Any app programmed with a configured configurable subcommand never calls ```CallForHelp()``` or ```CallForAllHelp()```.

Kinda of a dilemma.  Could check if all in ```parsed_subcommands_``` are configurable ones and go with the ```CallForHelp()``` in that case yet the user might want to do
```
cnpm --help security
```

Maybe a property that says where it was parsed from?

?

phlptp:
Move the help generation priority higher so it triggers before config file processing.

Fixes #1099

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,10 +28,11 @@ jobs:
 
       - name: Get LCov
         run: |
-          wget https://github.com/linux-test-project/lcov/releases/download/v1.16/lcov-1.16.tar.gz
-          tar -xzf lcov-1.16.tar.gz
-          cd lcov-1.16
-          sudo make install
+          sudo apt-get install ca-certificates lcov
+          #wget https://github.com/linux-test-project/lcov/releases/download/v1.16/lcov-1.16.tar.gz
+          #tar -xzf lcov-1.16.tar.gz
+          #cd lcov-1.16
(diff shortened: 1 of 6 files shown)


phlptp:
will merge in 2 days unless someone sees any issues with this change.

Issue #1094, opened 2024-11-25
Could not convert: integer option passed as a string with an extra space after the value


aphecetche:
A very simple example (using version 2.4.2) "fails" when called with `-i "42 "` (note the extra space at the end of the string), but works fine for e.g. `-i " 42"` (space before the integer) (and of course for the more natural way of calling it directly with an integer, e.g. `-i 42` 
 
Is that a feature or a bug ? Knowing that e.g. `std::stoi` convert both " 42" and "42 " just fine, I'd call it a bug, but there might be something obvious I'm missing here ? 

see e.g. https://godbolt.org/z/jad4MfsKP to run it 

```
#include <iostream>
#include "CLI/CLI.hpp"

int main(int argc, char** argv)
{
  int value{0};

  CLI::App app{"basic cli example"};

  app.option_defaults()->always_capture_default();

  app.add_option("-i", value);

  try {
    app.parse(argc, argv);
  }
  catch (const CLI::ParseError& error) {
    std::cout << "CLI11 " << error.what() << "\n";
    return app.exit(error);
  }

  std::cout << "value=" << value << "\n";
}
```

phlptp:
it is mostly on purpose.   Internally it uses `std::strtoll`  to do the conversion and it has a check to make sure that is converts the entire string to an integer or other number.   `std::stoi` ignores trailing characters as well so it converts " 523badNUMBER" just as well as "523"   This is generally not what is desired, so the general philosophy has been that if it can't use all the characters in the conversion then something is wrong.  

  The exact question of whether trailing spaces should be allowed is probably debatable.

aphecetche:
ok, I see.  Considering that "523badNUMBER" is not a valid number indeed makes sense to me. 

But rejecting "523 " is a bit more surprising (and probably less obvious to spot from the user point of view).

PR #1115, opened 2025-01-13
Allow trailing spaces on strings for conversions

phlptp:
Allow trailing spaces on strings for conversions to floating point and integers.

There was a potential confusing error that could occur if strings with trailing spaces were converted to integer or floating point values.   This PR allows trailing spaces in the strings that would otherwise be convertible to the appropriate numerical type.


phlptp:
Fixes #1094

include/CLI/TypeTools.hpp
```diff
@@ -968,6 +968,9 @@ bool integral_conversion(const std::string &input, T &output) noexcept {
         nstring.erase(std::remove(nstring.begin(), nstring.end(), '\''), nstring.end());
         return integral_conversion(nstring, output);
     }
+    if(std::isspace(static_cast<unsigned char>(input.back()))) {
+        return integral_conversion(trim_copy(input), output);
```
henryiii, line 972: Not covered by tests?

phlptp: missed that section on first try,  adding now.

PR #1132, opened 2025-03-03
fix: promoting uint8_t/int8_t to make it casting result printable

ComixHe:
The underlying type of uint8_t/int8_t is unsigned char and signed char, but IOStreams was specified to treat them just like char.

so promoting the casting value to get the expected value.


phlptp (approved):

phlptp:
@all-contributors

please add @ComixHe for code

PR #1136, opened 2025-03-06
add tests and fixes for array options

### LINKED ISSUE #1135: `capture_default_str` on an `std::array`
Since the new release of CLI11 (2.5), this short example doesn't compile with clang 17.0.6. I haven't tested it with other compilers.

```cpp
#include <array>

#include <CLI/CLI.hpp>

int main(int argc, char** argv)
{
    std::array<int, 2> a{0, 1};
    CLI::App app{"My app"};
    app.add_option("--a", a, "an array")->capture_default_str();
    app.parse(argc, argv);

    return 0;
}
```

with the following error

```
In file included from cli11_test.cpp:3:
In file included from /Users/loic/mambaforge/envs/samurai-test/include/CLI/CLI.hpp:38:
/Users/loic/mambaforge/envs/samurai-test/include/CLI/App.hpp:551:20: error: no matching function for call to 'checked_to_string'
  551 |             return CLI::detail::checked_to_string<AssignTo, ConvertTo>(variable);
      |                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/loic/mambaforge/envs/samurai-test/include/CLI/App.hpp:550:79: note: while substituting into a lambda expression here
  550 |         Option *opt = add_option(option_name, fun, option_description, false, [&variable]() {
      |                                                                               ^
cli11_test.cpp:12:9: note: in instantiation of function template specialization 'CLI::App::add_option<std::array<int, 2>, std::array<int, 2>, (CLI::detail::enabler)0>' requested here
   12 |     app.add_option("--a", a, "an array")->capture_default_str();
      |         ^
/Users/loic/mambaforge/envs/samurai-test/include/CLI/TypeTools.hpp:456:6: note: candidate template ignored: substitution failure [with T1 = std::array<int, 2>, T2 = std::array<int, 2>, T = std::array<int, 2> &, $3 = 0]: call to 'to_string' is ambiguous
  456 | auto checked_to_string(T &&value) -> decltype(to_string(std::forward<T>(value))) {
      |      ^                                        ~~~~~~~~~
/Users/loic/mambaforge/envs/samurai-test/include/CLI/TypeTools.hpp:465:13: note: candidate template ignored: requirement '!std::is_same<std::array<int, 2>, std::array<int, 2>>::value' was not satisfied [with T1 = std::array<int, 2>, T2 = std::array<int, 2>, T = std::array<int, 2> &]
  465 | std::string checked_to_string(T &&) {
      |             ^
1 error generated.
```

I think it's a bug. Could you confirm this or tell me what has changed since version 2.4 ?

phlptp:
Fixes #1135.  Adds enable check to certain to_string functions as some std::array operations were ambiguous.  

The addition of the capability to convert tuples to strings created an ambiguity in the case std::array, which acts like a tuple and a container.   So it worked with the container conversion before, but could also work with the new tuple conversion.  And we didn't have any tests to catch this.   

This PR resolves the ambiguity, and adds some tests to check that array is handled well.

### DIFF
--- a/examples/array_option.cpp
+++ b/examples/array_option.cpp
@@ -3,13 +3,12 @@
 #include <CLI/CLI.hpp>
 
 // from @gourin https://github.com/CLIUtils/CLI11/issues/1135
-int main(int argc, char** argv)
-{
-   std::array<int, 2> a{ 0, 1};
-   CLI::App app{"My app"};
-   app.add_option("--a", a, "an array")->capture_default_str();
-   app.parse(argc, argv);
+int main(int argc, char **argv) {
+    std::array<int, 2> a{0, 1};
+    CLI::App app{"My app"};
+    app.add_option("--a", a, "an array")->capture_default_str();
+    app.parse(argc, argv);
 
-    std::cout<<"pass\n";
+    std::cout << "pass\n";
     return 0;
 }
--- a/examples/array_option.cpp
+++ b/examples/array_option.cpp
@@ -1,0 +1,15 @@
+#include <array>
+
+#include <CLI/CLI.hpp>
+
+// from @gourin https://github.com/CLIUtils/CLI11/issues/1135
+int main(int argc, char** argv)
+{
+   std::array<int, 2> a{ 0, 1};
+   CLI::App app{"My app"};
+   app.add_option("--a", a, "an array")->capture_default_str();
+   app.parse(argc, argv);
+
+    std::cout<<"pass\n";
+    return 0;
+}
--- a/examples/array_option.cpp
+++ b/examples/array_option.cpp
@@ -4,7 +4,7 @@
 //
 // SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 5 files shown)


examples/array_option.cpp
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2017-2025, University of Cincinnati, developed by Henry Schreiner
+// under NSF AWARD 1414736 and by the respective contributors.
+// All rights reserved.
+//
+// SPDX-License-Identifier: BSD-3-Clause
+
+// Code modified from @gourin https://github.com/CLIUtils/CLI11/issues/1135
```
gouarin: My name is gouarin and not gourin. Thanks for the fix.

phlptp: corrected, thanks

henryiii (approved):

PR #1151, opened 2025-04-18
Ci build update

phlptp:
update ci build images, remove ubuntu 20.04 and update a few others


henryiii (approved):
Any specific reason for the enum changes?

phlptp:
Making clang tidy happy mostly.   Updated to clang 20, it picked up a bunch of other stuff that I am still working through, some legitimate, some I will probably ignore.  Was going to see if the changes affected anything else on the older compilers and if so change it back.   It will probably be a few days before I get to it all.

Issue #1149, opened 2025-04-16
Handling typos in commands and subcommands


Hind-M:
Hey there!
We are using `CLI11` in our [project](https://github.com/mamba-org/mamba), and we were wondering if there is anything that we can use to handle typos in set commands/subcommands?
For now if we run for example:

`mamba instal test`

The printed error would be:
```
The following arguments were not expected: test instal
Run with --help for more information.
```

Which is ok as we can guess that `instal` is not the right command, but that gets more confusing if multiple arguments are used as the error is not really explicit and does not point to the exact offending argument.

For example:

`mamba reqoquery whoneeds _openmp_mutex`

outputs:
```
The following arguments were not expected: _openmp_mutex whoneeds reqoquery
Run with --help for more information.
```

Furthermore, the arguments in the output are written in the reverse order of the command.

cf. https://github.com/mamba-org/mamba/issues/3856

Maybe the error with multiple arguments could be more explicit, or we could use some kind of validation methods if available?
Additionally, if suggestions are supported in `CLI11` that would be awesome!

Thanks!

cxw42:
Git's "the most similar commands are..." output  would be a neat feature!

phlptp:
Any thought on what a good measure for "close" strings would be?   It should be possible to insert a method into one of the app/subcommand callbacks that can check this and return a useful message, but not sure what a good simple algorithm for what the "best" match should be.

cxw42:
I could think of a few options, roughly in increasing order of complexity:
* Prefix matches only
* [Hamming distance](https://en.wikipedia.org/wiki/Hamming_distance) < N
  * N determined empirically?  Perhaps based on the lengths of the subcommands?
* [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) within a threshold (looks like `git` uses this)

For a first cut, how about something like this (assuming the shortest subcommand has at least two characters):
1. Space-pad input and all subcommands to same length
2. Compute Hamming distance for each subcommand
3. Select subcommands with distance <= N (N == 1 if the shortest subcommand has only two characters, otherwise 2?)
  * **Edit** Or maybe choose N based on the number of chars the user typed?  If the user typed more chars, give the user more options?
5. Sort the remaining by "is input prefix of the candidate?" first, then distance ascending, then ASCII order ascending.
(Made up on the spot as a draft :D so no guarantees!)

phlptp:
[...]

PR #1152, opened 2025-04-21
add an example of finding close matches

phlptp:
continue discussion on #1149 
Adds subcommand prefix matching as a modifier to CLI.  
Adds and example of close matching logic for further exploration.


jzakrzewski:
That looks OK, but I'd really love some simple API to enable that behaviour.
As a bonus, similar thing prefix matching would be great (for example `myapp i`, `myapp ins`, `myapp install` would all do the same).

jzakrzewski:
I like that. Something like this could be a great help to the user of our apps.

I'm wondering, however, if the prefix match could also be implemented for the parser itself - basically as long as the prefix is unambiguous, the corresponding command would be used. Kinda-sorta what NPM does, but better (`npm in` still runs `npm install`, where it should error out because there's `npm init`).

phlptp:
I added prefix matching modifier to the app class,  That will work for prefix matching on subcommands.   I will add some more documentation and a few more tests yet, but review comments/suggestions would be appreciated.

jzakrzewski:
Well, for me that looks beautiful!
Thank you :)

phlptp:
I integrated prefix matching and the example has the close matching output.   @henryiii  what do you think of the prefix matching.   I am not ready to integrate the close matching as a standard feature available that will take a different PR.  But I think the prefix matching is ready or close to merging.

phlptp:
@henryiii  need your thoughts on this one

cxw42:
@phlptp Thanks for taking this up so quickly!  Just to check, should this PR be prefix-match only?
* Your https://github.com/CLIUtils/CLI11/pull/1152#issuecomment-2847665223 says 
  > I am not ready to integrate the close matching
* but the tests give a closest-match example: 
  https://github.com/CLIUtils/CLI11/pull/1152/files#diff-940e5356fa1137e646ea135a6f68cacbfad4fe7124c3a69163468f588acf9283R267

Based on your comment, should the "close match"/Levenshtein material be removed?

phlptp:
Right now the close match/levenshtein matching is in an example and the test is against that example.   The comment was referring to trying to get that capability into the main files, which I am not ready to try yet.   But I am fine with it being in an example of how that might be done.   I want to get the prefix matching in first then start thinking about what the best way of including the close match stuff into the main header files, or whether or not that should be included at all.

phlptp:
@henryiii Going to merge this in two days unless you have further comments.

PR #1158, opened 2025-05-05
Fix the order of arguments in command-line parsing error message

agalanin-at-nvidia:
CLI11 code prints the command-line arguments in reversed order in the error message.

Code to reproduce:
```c++
#include <string>
#include <CLI/CLI.hpp>

int main(int argc, const char **argv)
{
    CLI::App app{"Bug report app"};
    std::string foo;

    app.add_option("--foo", foo, "Foo option");
    CLI11_PARSE(app, argc, argv);

    return 0;
}
```

Reproduction:
```
$ g++ -Wall -Wextra -I CLI11/include/ cli11-bug-order-in-error.cpp -o cli11-bug-order-in-error && ./cli11-bug-order-in-error --foo bar --fizz buzz
The following arguments were not expected: buzz --fizz
```
Expected result:
```
The following arguments were not expected: --fizz buzz
```

### DIFF
--- a/include/CLI/Error.hpp
+++ b/include/CLI/Error.hpp
@@ -310,13 +310,13 @@ class ExtrasError : public ParseError {
     explicit ExtrasError(std::vector<std::string> args)
         : ExtrasError((args.size() > 1 ? "The following arguments were not expected: "
                                        : "The following argument was not expected: ") +
-                          detail::rjoin(args, " "),
+                          detail::join(args, " "),
                       ExitCodes::ExtrasError) {}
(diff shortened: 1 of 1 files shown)


phlptp:
I think there are a couple spots where calling the ExtrasError was correct with the `rjoin` so we are probably going to have to either  reverse those vectors or the calls where this change is now correct.   I think we will also want a couple test cases that check for this error in the various situations ExtrasError arises.   I can help with the test cases if needed but might be a few days.

phlptp (approved):

phlptp:
Going to merge this, I will add a few tests and checks on some other situations soon

PR #968, opened 2023-12-31
Support all capabilities for toml integers and floating point numbers

phlptp:
This PR add support for all TOML integer types.
This includes 
```
{"+99", 99},
    { "99",99 },
    { "0xDEADBEEF",0xDEADBEEF },
    { "0xdeadbeef",0xDEADBEEF },
    { "0xdead_beef",0xDEADBEEF },
    { "0xdead'beef",0xDEADBEEF },
    { "0o01234567",001234567 },
    { "0o755",0755 },
    { "0755",0755 },
    { "995862_262",995862262 },
    { "995862262",995862262 },
    { "+995862275",+995862275 },
    { "995'862'275",995862275 },
    {"0b11010110",0xD6},
    {"0b1101'0110",0xD6},
    {"1_2_3_4_5",12345},
```

So support for separators (Toml and C++ standards),  binary(new), hex(worked previously), and octal(new)

PR #1160, opened 2025-05-05
Default value conversion error if the application locale is set

agalanin-at-nvidia:
**Short description**
Application fails with `CLI::ConversionError` in some locales if the application locale is set.

**Root cause**
CLI11 uses `std::stringstream` to format default value and `strtol`/`strtod` to parse them back. The first method adds locale-specific thousand separators but the second method parses the number only up to the first separator. This causes an exception when the same value is converted to string representation and then parsed back.

Examples: in _en_US_ locale 1234.56 is "1,234.56" and in de_DE locale 1234.56 is "1.234,56".

**Reproduction**
```c++
#include <string>
#include <CLI/CLI.hpp>

int main(int argc, const char **argv)
{
    std::locale::global(std::locale(""));
    CLI::App app{"Bug report app"};
    long foo;

    app.add_option("FOO", foo, "Foo option")
        ->envname("FOO")
        ->default_val(1234567);
    CLI11_PARSE(app, argc, argv);

    return 0;
}
```
Output:
```
$ g++ -Wall -Wextra -I CLI11/include/ cli11-bug-locale-string.cpp -o cli11-bug-locale-string
$ for locale in C en_US de_DE fr_FR de_CH; do echo "locale $locale"; LC_NUMERIC=$locale.UTF-8 ./cli11-bug-locale-string && echo OK; echo; done
locale C
OK

locale en_US
terminate called after throwing an instance of 'CLI::ConversionError'
what():  The value FOO is not an allowed value for given default value("1,234,567") produces an error : Could not convert: FOO = 1,234,567
Aborted (core dumped)

locale de_DE
terminate called after throwing an instance of 'CLI::ConversionError'
what():  The value FOO is not an allowed value for given default value("1.234.567") produces an error : Could not convert: FOO = 1.234.567
Aborted (core dumped)

locale fr_FR
terminate called after throwing an instance of 'CLI::ConversionError'
what():  The value FOO is not an allowed value for given default value("1 234 567") produces an error : Could not convert: FOO = 1 234 567
Aborted (core dumped)

locale de_CH
OK
```

**Fix description**
The fix strips group separators from the input string. This extends a bit the logic introduced in change #968 for `_` and `'` separators.
There are no unit tests for the change because it requires to have all mentioned locales to be installed in the system.

### DIFF
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -961,6 +961,13 @@ bool integral_conversion(const std::string &input, T &output) noexcept {
         output = (output_sll < 0) ? static_cast<T>(0) : static_cast<T>(output_sll);
         return (static_cast<std::int64_t>(output) == output_sll);
     }
+    // remove locale-specific group separators
+    char group_separator = std::use_facet<std::numpunct<char>>(std::locale()).thousands_sep();
+    if(input.find_first_of(group_separator) != std::string::npos) {
+        std::string nstring = input;
+        nstring.erase(std::remove(nstring.begin(), nstring.end(), group_separator), nstring.end());
+        return integral_conversion(nstring, output);
+    }
     // remove separators
     if(input.find_first_of("_'") != std::string::npos) {
         std::string nstring = input;
@@ -1018,6 +1025,13 @@ bool integral_conversion(const std::string &input, T &output) noexcept {
         // this is to deal with a few oddities with flags and wrapper int types
         output = static_cast<T>(1);
         return true;
+    }
+    // remove locale-specific group separators
+    char group_separator = std::use_facet<std::numpunct<char>>(std::locale()).thousands_sep();
+    if(input.find_first_of(group_separator) != std::string::npos) {
+        std::string nstring = input;
+        nstring.erase(std::remove(nstring.begin(), nstring.end(), group_separator), nstring.end());
+        return integral_conversion(nstring, output);
     }
     // remove separators and trailing spaces
     if(input.find_first_of("_'") != std::string::npos) {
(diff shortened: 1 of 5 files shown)


agalanin-at-nvidia:
The other fix option is to force all formatting/parsing operations to be in "C" locale but it may impact the current users of the library who may already use locale-specific parsing.

phlptp:
I like the solution,  I am starting to play with ways to get some tests in place.   If I get something working I will try to merge with your branch.

phlptp:
I made a PR into the branch for this PR,  if that is merged then it should clear up some of the missing coverage.   There will likely need to be a second test with an unsigned int, and floating point to get the rest.

agalanin-at-nvidia:
Thanks, @phlptp, I added your [test case](https://github.com/agalanin-at-nvidia/CLI11/pull/1) into the PR.

BTW, I found that the issue is more complex than I thought. I've hacked `main` function in catch2 to set locale to "de_DE.UTF-8" and found that it breaks not only tests when the input is hard-coded as a string (it's expected that these tests will fail) but also tests for tuples:
```
TupleDefault
../tests/AppTest.cpp:494

../tests/AppTest.cpp:494: FAILED:
  {Unknown expression after the reported line}
  The value -i is not an allowed value for given default value("[99,5,test2,87,
  total3]") produces an error : Could not convert: -i = 99,5,test2,87,total3

TupleComplex
../tests/AppTest.cpp:520

../tests/AppTest.cpp:520: FAILED:
  {Unknown expression after the reported line}
  The value -i is not an allowed value for given default value("[99,5,test2,87,
  [total3,total4]]") produces an error : Could not convert: -i = 99,5,test2,87,
  total3,total4
```

phlptp:
I am going merge this,  there are a few things that will need to build on this and the incorporation of locales in cli11 is interesting.  I think there are a few other things that can build on this.    There will be a PR soon that resolves the CI failures.

PR #1173, opened 2025-06-27
Allow rvalue references to be use with add_flag with no assignment

tnixeu:
I am starting to use CLI11 in a project and I came across an inconvenience in code like this

```c++
const std::string description(std::format("description"));
app.add_flag("--flag", description);
```
As far as I can tell. I cannot directly use the description here. The reason seems to be to distinguish between assignment and description. So only `const std::string` or `const char*` and so on can be used here. But a `const std::string &&` can not be used.

With the changes in the PR the code can also be written like this:
```c++
app.add_flag("--flag", std::format("description"));
```

The `add_flag` function can then be used with a const string or an rvalue reference to a string.

A bit OT: I was a bit confused by this part of the existing comment ` if a variable string is passed that variable will be assigned the results from the flag` I think this might be wrong, but I am not sure.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -654,7 +654,7 @@ class App {
     /// Add a flag with no description or variable assignment
     Option *add_flag(std::string flag_name) { return _add_flag_internal(flag_name, CLI::callback_t(), std::string{}); }
 
-   /// Add flag with description but with no variable assignment or callback
+    /// Add flag with description but with no variable assignment or callback
     /// takes a constant string or a rvalue reference to a string,  if a variable string is passed that variable will be
     /// assigned the results from the flag
     template <typename T,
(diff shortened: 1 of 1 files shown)


phlptp:
>> I was a bit confused by this part of the existing comment  if a variable string is passed that variable will be assigned the results from the flag I think this might be wrong, but I am not sure.

This is basically saying that if you pass a non const string here, that is an assignment variable, so the results of parsing that flag would get passed into that variable.   `add_flag("--flag", variable, "description")`   so if the flag were used in the command line then  the output would get put into variable.

phlptp:
I like it,  may add a few more tests later on, but this looks good,  I am going to merge it

tnixeu:
> > > I was a bit confused by this part of the existing comment  if a variable string is passed that variable will be assigned the results from the flag I think this might be wrong, but I am not sure.
> 
> This is basically saying that if you pass a non const string here, that is an assignment variable, so the results of parsing that flag would get passed into that variable. `add_flag("--flag", variable, "description")` so if the flag were used in the command line then the output would get put into variable.

I see, but in this case the `std::enable_if` would block it. It seems as if the comment is meant for another function.

phlptp:
Yeah you are probably right, that comment is probably copied from the other overloads.   I will clean that up in another PR soon.  
Thanks

Issue #1175, opened 2025-07-10
Address Sanitizer failure on macOS when using C++20


chris-t-w:
When building my project with Address Sanitizer enabled, I'm seeing issues in CLI11 code. I have managed to create a small reproduction case.

| Field              | Value              |
| -------------------| ------------------ |
| **CLI11 Version**  | 2.5.0              |
| **OS**             | macOS Sequoia 15.5 |
| **CMake**          | 4.0                |
| **Compiler**       | Apple clang 17.0.0 |
| **Arch**           | arm64              |

---

# Reproduction Case

Create the following 3 files in a directory:

## main.cpp

```c++
#include <CLI/CLI.hpp>

int main(int argc, const char *argv[])
{
    CLI::App app;

    bool test = false;
    app.add_flag("--flag", test, "Test flag")->group("Test group");
    app.parse(argc, argv);

    return 0;
}
```

## CMakeLists.txt

```cmake
cmake_minimum_required(VERSION 3.20 FATAL_ERROR)

project(clitest)

find_package(CLI11 CONFIG REQUIRED)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
set(CMAKE_CONFIGURATION_TYPES "Sanitize")
set(CMAKE_BUILD_TYPE "Sanitize")
set(CMAKE_CXX_FLAGS_SANITIZE "-U_FORTIFY_SOURCE -O1 -g -fsanitize=address,undefined -fno-omit-frame-pointer -fno-common")
set(CMAKE_EXE_LINKER_FLAGS_SANITIZE "-fsanitize=address,undefined")
set(CMAKE_SHARED_LINKER_FLAGS_COVERAGE "-fsanitize=address,undefined")
set(CMAKE_MAP_IMPORTED_CONFIG_SANITIZE "Sanitize;RelWithDebInfo;Release;Debug")

add_executable(clitest "main.cpp")
target_link_libraries(clitest PRIVATE CLI11::CLI11)
target_compile_features(clitest PRIVATE cxx_std_20) # No issue with cxx_std_17
```

## vcpkg.json

```json
{
  "name": "clitest",
  "version-string": "0.1.0",
  "builtin-baseline": "43819d3734f73ef27055ac80b44f1e75be2f00a8",
  "dependencies": [
    "cli11"
  ],
  "overrides": [
    {
      "name": "cli11",
      "version": "2.5.0"
    }
  ]
}
```

## Run Reproduction Case

To run the reproduction case:

[...]

PR #1178, opened 2025-07-15
Fix warnings and small issues

phlptp:
try some things with no standard libraries
Add arm64 and freebsd build and tests
fix a discrepancy in the handling of chars on Arm processors


phlptp:
Fixes #1175

PR #1179, opened 2025-07-18
add github actions job for armv7

phlptp:
more tests with other platforms


phlptp:
I don't think this adds much of value, not going to merge

PR #1180, opened 2025-07-18
Update version in license to match source.

pmundkur:
For packagers that need to install license files, it helps when the version in the license matches the version of the source used in the package.


henryiii:
We probably should add some automation somewhere to make sure this stays in sync.

phlptp (approved):

Issue #1184, opened 2025-07-23
needs() with add_option_function()


glonnber:
I'm not sure if I've done this correctly but it seems that ->needs() does not work with add_option_function(), i.e. having the following:

    cli_app.add_option_function<std::string>("file", [&](std::string const& str) { foobar(str); }, "input file");
    cli_app.add_option_function<std::string>("--something" [&](std::string const& str) { kebab(str); }, "something")->needs("file");

will launch the lambda for "--a" even if it's defined by itself. Have I messed up something, is this by design or is this a bug?

phlptp:
I will test it out

phlptp:
This was interesting.  I did verify it.   The fix was pretty straightforward.

 I was thinking about why this hasn't come up before.     My conclusion was in nearly all use cases it doesn't matter,  if the parser generates an exception then you either deal with it or exit, and it matters very little what other stuff was executed in the meantime.   But to be a little better CLI should be exception invariant for user callbacks as @glonnber pointed out.   So switching the order to do so had minimal impact on operations and other than some config file test cases in which the exception type changed as a result there were no other impacts.

PR #1186, opened 2025-07-24
reorder the processing of requirements and callbacks.  

phlptp:
Check the requirements first.
Previously the callbacks were done first,  but with custom callbacks this cause side effects unexpectedly.


phlptp:
Fixes #1184

Issue #881, opened 2023-05-10
Encoding problem with `std::string_view` on macOS


Jean1995:
I have the following minimal example:

```
#include "CLI11.hpp"

#include <iostream>

int main(int argc, char** argv) {
    CLI::App app{"App description"};

    app.add_option("--input", "input option")
     ->default_val("optionA")
     ->check(CLI::IsMember({"optionA", "optionB", "optionC"}));

    CLI11_PARSE(app, argc, argv);

    auto const inputStr = app["--input"]->as<std::string_view>();

    std::cout << inputStr << std::endl;

    return 0;
}
```

If I run the program without any input arguments on Linux (g++), I get the expected output `optionA`.

However, if I run it on macOS (clang++), I get the output ` ����G`, so some encoding problems. If I specify an argument via `--input`, the problem is gone. The problem is also gone by replacing `std::string_view` with `std::string`, so this appears to be a bug specific to the usage of `std::string_view`.

phlptp:
I can verify this does produce an inconsistent result.  
Not entirely sure what to do about it though.    The fact that it works on g++ is either an inconsistent fluke or some internal compiler magic.    

`as<T>()` when no arguments have been passed does some processing on the default_str as a temporary and does a conversion to get the results in the desired type.     String_view is a view on a string so in this case is just creating a view of a temporary which is undefined behavior, as seen by sometimes working, sometimes not.    When you have passed the variable it is stored internally so a `view` of it works fine.   

I need to think a bit to see if there is a way to make this work.  I agree it is an inconsistency in the API for this not to work, but not entirely sure it should work or we need to make the fail more explicit.  stay tuned.

phlptp:
There are a couple different ways I see to fix this.  They all have some drawbacks.  
The first is to make one of the internal results arrays `mutable`.    Then when a results method is called if no arguments have been passed then it gets copied to the internal array.   This internal array is already used for that purpose so add no new structure and doesn't impact existing operations in any way also doesn't add any computation.    The downside is there is now a mutable structure and in a few select cases around the  operations, this method is now not thread compatible, and internal things can change if the same method is executed on multiple threads could cause a race condition.  This is not a race condition that affects state.   

The other method would involve executing part of the processing on all options all the time, instead of just as needed.  This gets around the race condition but does add extra computation that in most cases is unnecessary.  

Not entirely sure which method is preferred.

PR #1187, opened 2025-07-26
use of string_view in as<T> method

phlptp:
Address Issue #881, allowing use of string_view in the as<XX> method on options.

### DIFF
--- a/include/CLI/Option.hpp
+++ b/include/CLI/Option.hpp
@@ -319,7 +319,7 @@ class Option : public OptionBase<Option> {
     /// complete Results of parsing
     results_t results_{};
     /// results after reduction
-    results_t proc_results_{};
+    mutable results_t proc_results_{};
     /// enumeration for the option state machine
     enum class option_state : char {
         parsing = 0,       //!< The option is currently collecting parsed results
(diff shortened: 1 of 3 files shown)


include/CLI/Option.hpp
```diff
@@ -319,7 +319,7 @@ class Option : public OptionBase<Option> {
     /// complete Results of parsing
     results_t results_{};
     /// results after reduction
-    results_t proc_results_{};
+    mutable results_t proc_results_{};
```
henryiii, line 322: Why mutable? Generally that means this is thread safe, while `results_t` won't necessarily be thread safe?

phlptp: The key issue is the `results` and related `as<T>` method.  These methods are `const` as normally they  should not result in changes to the data structure. This converts the results to whatever is requested.   But in cases where the results are requested as a string_view and there was no data given in the parsing, and there was a default value given.  This can lead to undefined behavior in the way we were doing it, as the string_view could end up pointing to a temporary.  Some compilers handled this gracefully others didn't.   The simplest lowest impact solution to this was to make proc_results which is where processed results are stored in the mutable object so it can be modified in a `const` method in those few select cases and prevent undefined behavior.   The other options were to just ban string_view but would not solve the underlying problem, just the most common exemplar of it.   Make some more complicated storage buffers,  or force processing for all options to pre fill the results structures.    My assessment was that this was the least intrusive and simplest option.

phlptp:
@henryiii unless you have an objection I will merge this Saturday morning.

Issue #1191, opened 2025-08-09
`check(CLI::ValidIPV4)` Does Not Account For Trailing Period


adachng:
## Description
If I have the code:
```cpp
        CLI::App app;

        std::string address_str;

        app.add_option("-a, --address",
                       address_str,
                       "IPv4 address of server")
            ->required()
            ->check(CLI::ValidIPV4);
```
and I run the app with `./main -a 127.0.0.1.`, the validator allows this when this is not a valid IPv4 address.

PR #1192, opened 2025-08-10
move some of the Validators to an ExtraValidators file

phlptp:
Rework some of the validator locations, add documentation, and fix some lingering issues with validators.
The extra will will enable additions of some new validators and reduce compile times for those that are not needed.


phlptp:
Fixes #1191

phlptp:
@henryiii I think you are going to need to modify the codacy configuration on their website,  I don't think I can fix the false positives on codacy in a configuration, so that will need to be handled by the administrator of the codacy configuration, which I think is you.   I think we might want to just disable cppcheck on codacy, I am not seeing it providing much value over the other static checking we are doing.

phlptp:
@henryiii This change will allow us to add some additional validators in a way that doesn't impact compile time unless they are needed.   Also adds some additional documentation on custom validators that has been requested.

phlptp:
@henryiii I will give it a couple more days then merge this.

PR #1199, opened 2025-08-28
fix edge case with config files pointer 

### LINKED ISSUE #1197: Code that runs correctly with version 2.3.2 produces a seg-fault when built and executed with version 2.5.0.
Upgrading cli11 from version 2.3.2 to 2.5.0.
Code that previously functioned correctly when built using version 2.3.2 throws a seg-fault when built with version 2.5.0.
Reading through the release notes for the versions above 2.3.2 I can see no advisories or breaking-changes noted.

Here's a doctest test-case that demonstrates the issue.  The test runs correctly when built against 2.3.2 but generates a seg-fault when built against 2.5.0.  The code initialises the CLI app to handle a config file, but no config file is supplied on the input parameters.  The seg-fault is generated during the attempt to convert the config pointer returned by the get_config_ptr() API to a string via the 'as' API.

Replacing '**app.get_config_ptr()->as<<std::string>>()**' with '**app["--conf"]->as<<std::string>>()**' produces the same result (seg-fault with 2.5.0).

Using GCC12 compiler on Debian 12.

Am I doing something wrong, or is this a bug that has been introduced somewhere after version 2.3.2?

```
TEST_CASE("Crash demo") {
    const char* argv[] = {"cli", "spdlog", "--level=off"};

    CLI::App app;
    app.configurable()->allow_config_extras(false);
    app.set_config("--conf")->check(CLI::ExistingFile);

    std::string level;

    auto *command = app.add_subcommand("spdlog");
    command->add_option("--level", level, "Log level")->default_val("info");

    app.parse(3, argv);


    if (auto conf_filename = app.get_config_ptr()->as<std::string>(); !conf_filename.empty()) {
        std::cout << "Configuration filename: " << conf_filename << std::endl;
    } else {
        std::cout << "No configuration file used" << std::endl;
    }

    std::cout << "level is " << level << std::endl;
}
```

phlptp:
generating a seg fault if no default and no config file provided.

Fixes #1197 

This was likely introduced by the combination of fixes for some issues with the config parsing and some updates to the as<T> method a while back.   This edge case on the handling of the config pointer with as was not overlooked in the earlier testing.

### DIFF
--- a/include/CLI/Option.hpp
+++ b/include/CLI/Option.hpp
@@ -681,18 +681,15 @@ class Option : public OptionBase<Option> {
         bool retval = false;
         if(current_option_state_ >= option_state::reduced || (results_.size() == 1 && validators_.empty())) {
             const results_t &res = (proc_results_.empty()) ? results_ : proc_results_;
-            if (!res.empty())
-            {
+            if(!res.empty()) {
                 retval = detail::lexical_conversion<T, T>(res, output);
-            }
-            else
-            {
+            } else {
                 results_t res2;
                 res2.emplace_back();
                 proc_results_ = std::move(res2);
                 retval = detail::lexical_conversion<T, T>(proc_results_, output);
             }
-            
+
         } else {
             results_t res;
             if(results_.empty()) {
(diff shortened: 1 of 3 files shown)


phlptp:
this edge case could also apply if someone was manipulating and running an option callback outside of the app.  The fix resolves both situations.

PR #1204, opened 2025-09-05
App::_process: add conditional logic for old vs. new help processing

VolkerChristian:
Some use cases require altering subcommand/option states (and thus the help output) based on a flag stored in the config file. In this case, running `_process_callbacks()` before `_process_help_flags()` is necessary.

A new compile-time option `CLI11_USE_OLD_HELP_PROCESSING` restores this “old” behavior by executing callbacks prior to help flag processing.

Extends #1186

### DIFF
--- a/include/CLI/impl/App_inl.hpp
+++ b/include/CLI/impl/App_inl.hpp
@@ -1413,9 +1413,11 @@ CLI11_INLINE void App::_process_requirements() {
 }
 
 CLI11_INLINE void App::_process() {
+    #ifndef CLI11_USE_OLD_HELP_PROCESSING
     // help takes precedence over other potential errors and config and environment shouldn't be processed if help
     // throws
     _process_help_flags();
+    #endif // CLI11_USE_OLD_HELP_PROCESSING
     std::exception_ptr config_exception;
     try {
         // the config file might generate a FileError but that should not be processed until later in the process
@@ -1427,11 +1429,19 @@ CLI11_INLINE void App::_process() {
     } catch(const CLI::FileError &) {
         config_exception = std::current_exception();
     }
+    #ifndef CLI11_USE_OLD_HELP_PROCESSING
     // callbacks and requirements processing can generate exceptions which should take priority
     // over the config file error if one exists.
     _process_requirements();
+    #endif // CLI11_USE_OLD_HELP_PROCESSING
 
     _process_callbacks();
+
+    #ifdef CLI11_USE_OLD_HELP_PROCESSING
+    _process_help_flags();
+
+    _process_requirements();
+    #endif // CLI11_USE_OLD_HELP_PROCESSING
 
     if(config_exception) {
         std::rethrow_exception(config_exception);


phlptp:
should this be configurable at runtime vs the compile time switch

VolkerChristian:
> should this be configurable at runtime vs the compile time switch

That’s a very good question. If the old help procedure is needed, I’d assume it applies to the entire project. So it probably wouldn’t make much sense to expose this as a simple runtime flag on individual commands.
If anywhere, I could imagine it as a setting on the CLI::Formatter, since the formatter is really the component that benefits from this behavior.

What do you think?

phlptp:
I don't think the formatter could play a role in this, as that doesn't come into play until after the help Exception is triggered.   The question is in _process() method in the app
```c++
_process_help_flags();
try {
    // the config file might generate a FileError but that should not be processed until later in the process
    // to allow for help, version and other errors to generate first.
    _process_config_file();

    // process env shouldn't throw but no reason to process it if config generated an error
    _process_env();
} catch(const CLI::FileError &) {
    // callbacks can generate exceptions which should take priority
    // over the config file error if one exists.
    _process_callbacks();
    throw;
}

_process_callbacks();

_process_requirements();
```

The current ordering came about through a variety of issues and conditions and I think I am happy with it as a default.   But other than config and env processing needing to be before callbacks and requirements any other ordering is possible.   

help, callbacks, requirements
help, requirements, callbacks,
requirements, help, callbacks
requirements, callbacks, help
callbacks, requirements, help
callbacks, help, requirements

All will have slightly different behavior in the mixed cases, and determine if you want to prioritize help, or other errors.  And also as a look closer there is going to be some deviations in behavior in the ordering for processing subcommands and help and requirements checking, that should be made consistent at some point.

phlptp:
@VolkerChristian are you going to do anything further with this in the very near future.   I am looking to get a release out soon, and would like to know if I should plan on this being in a new release or should wait until after that?

VolkerChristian:
> @VolkerChristian are you going to do anything further with this in the very near future. I am looking to get a release out soon, and would like to know if I should plan on this being in a new release or should wait until after that?

Oh, I see! I thought a lot about this patch and how to cleanly add all of the mentioned six combinations and make that configurable. I didn't find a elegant and clean solution. As I am obviously the only one who suffers from this new processing order I decided that I stay with my local patched version. Thus, this pull request can be closed for now.

**Nevertheless**, I prepared two further pull request which would be fine to find their way into the next release.

phlptp:
@VolkerChristian I am trying out an alternative to this with fine grained control on the options to determine when the option callbacks execute with respect to help, and requirements processing.  I think that addresses the issue you raised in this PR and the issue in #1221

VolkerChristian:
> @VolkerChristian I am trying out an alternative to this with fine grained control on the options to determine when the option callbacks execute with respect to help, and requirements processing. I think that addresses the issue you raised in this PR and the issue in #1221

@phlptp: I took a quick look at the patch. It's a nice approach! I will give it a try soon.

phlptp:
@VolkerChristian have you had a chance to try out the changes in #1223 ?

VolkerChristian:
> @VolkerChristian have you had a chance to try out the changes in #1223 ?

Yes i did a view days ago but haven't reported back.
It is indeed a nice approach but exactly my priority ordering could not be created with the current implementation. I have create another pull request (#1226) where all possible priority ordering is possible. Additionally I have added a callback_priority field in OptionBase to be able to set a default priority.

phlptp:
Closing superseded by #1226

PR #1206, opened 2025-09-14
Update App.hpp

romanholidaypancakes:
Compatible compile option: /GR-


phlptp:
Think it worth adding this flag to one of the windows CI builds?  If so which one?

phlptp:
@all-contributors please add @romanholidaypancakes for code

PR #1212, opened 2025-09-23
Recursive needs processing

VolkerChristian:
The current way needs relations are checked in void App::_process_requirements() is missleading.
Assume you have an application 'test' with a CLI11 config system configuring following needs relations between some subcommands:

```
app ---> sub1 ---> sub11
     |         |-> sub12
     |-> sub2 ---> sub21
```
A) Current needs processing gives:
```
./test                              -> test requires sub2  // OK for the user
./test sub2                         -> test requires sub1  // OK for the user
./test sub2 sub1                    -> sub2 requires sub21 // Confuses user
./test sub2 sub1 sub21              -> sub1 requires sub12 // Confuses user
./test sub2 sub1 sub21 sub12        -> sub1 requires sub11 // Hm
./test sub2 sub1 sub21 sub12 sub11  -> complete
```

B) Recursive needs processing gives:
```
./test                              -> test requires sub2  // OK
./test sub2                         -> sub2 requires sub21 // More intuitive for user
./test sub2 sub21                   -> test requires sub1  // More intuitive for user
./test sub2 sub21 sub1              -> sub1 requires sub12 // OK
./test sub2 sub21 sub1 sub12        -> sub1 requires sub11 // OK
./test sub2 sub21 sub1 sub12 sub11  -> complete
```

### DIFF
--- a/include/CLI/impl/App_inl.hpp
+++ b/include/CLI/impl/App_inl.hpp
@@ -1414,7 +1414,7 @@ CLI11_INLINE void App::_process_requirements() {
                 }
             }
         }
-        if(sub->count() > 0 || sub->name_.empty() && need_subcommands_.find(sub.get()) == need_subcommands_.end()) {
+        if((sub->count() > 0 || sub->name_.empty()) && need_subcommands_.find(sub.get()) == need_subcommands_.end()) {
             sub->_process_requirements();
         }
 
--- a/include/CLI/impl/App_inl.hpp
+++ b/include/CLI/impl/App_inl.hpp
@@ -1317,10 +1317,22 @@ CLI11_INLINE void App::_process_requirements() {
             missing_need = opt->get_name();
         }
     }
+    for(const auto &subc : need_subcommands_) { // Process subcommands given on the commandline first
+        if (subc->count() > 0) {
+            subc->_process_requirements();
+        }
+    }
     for(const auto &subc : need_subcommands_) {
-        if(subc->count_all() == 0) {
-            missing_needed = true;
-            missing_need = subc->get_display_name();
+        if (subc->count() == 0) {
+            try {
+                subc->_process_requirements();
+            } catch (const CLI::RequiresError&) {
+                throw RequiresError(get_display_name(), subc->name_);
+            }
+            if(subc->count_all() == 0) {
+                missing_needed = true;
+                missing_need = subc->get_display_name();
+            }
         }
     }
     if(missing_needed) {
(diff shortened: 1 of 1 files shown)


include/CLI/impl/App_inl.hpp
```diff
@@ -1317,10 +1317,22 @@ CLI11_INLINE void App::_process_requirements() {
             missing_need = opt->get_name();
         }
     }
+    for(const auto &subc : need_subcommands_) { // Process subcommands given on the commandline first
```
phlptp, line 1320: So you want to check the subcommands that are needed and were given and make sure all their requirements are met first.   This is to get at the lower level errors with priority.

include/CLI/impl/App_inl.hpp
```diff
@@ -1317,10 +1317,22 @@ CLI11_INLINE void App::_process_requirements() {
...
+    }
     for(const auto &subc : need_subcommands_) {
-        if(subc->count_all() == 0) {
-            missing_needed = true;
-            missing_need = subc->get_display_name();
+        if (subc->count() == 0) {
+            try {
+                subc->_process_requirements();
+            } catch (const CLI::RequiresError&) {
+                throw RequiresError(get_display_name(), subc->name_);
```
phlptp, line 1330: next you are going to check the subcommands that were not parsed and give priority to missing requirements in those needed subcommands, and make the error clearer.

include/CLI/impl/App_inl.hpp
```diff
@@ -1402,7 +1414,7 @@ CLI11_INLINE void App::_process_requirements() {
                 }
             }
         }
-        if(sub->count() > 0 || sub->name_.empty()) {
+        if(sub->count() > 0 || sub->name_.empty() && need_subcommands_.find(sub.get()) == need_subcommands_.end()) {
```
phlptp: this check is to process the requirements for the subcommands that were given but were not in the needed list

phlptp:
I can see this being reasonable and cleaning up some of the error messages in specific situations.   Can we add some test cases, figure out what is going on with the existing test cases and we can merge it.

include/CLI/impl/App_inl.hpp
phlptp: I think this does need another set of parenthesis around the original 2 conditions, otherwise the logic is odd.   And I think that is why the test cases are failing.

VolkerChristian:
> I can see this being reasonable and cleaning up some of the error messages in specific situations. Can we add some test cases, figure out what is going on with the existing test cases and we can merge it.

Fine that you want to integrate this behavior. Nevertheless, I am extremely busy currently and I don't know when there is enough time to add the test cases and to have a deeper look at the now failed test about option groups. I want to suggest to postpone this patch and not to integrate it in the next release as I don't want to delay the release.

PR #1213, opened 2025-09-23
Add a required_callback to inject user defined requirement checks

VolkerChristian:
In some situations the standard requirement checks are not sufficient. In such cases the user can provide a required_callback doing additional requirement checks. This callback is executed after all standard requirement checks.


phlptp:
will need a few test cases and some documentation.   Do you have an example use case that is driving this change?

Issue #1224, opened 2025-10-14
Proposing OSS-Fuzz integration (needs primary_contact Google-account email)


TheodorNEngoy:
We've prepared an OSS-Fuzz PR: https://github.com/google/oss-fuzz/pull/14131.
OSS-Fuzz requires a **committer** using a **Google account** as  for access to ClusterFuzz reports. Please reply with the email to add. Thanks!

phlptp:
@TheodorNEngoy looks like you need to update the copyright dates on the PR to 2026

PR #1225, opened 2025-10-14
ci: add CIFuzz

TheodorNEngoy:
Add OSS-Fuzz CIFuzz workflow to run fuzzers on PRs for project 'cli11'.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,7 @@
 name: CIFuzz
-true:
+on:
+  pull_request:
+  workflow_dispatch:
 - pull_request
 jobs:
   fuzz:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -3,6 +3,7 @@ on: [pull_request]
 jobs:
   fuzz:
     runs-on: ubuntu-latest
+    continue-on-error: true
     permissions:
       security-events: write
     steps:
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,0 +1,32 @@
+name: CIFuzz
+on: [pull_request]
+jobs:
+  fuzz:
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
+      - name: Build Fuzzers
+        id: build
+        uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+        with:
+          oss-fuzz-project-name: 'cli11'
+          language: c++
+      - name: Run Fuzzers
+        uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+        with:
+          oss-fuzz-project-name: 'cli11'
+          language: c++
+          fuzz-seconds: 600
+      - name: Upload Crash
+        if: failure() && steps.build.outcome == 'success'
+        uses: actions/upload-artifact@v4
+        with:
+          name: artifacts
+          path: out/artifacts
+      - name: Upload SARIF
+        if: always() && steps.build.outcome == 'success'
+        uses: github/codeql-action/upload-sarif@v2
+        with:
+          sarif_file: cifuzz-sarif/results.sarif
+          checkout_path: cifuzz-sarif
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,38 +1,31 @@
 name: CIFuzz
+
 on:
   pull_request:
   workflow_dispatch:
-- pull_request
+
+permissions:
+  contents: read
+
 jobs:
   fuzz:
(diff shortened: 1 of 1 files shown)


phlptp:
@TheodorNEngoy I am excited to help with this, though I would like it to use the Fuzzer code that is built into CLI11 already rather than a new file.

TheodorNEngoy:
Please approve Actions so CIFuzz runs. Also, for OSS‑Fuzz we need a committer’s Google‑account email to use as `primary_contact` on https://github.com/google/oss-fuzz/pull/14131 (tracking issue: https://github.com/CLIUtils/CLI11/issues/1224). Thanks!

TheodorNEngoy:
Thanks @phlptp — I set you as `primary_contact` in the OSS‑Fuzz PR (google/oss-fuzz#14131).

phlptp:
I presume this isn't supposed to work yet?

TheodorNEngoy:
CIFuzz uses the OSS‑Fuzz project config for 'cli11', so it will stay red until google/oss-fuzz#14131 is merged. All other checks are green.

TheodorNEngoy:
CIFuzz is set to soft-fail temporarily per OSS‑Fuzz CI requirements (project must exist in OSS‑Fuzz). We’ll revert once google/oss-fuzz#14131 is merged.

phlptp:
So I am clear on this.  This action will execute the configuration found in oss-fuzz related to cli11.   It will run the fuzzer for 600 seconds.   Is that 600 for each type of fuzzer or total?   Is there anything special about 600?    I understand wanting to test the workflow for OSS-fuzz with each PR in CLI11, but is there much to be gained from running for 10 minutes vs 5 minutes for example?

TheodorNEngoy:
@phlptp — quick answers on CIFuzz:

• Yes: the job builds and runs the fuzz targets using the OSS‑Fuzz project config for `cli11`.
• `fuzz-seconds` is the *total* time budget for the “Run Fuzzers” step (default 600s). With OSS‑Fuzz coverage enabled, CIFuzz runs only fuzzers affected by the PR; otherwise it divides that time among all fuzz targets.
• 600s isn’t special beyond being the documented default/minimum. Maintainers should choose the longest acceptable time for the repo. If we later use a sanitizer matrix, each matrix job gets its own budget (e.g., ASan 600s, UBSan 600s).

Docs: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

TheodorNEngoy:
Switched CIFuzz to dry-run temporarily because the OSS-Fuzz project for 'cli11' isn’t merged yet. This keeps CI green while still running fuzzers. We’ll remove dry-run after google/oss-fuzz#14131 lands.

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,36 @@
+name: CIFuzz
+true:
```
phlptp:
```suggestion
on:
```

this should be "on" to be consistent with the other workflows

phlptp:
@TheodorNEngoy lets get the requested change in then I think this can be merged unless you want further changes.

TheodorNEngoy:
All checks are green here. The requested change on the OSS‑Fuzz side is in (only packaging fuzz_dictionary1). CIFuzz is temporarily dry‑run per docs, and we’ll flip it off after google/oss-fuzz#14131 merges. From my side this is ready — please approve the pending workflow and merge when convenient.

phlptp:
The use of "true" in the workflow needs to be changed to "on" otherwise I think this will run on every commit, which I don't think we want.   once that change is made I will merge this.

TheodorNEngoy:
Updated: use a proper `on:` block with `pull_request` (+ `workflow_dispatch`) so CIFuzz doesn’t run on every push. Still `dry-run: true` until google/oss-fuzz#14131 lands. Ready from my side.

phlptp (approved):

.github/workflows/cifuzz.yml
```diff
@@ -0,0 +1,38 @@
+name: CIFuzz
+on:
+  pull_request:
+  workflow_dispatch:
+- pull_request
```
phlptp: looks like this is causing an invalid file error on yaml.   I think the -pull_request shouldn't be there

phlptp (changes requested):
need to fix the work flow file now , pull request shows up twice

TheodorNEngoy:
pre-commit.ci autofix

TheodorNEngoy:
pre-commit.ci run

TheodorNEngoy:
pre-commit.ci run

TheodorNEngoy:
Fixed the YAML (removed duplicate `pull_request` item, proper `on:` block). CIFuzz remains `dry-run: true` in Build+Run per OSS‑Fuzz docs. Ready to merge once checks pass.

phlptp:
@all-contributors please add @TheodorNEngoy  for infrastructure

Issue #1209, opened 2025-09-19
Trailing empty lines with new help formatter


sebsura:
I recently tried out the new help formatter in 2.5.0 and I am wondering if the following changes are intended or not.

```C++
#include "CLI/App.hpp"
#include "CLI/Config.hpp"
#include "CLI/Formatter.hpp"

int main(int argc, char const* argv[])
{
  CLI::App app;

  bool big_option;
  app.add_option("--big", big_option);

  auto* group = app.add_option_group("group");

  std::string x;
  int y;
  char z;
  group->add_option("a", x);
  group->add_option("b", y);
  group->add_option("c", z);

  CLI11_PARSE(app, argc, argv);
}
```

This leads to some unexpected output now:
```
$ clitest --help

clitest [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit 
          --big BOOLEAN       
[Option Group: group]
  
POSITIONALS:
  a TEXT                      
  b INT                       
  c CHAR                      

OPTIONS:
  -h,     --help              Print this help message and exit 

```

Previously the output looked like
```
Usage: clitest [OPTIONS]

Options:
  -h,--help                   Print this help message and exit
  --big BOOLEAN               
[Option Group: group]
  Positionals:
    a TEXT                      
    b INT                       
    c CHAR      
```

The big differences for me are:
1. Previously there was no trailing empty newline
2. the option groups are not indented anymore
3. option groups contain the `--help` flag; 

While 2. may be intentional, I would think that at least 1) and 3) are not.

phlptp:
I think the help might be fixed in the current main.  The trailing newline might be as well.  Are you working on the 2.5.0 release or current main?

sebsura:
Yes, it looks much better on the current main.  I was using 2.5.0.

On main there is still a trailing new line if using trailers though:
```
#include "CLI/App.hpp"
#include "CLI/Config.hpp"
#include "CLI/Formatter.hpp"

int main(int argc, char const* argv[])
{
  CLI::App app;

  app.footer("some footer");

  CLI11_PARSE(app, argc, argv);
}
```
outputs
```

./core/src/tests/clitest [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit 

some footer 

```

PR #1229, opened 2025-10-18
Help spacing

phlptp:
Fix the help spacing and double printing of footers in certain circumstances


phlptp:
Fixes #1209

phlptp:
split off changes from the version update for better clarity

PR #1233, opened 2025-10-19
ci(cifuzz): enable real fuzzing (remove dry-run)

TheodorNEngoy:
Now that cli11 is in OSS‑Fuzz, switch CIFuzz from dry‑run to live per the CI docs.


phlptp:
This will be merged after https://github.com/google/oss-fuzz/pull/14131 is merged

phlptp:
@TheodorNEngoy looks like there is some changes that need to be merged into the oss-fuzz PR.   Can you take care of that?

Issue #1237, opened 2025-10-22
Missing symbols in 2.6.0 without CLI11_ENABLE_EXTRA_VALIDATORS


sethrj:
I built CLI11 2.6.0 with default options (no explicit CLI11_ENABLE_EXTRA_VALIDATORS) and with the precompiled library (CLI11_PRECOMPILED). In my downstream library it compiles correctly, using `-DCLI11_COMPILE`, but fails to link (despite being linked to libCLI11.a):
```
Undefined symbols for architecture arm64:
  "CLI::detail::PermissionValidator::PermissionValidator(CLI::detail::Permission)", referenced from:
      ___cxx_global_var_init.11 in celer-sim.cc.o
      ___cxx_global_var_init.12 in celer-sim.cc.o
      ___cxx_global_var_init.13 in celer-sim.cc.o
      ___cxx_global_var_init.11 in libceler_app_utils.a[2](CliUtils.cc.o)
      ___cxx_global_var_init.12 in libceler_app_utils.a[2](CliUtils.cc.o)
      ___cxx_global_var_init.13 in libceler_app_utils.a[2](CliUtils.cc.o)
```
Note that `libCLI11.a` has most other symbols but not PermissionValidator.

These symbols aren't used in my project (which worked fine with 2.5) but I think are from the global static variables
```c++
const detail::PermissionValidator ReadPermissions(detail::Permission::read);
const detail::PermissionValidator WritePermissions(detail::Permission::write);
const detail::PermissionValidator ExecPermissions(detail::Permission::exec);
```
which are somehow being included through `#include <CLI/CLI.hpp>`. I think there may be an error with the logic for `CLI11_ENABLE_EXTRA_VALIDATORS`.

Installation was done with:
```
[...]

Issue #1236, opened 2025-10-20
Build failure with catch 3.10.0


jamesjer:
With catch 3.10.0 and cli11 2.6.0, compilation of a test fails:
```
In file included from /usr/include/catch2/catch_test_macros.hpp:11,
                 from /usr/include/catch2/catch_template_test_macros.hpp:20,
                 from /builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/catch.hpp:14,
                 from /builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/app_helper.hpp:15,
                 from /builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp:8:
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp: In function ‘void CATCH2_INTERNAL_TEST_9()’
:
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp:153:5: error: expected ‘,’ or ‘;’ before ‘do
’
  153 |     CHECK(result);
      |     ^~~~~
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp: In function ‘void CATCH2_INTERNAL_TEST_26()
’:
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp:356:5: error: expected ‘,’ or ‘;’ before ‘if
’
  356 |     if(pstring_start > 0) {
      |     ^~
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp:359:7: error: ‘else’ without a previous ‘if’
  359 |     } else {
      |       ^~~~
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp: In function ‘void CATCH2_INTERNAL_TEST_30()
’:
/builddir/build/BUILD/cli11-2.6.0-build/CLI11-2.6.0/tests/FuzzFailTest.cpp:396:5: error: expected ‘,’ or ‘;’ before ‘tr
y’
  396 |     try {
      |     ^~~
[...]

PR #1238, opened 2025-10-22
try adding a test for failure with macos arm precompile

phlptp:


phlptp:
Fixes #1236

phlptp:
test and possibly resolve #1237

PR #1247, opened 2025-10-29
option name formatting in help

### LINKED ISSUE #857: Option to hide the default value in the help text
When I set a flag like this:
```cpp
app.add_flag("--hidden,!--no-hidden", showHiddenFiles, "Show or hide hidden files");
```

the resulting help text looks like this:
```
--hidden, --no-hidden{false}          Show or hide hidden files
```

It would be nice to have the ability to hide the `{false}` part. I did not find such an option in the docs/code. Also using a custom formatter doesn't seem to help, since this behaviour is defined in `std::string Option::get_name(bool positional, bool all_options)`.

phlptp:
Add some controls to manipulate option string formatting, including disabling the default values, disabling default flag values, disabling type names.

Fixes #857


phlptp:
will merge Monday morning, unless there are objections.

PR #1251, opened 2025-11-07
Add reverse_string utility function in cliutils namespace

jamisiveshkumar:
### Summary
Added a small utility function reverse_string() under include/cliutils/custom_util.hpp.

### Details
- Returns the reversed string of given input.
- This is a minimal code-only addition, no documentation or example.
- Code compiles independently with existing includes.

### DIFF
--- a/include/cliutils/custom_util.hpp
+++ b/include/cliutils/custom_util.hpp
@@ -4,9 +4,6 @@
 namespace cliutils {
 
 // Returns the reverse of a given string
-inline std::string reverse_string(const std::string &input) {
-    return std::string(input.rbegin(), input.rend());
-}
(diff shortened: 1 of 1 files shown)


phlptp:
can you give some explanation of the value this adds to CLI11, what issue it is solving?  Assuming it was needed, it would probably be better to add this to stringTools.hpp rather than adding a new namespace and header file

jzakrzewski:
I don't think such a thing belongs to this library because it has nothing to do with the command line / program arguments / etc.

jagerman:
> can you give some explanation of the value this adds to CLI11, what issue it is solving? Assuming it was needed, it would probably be better to add this to stringTools.hpp rather than adding a new namespace and header file

This is AI slop.  Lovely that we have to deal with this crap nowadays, isn't it?

Issue #1254, opened 2025-11-11
[Feature] Add C++20 modules support


iTrooz:
Hello !

I'm trying to use CLI11 with C++20 modules, but I'm getting tons of errors when trying to pass a `CLI::App` through the module's interface. I think they are mostly about the `dummy` variable in TypeTools.hpp being TU-local due to being constexpr.

Full error log is available here: https://pastebin.com/PJYpZhAV, but it's mostly a spam of this:
```
/bin/g++   -std=gnu++23 -fmodules-ts -MD -MT CMakeFiles/math_mod.dir/src/math.ixx.o -MF CMakeFiles/math_mod.dir/src/math.ixx.o.d -fmodules-ts -fmodule-mapper=CMakeFiles/math_mod.dir/src/math.ixx.o.modmap -MD -fdeps-format=p1689r5 -x c++ -o CMakeFiles/math_mod.dir/src/math.ixx.o -c /home/itrooz/tmp3/CppModules/src/math.ixx
In file included from /usr/include/CLI/CLI.hpp:26,
                 from /home/itrooz/tmp3/CppModules/src/math.ixx:3:
/usr/include/CLI/TypeTools.hpp:456:6: error: ‘template<class T1, class T2, class T, typename std::enable_if<std::is_same< <template-parameter-1-1>, <template-parameter-1-2> >::value, CLI::detail::enabler>::type <anonymous> > decltype (CLI::detail::to_string(forward<T>(value))) CLI::detail::checked_to_string(T&&)’ exposes TU-local entity ‘CLI::detail::dummy’
  456 | auto checked_to_string(T &&value) -> decltype(to_string(std::forward<T>(value))) {
      |      ^~~~~~~~~~~~~~~~~
/usr/include/CLI/TypeTools.hpp:39:19: note: ‘CLI::detail::dummy’ declared with internal linkage
   39 | constexpr enabler dummy = {};
      |                   ^~~~~
```

Example program here: https://github.com/iTrooz/CppModules/tree/cli11, compile (with errors) with `CC=/bin/gcc CXX=/bin/g++ cmake -G Ninja -B build && (cd build && ninja)`

I am using gcc 15.2.1, cmake 4.1.2 and latest CLI11 commit (c9921127d72caea846d857e3571dd2ee119b3bc2), on ArchLinux

This problem does not happen on clang, probably because of https://github.com/llvm/llvm-project/issues/112294

phlptp:
[...]

PR #1255, opened 2025-11-14
Module testing

phlptp:
Address issue #1254


CMakeLists.txt
```diff
@@ -65,6 +65,10 @@ endif()
 option(CLI11_WARNINGS_AS_ERRORS "Turn all warnings into errors (for CI)")
 option(CLI11_SINGLE_FILE "Generate a single header file")
 option(CLI11_PRECOMPILED "Generate a precompiled static library instead of a header-only" OFF)
+if(CMAKE_CXX_STANDARD GREATER 19)
+  option(CLI11_MODULE "modify some code to support inclusion of CLI11 in a module" OFF)
+endif()
```
henryiii:
```suggestion
cmake_dependent_option(CLI11_MODULE "Modify some code to support inclusion of CLI11 in a module" OFF "CMAKE_CXX_STANDARD GREATER_EQUAL 20" OFF)
```

tests/module_test/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+project(CLI11-module-test)
+
+include(CTest)
+
+if(CLI11_DIR)
+  set(CMAKE_PREFIX_PATH ${CLI11_DIR})
+endif()
+
+set(CMAKE_CXX_STANDARD 23)
```
henryiii, line 11: This could be 20 as well?

.github/workflows/tests.yml
```diff
@@ -295,6 +295,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - uses: ./.github/actions/quick_cmake
+        with:
+          cmake-version: "4.2"
+      - uses: seanmiddleditch/gha-setup-ninja@master
+      - name: Configure
+        run: cmake -S . -B build -DCLI11_INSTALL_PACKAGE_TESTS=ON -DCLI11_MODULE=ON -DCMAKE_INSTALL_PREFIX=/home/runner/work/install -DCMAKE_CXX_STANDARD=23
```
henryiii: (same)

include/CLI/StringTools.hpp
```diff
@@ -45,7 +45,7 @@ using enums::operator<<;
 namespace detail {
 /// a constant defining an expected max vector size defined to be a big number that could be multiplied by 4 and not
 /// produce overflow for some expected uses
-constexpr int expected_max_vector_size{1 << 29};
+CLI11_MODULE_INLINE constexpr int expected_max_vector_size{1 << 29};
```
henryiii, line 48: By the way, can we not have `inline` here all the time?

phlptp: inline constexpr variables are only available in C++17, so putting inline here doesn't compile on on C++11

.github/workflows/tests.yml
phlptp: I wanted to match the test environment of the issue that triggered this, so that is why it is at gcc 15 and C++23

include/CLI/StringTools.hpp
henryiii: But could it be based on C++ level, then? It would be nice not to have more config options if possible.

phlptp: seems to work if I condition on C++17 and have it be inline or static,  will go with that.

phlptp:
@henryiii  It did work to not have a cmake option (other than for the added test), I will merge Saturday morning unless you have other suggestions.

PR #1263, opened 2025-12-03
Detail namespace

RashidAGP:
for loop in `App_inl.hpp` was changed to range-based for loop with utilizing *span*.


phlptp:
this will only compile on C++20 and above, we can't merge it, until we commit to only supporting C++20 or higher.

Issue #1264, opened 2025-12-08
Allow extra arguments only after positional mark --


LorienLV:
CLI11 version: `2.6.1`

`prefix_command()` is described as the ideal way to use a CLI11-based program as a prefix or wrapper for another application. My understanding is that the intended use case is:

```
./my_app [my_app_args] -- other_app [other_app_args]
```

With `prefix_command()`, this use case works well as long as the arguments before `--` are valid. The issue is that handling unexpected or invalid arguments becomes difficult: CLI11 currently accepts all unrecognized arguments as "extras", even if they appear before the `--` separator. This makes it hard to detect unexpected arguments.

Consider the following example:
```c++
#include <vector>
#include <optional>
#include <iostream>

#include "CLI11.hpp"

struct Options {
    bool flag;
    std::optional<int> value;
    std::optional<std::vector<int>> list;
};

int main(int argc, char *argv[]) {
    CLI::App app;

    Options options;

    app.add_flag("--flag", options.flag, "")->default_val(false);
    app.add_option("--value", options.value, "");
    app.add_option("--list", options.list, "");

    app.prefix_command();

    try {
        CLI11_PARSE(app, argc, argv);
    }
    catch (const CLI::Error &e) {
        std::cerr << "Error when parsing arguments: " << e.what() << "\n";
        return EXIT_FAILURE;
    }

    std::cout << "Extras captured:\n";
    for (const auto &extra : app.remaining()) {
        std::cout << extra << "\n";
    }

    return 0;
}
```

If the user provides valid arguments followed by a `--`, extras correspond strictly to the wrapped command:

```
$ ./my_app --flag --list 2 3 4 --value 1 -- ls -l
Extras captured:
--
ls
-l
```

If the user passes unexpected arguments before the `--`, CLI11 treats them as regular extras and mixes them with the wrapped command:
```
$ ./my_app -u --flag --unsupported --list 2 3 4 --value 1 --wrong -- ls -l
Extras captured:
-u
--unsupported
--wrong
--
ls
-l
```

Here, `-u`, `--unsupported`, and `--wrong` are unexpected arguments that should be errors because they appear before the `--`. But since `prefix_command()` accepts all unknown arguments, the application must manually separate "unexpected extras before `--`" from "the wrapped command and its arguments after `--`".

The use case of a wrapper application becomes even more difficult to implement when it includes subcommands, and only one of those subcommands is intended to be used as the wrapper.

[...]

PR #1266, opened 2025-12-13
prefix_command tests and improvements

phlptp:
adding a PrefixCommandMode option to the prefix_command to allow specification of a separator and catch other errors

Addresses #1264


phlptp:
@henryiii  I will merge this on Wednesday, unless you have objections, or comments.

Issue #1194, opened 2025-08-13
Problem of parsing with additional optional arguments


najibprog:
**Context**

Hi, I'm using CLI11 in my project, and it's a very nice alternative to the Boost parser!

This is my variable parser:

`CLI::App  _cmdLineOptionsCLI;`

I'm using the function `_cmdLineOptionsCLI.parse(argc, argv)` by giving a vector of options.
It may be that the options in argv are not in the options defined by `add_options`, so I added a 

`cmdLineOptionsCLI.allow_extras(true);`

**The problem**

When I have for example `"--nbVar ; 5"` in my argv and are not defined in add_options, `"--nbVar"` is considered by the classifier like a `Classifier::LONG`  and  "5" is considered like a `Classifier::NONE`, so with the first arguments, we pop correctly the option, but with 5 we reach this line of code 

`                posOpt = opt.get();` in `CLI11_INLINE bool App::_parse_positional`

By reaching this line, it fills a variable containing options defined by `add_options`, which is incorrect.

**A solution proposed**

If we add some lines to pop_back the argument and the value if the classifier is `SHORT, LONG, WINDOWS_STYLE`, it solves the problem. We suppose that if we have extra args like this "--nbVar 5", we can remove two elements instead of only one.
Is it possible to add these lines of code like this?  

```
        if(current_type == detail::Classifier::LONG || current_type == detail::Classifier::WINDOWS_STYLE ||
           current_type == detail::Classifier::SHORT ) {
                args.pop_back();
        }

``` 
App_inl.hpp line 2005.

Best regards,
Najib

phlptp:
One way we have gotten around this issue before is by using `validate_positionals()` on the app.  This ensures that each assigned positional matches what is expected.   Doesn't fix all cases but does in many.

The problem with the above approach is that options can have variable numbers of arguments (0,1, 2, etc).   0 and 1 are most common and assumptions about how many there are would break quite a bit of code, so ends up worse than the situation you described.   The only other option would be to develop the logic for a 2 pass parsing in which some unrecognized options are delayed in parsing until the second pass to fill up positionals.   

My recommendation would be to check if validate_positionals would work in your case, which works particularly nicely if the positionals define file paths or something of that sort.

najibprog:
Thank you for your reply ! I have tested using validate_positionals, but I have the same problem. In my case, I think that I will integrate CLI11 inside my project by making this modification and informing the users to use extra arg with type (--, -, /) by having only one element after the option. 

[...]

Issue #986, opened 2024-01-24
Allow for earlier exit with non matching options


levi-de-koning:
During testing, we encountered an unexpected error message. Let me introduce this small sample to explain the error message and how it confused us.

```cpp
CLI::App testApp;
int value;
testApp.add_option("--foo", value);
std::string file;
testApp.add_option("file", file)->check(CLI::ExistingFile);

CLI11_PARSE(testApp);
```
We will use `-bar 5 somefile.txt` for the command line arguments. We can already see this command line is invalid because `-bar` is not a valid option.  When we run this, we get the error:  `file: File does not exist: 5`. To a user, this is a super unexpected error as well. They specified some file.txt as their filename. Now, let's say we remove this validator. The error changes to `The following arguments were not expected: --bar.` This is the actual logical error I would expect for this command line.

The underlying reason for this is that. `--bar` will fail to match any subcommand and is added to the missing/remaining list.  then 5 is parsed next, and this matches the position argument, which now fails the validation. 

So, in conclusion, it would be nice if we could specify that any non-matching argument should throw an error. I don't necessarily think this should be the default, as you would lose the ability to get a list of all options that do not match. but at least being able to configure this would greatly clarify the error messages for cases like this.

phlptp:
can you try this with the `validate_positionals` modifier and see if the error is the same.  My guess is that will do what you want in this case.

levi-de-koning:
> can you try this with the `validate_positionals` modifier and see if the error is the same. My guess is that will do what you want in this case.

Indeed that should work in this example I believe, however if you where to combine it with a positional argument consuming all arguments like https://github.com/CLIUtils/CLI11/blob/main/examples/arg_capture.cpp

it should lead to to the the argument in this case being matched to the final consuming argument if I am understanding this correctly:
```
.validate_positionals(): Specify that positionals should pass validation before matching. Validation is specified through transform, check, and each for an option. If an argument fails validation it is not an error and matching proceeds to the next available positional or extra arguments.
```

phlptp:
It requires positionals to have validation on them, yes.   
In the example given you are using the `remaining` to capture all the arguments after a particular prefix command.  which forces everything after that to be a positional.  

So I am still a little unclear under what scenarios cannot be handled appropriately with some of the different options between the prevalidation, and trigger_on_parse modifiers

levi-de-koning:
[...]

PR #1270, opened 2025-12-20
Extras configuration

phlptp:
Allow more control over how extras are interpreted.  Specifically allowing some assumptions to be made about arguments to unrecognized options.   With `CLI::ExtrasMode::AssumeMultipleArguments` all positional arguments following an unrecognized option will be considered extra as well even if there are open positional arguments.   With `CLI::ExtrasMode::AssumeSingleArgument` a positional argument following an unrecognized option will also be considered unrecognized and go into the `remaining()` bin.


phlptp:
Fixes #986

phlptp:
Fixes #1194

phlptp:
@henryiii will merge Jan 2 unless there are other comments or concerns

PR #1272, opened 2025-12-23
Fix missing impl headers in Spack/Precompiled installs (Fixes #1198)

### LINKED ISSUE #1198: Missing header files in spack installed v2.4.2
I've installed v2.4.2 using spack v0.23.1 and our application includes `CLI/App.hpp`.  It seems that the headers in the `imp` directory are not included in the installation procedure and are not present in our installation directory.

Build log selection:

```
In file included from /home/foo/builds/XpY_9Zdj/5/bar/my_app/packages/Utilities/core/tests/tstCLI11.cc:10:
In file included from /projects/mp_common/sw_env/dev/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder___/linux-rhel8-zen2/gcc-14.2.0/cli11-2.4.2-shhrifkzyllkcdbwiynww7igmvto57ah/share/cmake/CLI11/../../../include/CLI/App.hpp:27:
In file included from /projects/mp_common/sw_env/dev/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder___/linux-rhel8-zen2/gcc-14.2.0/cli11-2.4.2-shhrifkzyllkcdbwiynww7igmvto57ah/share/cmake/CLI11/../../../include/CLI/ConfigFwd.hpp:20:
In file included from /projects/mp_common/sw_env/dev/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder___/linux-rhel8-zen2/gcc-14.2.0/cli11-2.4.2-shhrifkzyllkcdbwiynww7igmvto57ah/share/cmake/CLI11/../../../include/CLI/Error.hpp:21:
/projects/mp_common/sw_env/dev/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder___/linux-rhel8-zen2/gcc-14.2.0/cli11-2.4.2-shhrifkzyllkcdbwiynww7igmvto57ah/share/cmake/CLI11/../../../include/CLI/StringTools.hpp:268:10: fatal error: 'impl/StringTools_inl.hpp' file not found
  268 | #include "impl/StringTools_inl.hpp"  // IWYU pragma: export
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

From the app source file, `tstCLI11.cc`, starting at line 10:
```
#include "CLI/App.hpp"
#include "CLI/Config.hpp"
#include "CLI/Formatter.hpp"
```

Our installation include directory contents:
```
$ ls -al  /projects/mp_common/sw_env/dev/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder___/linux-rhel8-zen2/gcc-14.2.0/cli11-2.4.2-shhrifkzyllkcdbwiynww7igmvto57ah/include/CLI/
total 284
drwxr-sr-x 1 i24 users     0 Aug 14 14:33 .
drwxr-sr-x 1 i24 users     0 Aug 14 14:33 ..
-rw-r--r-- 1 i24 users 58457 May  4  2024 App.hpp
-rw-r--r-- 1 i24 users   749 May  4  2024 Argv.hpp
-rw-r--r-- 1 i24 users   720 May  4  2024 CLI.hpp
-rw-r--r-- 1 i24 users  5917 May  4  2024 ConfigFwd.hpp
-rw-r--r-- 1 i24 users  1743 May  4  2024 Config.hpp
-rw-r--r-- 1 i24 users  1702 May  4  2024 Encoding.hpp
-rw-r--r-- 1 i24 users 16022 May  4  2024 Error.hpp
-rw-r--r-- 1 i24 users  6131 May  4  2024 FormatterFwd.hpp
-rw-r--r-- 1 i24 users   632 May  4  2024 Formatter.hpp
-rw-r--r-- 1 i24 users  3889 May  4  2024 Macros.hpp
-rw-r--r-- 1 i24 users 29545 May  4  2024 Option.hpp
-rw-r--r-- 1 i24 users  1764 May  4  2024 Split.hpp
-rw-r--r-- 1 i24 users 10617 May  4  2024 StringTools.hpp
-rw-r--r-- 1 i24 users  4995 May  4  2024 Timer.hpp
-rw-r--r-- 1 i24 users 71706 May  4  2024 TypeTools.hpp
-rw-r--r-- 1 i24 users 36659 May  4  2024 Validators.hpp
-rw-r--r-- 1 i24 users   458 May  4  2024 Version.hpp
```

Not sure if this is an issue on the spack installation script or in the cli11 cmake scripts.  I do see that the `CLI11_impl_headers` is included in the Cmake target `CLI11` so I don't think that's the issue, but I can't tell.  Any direction here?

dhaneesh646:
**Description**
This PR fixes Issue #1198 where `impl` headers were missing when installing in `CLI11_PRECOMPILED` mode (e.g., via Spack).

**Changes Made:**
1. **Installation Fix:** Removed the conditional check that prevented `impl` headers from being installed when `CLI11_PRECOMPILED` was ON. The public headers include these files, so they must always be present.
2. **Windows Build Fix:** Added `target_link_libraries(CLI11 PUBLIC Shell32)` for Windows builds in precompiled mode. Without this, the build fails on Windows with `LNK2019: unresolved external symbol __imp_CommandLineToArgvW`.

**Verification**
Tested locally on Windows with:
- `cmake .. -DCLI11_PRECOMPILED=ON`
- `cmake --build . --target install`
Verified that `include/CLI/impl/` is now populated correctly.
image_ca10e6


phlptp:
what build system/compiler are you using on Windows? This is not required to be included on Visual studio?   Also the impl headers are not required for the precompiled library.  The only reason they would be needed is if you want to use the library as header only, and still have the option of the precompiled part?

dhaneesh646:
Hi @phlptp, thanks for the review!

Regarding the build system: I am using Visual Studio 2022 (MSVC 19.50) with CMake.
When CLI11_PRECOMPILED is ON, the build fails on Windows without linking Shell32. The specific error is:
error LNK2019: unresolved external symbol __imp_CommandLineToArgvW referenced in function compute_win32_argv.
It seems CommandLineToArgvW is being used for argument parsing on Windows, so the link is necessary.

Regarding the impl headers:
This change fixes the reported Issue #1198. Even when CLI11_PRECOMPILED is used, the public headers (like StringTools.hpp) still contain #include "impl/StringTools_inl.hpp".
If the impl folder is not installed, consumers of the library encounter:
fatal error: 'impl/StringTools_inl.hpp' file not found
(as documented in the linked issue).

This PR ensures those files are present so the public headers remain valid.

phlptp:
I have a few concerns with these changes.   I will look into the shell issue more, I am thus far unable to replicate the issue.   Regarding the installation of the impl headers I am still not clear why this is necessary.   The impl headers are not needed when CLI11_PRECOMPILED is enabled.   The only reason they would be is if the installation is needed for both precompiled and header only in the same installation.  Which is the reason the `CLI11_FULL_INSTALL` was enabled and why future versions of spack installer will include that option by default.   So just installing all the headers all the time is wrong and potentially confusing.

dhaneesh646:
Thanks for the feedback!

**1. Regarding the Windows Shell32 Issue:**
I am consistently reproducing this on **Windows 11 with Visual Studio 2022 (MSVC 19.50)**.
The error happens because `CLI::detail::compute_win32_argv` calls `CommandLineToArgvW`, which requires linking against `Shell32.lib`.

Here is the exact linker error from my build log:
```text
CLI11.lib(Precompile.obj) : error LNK2019: unresolved external symbol __imp_CommandLineToArgvW referenced in function "class std::vector<class std::basic_string<char... > ... compute_win32_argv(void)"

PR #1273, opened 2025-12-25
Fix 'UTF-8 config filenames not supported on Windows'  #912

### LINKED ISSUE #912: UTF8/Unicode config filename not supported on Windows
If I create a config filename with non ascii characters,

`./my_app --print > déjà`

When I call: `./my_app --config déjà`, I got:

```
my_app-cli.exe --config déjà
déjà was not readable (missing?)
```

The0Dev:
This PR should fix #912

### DIFF
--- a/include/CLI/ConfigFwd.hpp
+++ b/include/CLI/ConfigFwd.hpp
@@ -70,7 +70,12 @@ class Config {
 
     /// Parse a config file, throw an error (ParseError:ConfigParseError or FileError) on failure
     CLI11_NODISCARD std::vector<ConfigItem> from_file(const std::string &name) const {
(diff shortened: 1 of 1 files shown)


phlptp:
We are either going to have to put the change inside a conditional compilation block or define `to_path` as an empty operation for compilations without `<filesystem>`

ghost:
I've added a conditional compilation block. I think the fix should cover many users. However, since the --config path is stored as a std::string, it must be encoded in UTF-8. I assume that, even on _older_ C++ standards, the to_path function on Windows should _convert_ from UTF-8 to the current locale or directly to UTF-16 (the overload for the std::ifstream constructor that takes a wide string is non-standard though.)

phlptp (approved):

phlptp:
@all-contributors add @The0Dev  for code

PR #1279, opened 2026-01-03
Add support for C++ modules

mikomikotaishi:
This PR adds full support for C++ modules, enabled with `CLI11_MODULES` and creates a module `cli11`.

Solves #1254, #1255.


phlptp:
how complicated is it to add some tests for this feature, and some documentation on how to use it.

mikomikotaishi:
We can just modify tests/module_test

Where should documentation for this be written?

mikomikotaishi:
I'll update the book pages

mikomikotaishi:
@phlptp OK, I've added information to the book and tests as requested.

mikomikotaishi:
I'm not sure why the Codacy static analysis test is failing, considering the test is C++23 and should detect `<print>`. It also seems to complain about `<string>` though which doesn't make much sense

phlptp:
you can ignore codacy,  @henryiii  needs to adjust the settings on it, I don't have full access to it.

mikomikotaishi:
Is there anything else left to do before we can merge?

phlptp:
The new module test appears to be failing, that will need to be fixed.

mikomikotaishi:
@phlptp OK, whatever the problem was it should be fixed now (it was just a simple typo of writing `-DCLI11_MODULES` which instead should have been `-DCLI11_MODULES=ON`).

tests/module_test/cmodule.ixx
phlptp, line 1: I haven't worked with module creation very much.   Is it valid to assume that if cli11 can be created as an separate module it is guaranteed that the headers (not as a module) can be included in another module?   Basically trying to make sure this test has no remaining value once this PR is merged?

mikomikotaishi: Yes, of course, the headers will not change. Modules are essentially just another means of accessing the library (like class-path and module-path in Java).

mikomikotaishi:
@phlptp I think this should fix the problems. Decided to remove `<print>` from the test and just move back to `<iostream>`.

tests/module_test/cmodule.ixx
phlptp: The question is whether removing this test has any potential negatives?   Is there any value in testing whether the defined module works for cli11, and having a test of where the headers of CLI is included in a different module without exposing the CLI11 data types directly.  Or does the test of CLI11 as a module guarantee no issues in the latter.

phlptp:
still failing the module test

mikomikotaishi:
OK, I managed getting it working on my machine so it's probably something up with the CI configuration, I'll look closer

phlptp:
Getting the CI tests working is a bit of a pain, I might have some time in the weekend to take a closer look if it is still giving you trouble by then.

mikomikotaishi:
If you could help investigate what the problem is it would be highly appreciated. I am able to get the tests passing locally but haven't been able to figure out why it fails on CI.

.github/workflows/tests.yml
```diff
@@ -219,7 +219,7 @@ jobs:
         with:
           submodules: true
       - name: Configure
-        run: cmake -S . -B build -DCLI11_INSTALL_PACKAGE_TESTS=ON -DCMAKE_INSTALL_PREFIX=/home/runner/work/install
+        run: cmake -S . -B build -DCLI11_INSTALL_PACKAGE_TESTS=ON -DCMAKE_INSTALL_PREFIX=/home/runner/work/install -DCMAKE_CXX_STANDARD=23
```
phlptp: why are we changing this test to c++23?

tests/CMakeLists.txt
```diff
@@ -370,6 +370,7 @@ if(CLI11_INSTALL_PACKAGE_TESTS)
       "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}"
       "-DCMAKE_CXX_STANDARD=${CMAKE_CXX_STANDARD}"
       "-DCLI11_DIR=${CMAKE_INSTALL_PREFIX}"
+      "-DCLI11_MODULES=ON"
```
phlptp: I don't think this does anything here

tests/module_test/CMakeLists.txt
```diff
@@ -8,16 +8,16 @@ if(CLI11_DIR)
...
 # Test the CLI11 CMake package config
 find_package(CLI11 2.5 REQUIRED)
 
 # Test the target
 add_executable(module-test module_test.cpp)
 
-target_sources(module-test PUBLIC FILE_SET cmodule TYPE CXX_MODULES FILES cmodule.ixx)
-
-target_link_libraries(module-test CLI11::CLI11)
+target_link_libraries(module-test CLI11::CLI11 CLI11::CLI11_Module)
```
phlptp:
```suggestion
target_link_libraries(module-test CLI11::Module)
```
If I understand right it should not need the CLI11:CLI11 library and should link to the Namespaced version of the module library

.github/workflows/tests.yml
mikomikotaishi: I thought the test always was C++23? The previous CMake had this as C++23.

tests/CMakeLists.txt
mikomikotaishi: I'll remove it

tests/module_test/CMakeLists.txt
mikomikotaishi: I'll apply this change

.github/workflows/tests.yml
phlptp: this test was not, only the previous module test used C++23, which is further down the yaml file.

mikomikotaishi:
@phlptp Do you still think you can help me fix the failing module test? I am able to build and run the test locally, but for whatever reason it fails here, and I do not have much experience using CI.

.github/workflows/tests.yml
```diff
@@ -328,7 +328,7 @@ jobs:
           cmake-version: "4.2"
       - uses: seanmiddleditch/gha-setup-ninja@master
       - name: Configure
-        run: cmake -S . -B build -DCLI11_INSTALL_PACKAGE_TESTS=ON -DCLI11_MODULE_TESTS=ON -DCMAKE_INSTALL_PREFIX=/home/runner/work/install -DCMAKE_CXX_STANDARD=23
+        run: cmake -S . -B build -G Ninja -DCLI11_INSTALL_PACKAGE_TESTS=ON -DCLI11_MODULE_TESTS=ON -DCLI11_MODULES=ON -DCMAKE_INSTALL_PREFIX=/home/runner/work/install
```
phlptp: this one probably should have the -DCMAKE_CXX_STANDARD=23 on it

tests/module_test/CMakeLists.txt
```diff
@@ -9,15 +9,14 @@ if(CLI11_DIR)
...
 # Test the CLI11 CMake package config
 find_package(CLI11 2.5 REQUIRED)
 
 # Test the target
 add_executable(module-test module_test.cpp)
 
-target_sources(module-test PUBLIC FILE_SET cmodule TYPE CXX_MODULES FILES cmodule.ixx)
-
-target_link_libraries(module-test CLI11::CLI11)
+target_link_libraries(module-test CLI11::Module)
```
phlptp:
```suggestion
target_link_libraries(module-test CLI11::CLI11_Module)
```

I think there might be some differences in the naming the in the tests, pretty sure the CLI11_Module target got mapped onto the CLI11 namespace, so we might want to think about what name to use for this target, but lets try this first.

mikomikotaishi: OK, for now we'll keep it (not sure why this happened).

.github/workflows/tests.yml
phlptp:
```suggestion
        run: cmake -S . -B build -G Ninja -DCLI11_INSTALL_PACKAGE_TESTS=ON -DCLI11_MODULE_TESTS=ON -DCLI11_MODULES=ON -DCMAKE_CXX_STANDARD=23 -DCMAKE_INSTALL_PREFIX=/home/runner/work/install
```

mikomikotaishi: OK, thanks

src/modules/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+add_library(CLI11_Module)
+
+target_sources(CLI11_Module PUBLIC FILE_SET CXX_MODULES FILES CLI11.cppm)
+
+target_compile_features(CLI11_Module PUBLIC cxx_std_20)
```
phlptp:
```suggestion
#target_compile_features(CLI11_Module PUBLIC cxx_std_20)
```

src/modules/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+add_library(CLI11_Module)
+
+target_sources(CLI11_Module PUBLIC FILE_SET CXX_MODULES FILES CLI11.cppm)
+
+#target_compile_features(CLI11_Module PUBLIC cxx_std_20)
```
phlptp:
```suggestion
target_compile_features(CLI11_Module PUBLIC cxx_std_23)
```

mikomikotaishi:
I was suggested to try moving `import cli11;` before the `#include` directives - I am sceptical of whether this will fix things, but I'll try it anyway.

Worst case scenario, we go back to using `<iostream>` instead of `<print>`, it doesn't really matter; I just personally dislike `<iostream>`.

phlptp:
I am not entirely sure what to do with an internal compiler error?  Likely something to do with locale but anything is just a guess

mikomikotaishi:
It works for me on Clang 21, so I do think it's a GCC problem most of all.

mikomikotaishi:
@phlptp It seems to be a problem with `#include <cstdlib>`, so I removed that line.

phlptp:
I am going to try to stand up a gcc 15 system this weekend and see if I can make any progress on figuring out what the issue(s) are, still kind of fuzzy on what is going on with this test

mikomikotaishi:
OK, I'm going to try just commenting out `std::println()` and `#include <print>`, and just use `#include <cstdio>` and `std::printf()` for the time being. If this passes, we'll just return to the `#include <print>` problem at a later time.

mikomikotaishi:
The indicated compiler in the tests is GCC 15, though.

phlptp:
I don't think this works with GCC 15.  I got past the internal compiler error for <cstdio>  but it is getting a whole mess of multiple reference errors and repeated definition errors from the template definitions in CLI11.   And a bunch more missing definitions, some of which I can fix by including other headers prior to the import, but I don't think I can fix them all.   

I suspect GCC 15 is just not workable for this.  

What we probably want to do is restore the old test.   Then make a new test that uses clang to verify the new module works on that compiler.

mikomikotaishi:
I don't think there was an actually useful purpose for the old test - it just uses a module with some symbols from the library. Why don't we just adapt the current test to use Clang instead?

phlptp:
The purpose of the previous test was to ensure that CLI11 could be included in a separate module.  Before the test there were a few definitions that triggered multiple definition errors when included in separate module even though nothing from CLI11 was exposed in the module.   Those have been fixed but it remains useful to ensure nothing new comes in.

phlptp:
@mikomikotaishi I think I finally got this working on clang in #1286 I don't think this module will work on GCC yet.  Can you take a look at that one make sure it is the same.

mikomikotaishi:
OK

I think Clang just has better module support than GCC at the present. They recently improved module support in GCC 15 but I still find that Clang has fewer bugs and points of breakage than GCC.

mikomikotaishi:
@phlptp Looks pretty good to me. Should I just close this PR? Also, since we're using C++23 for modules, could we test using `std::println()` instead of `std::cout`?

phlptp:
I moved it back to using `println`.  If you are good with the changes I will go ahead and merge tomorrow unless @henryiii  has anything to add.   Then we can close this one when that gets merged.

mikomikotaishi:
Sounds good to me

phlptp:
I merged #1286.   I think we can close this one now.   Thank you for your work on this.

PR #1280, opened 2026-01-13
Refactor meson build system

mochaaP:
- Bump meson version to 1.3
- Bump default cpp_std (required by Catch2 3.x, still possible for c++11 fallback)
- Move option `single-file-header` and `precompiled` to `mode`
- Make pkgconfig optional
- Refactor tests, fix a bizzare situation where filename contains quote
  or escape characters

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ concurrency:
 
 permissions:
   contents: read
-  
+
 env:
   CTEST_OUTPUT_ON_FAILURE: "1"
 
@@ -191,7 +191,7 @@ jobs:
           pipx install ninja
 
       - name: Configure
-        run: meson setup build-meson . -Dtests=true
+        run: meson setup build-meson . -Dtests=enabled
 
       - name: Build
         run: meson compile -C build-meson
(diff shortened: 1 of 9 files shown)


phlptp:
looks like you may need to put some pragmas to clear the warning in the Timer.hpp  file, to get it to pass the test.  
```c++
#ifdef __GNUC__
#pragma GCC diagnostic push

// 2. Turn off the specific warning (e.g., "-Wunused-variable")
#pragma GCC diagnostic ignored "-Weffc++"
#endif

    /// Division sets the number of cycles to divide by (no graphical change)
    Timer &operator/(std::size_t val) {
        cycles = val;
        return *this;
    }

#ifdef __GNUC__
// 3. Restore the original diagnostic state
#pragma GCC diagnostic pop
#endif

mochaaP:
~~The test was failing during runtime not compile time, should be a race condition in CI because I cannot reproduce this locally.~~

The workdir was wrong.

phlptp (approved):

phlptp:
is this ready for merge?

phlptp:
@all-contributors add @mochaaP for platform

mochaaP:
> is this ready for merge?

yup.

PR #1287, opened 2026-01-23
Fix build rules for Bazel 9

mikael-s-persson:
Fix build rules for Bazel 9

The cc_* rules have to be loaded explicitly from rules_cc now. I tested down to version 7.4.0 and added compatibility constraint (I also checked on main branch, it only works for version >= 7.4.0 too).

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,3 +1,5 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+
 cc_library(
     name = "cli11",
     srcs = glob(["src/**/*.cpp"]),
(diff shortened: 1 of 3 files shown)


phlptp:
@all-contributors add @mikael-s-persson for platform,infra

phlptp:
Thank you,  I had made a few fixes in one of the other branches for testing but I like this better

mikael-s-persson:
Thanks! You might consider a new release and update the central registry (https://registry.bazel.build/modules/cli11).

phlptp:
I am hoping to resolve the module issues and potentially a few others then get another release out soon.

mikael-s-persson:
@phlptp I saw the new version and tested it out on some repos with bazel 9, so I created a PR for a new BCR entry: https://github.com/bazelbuild/bazel-central-registry/pull/7884

Issue #1269, opened 2025-12-19
Help for a fallthrough subcommands does not show upper subcommand help


maxsharabayko:
A help for a subcommand `subc2` does not shown upper level subcommand options.
Consider the following structure
```
app subc1 [OPTIONS1] subc2 [OPTIONS2]
```
where subc2 is a subcommand of subc1.

The following command will show help message for subcommand 1 (subc1) with the list of `[OPTIONS2]`, but will not include `[OPTIONS1]`.
```
app subc1 subc2 --help
```

`subc2 ` has the fallthrough() flag set.

Similar to https://github.com/CLIUtils/CLI11/issues/129

phlptp:
You are saying when the help is called for subc2, the help output is missing the commands from subc1 that should be included by virtue of subc2 having fallthrough enabled?

maxsharabayko:
> You are saying when the help is called for subc2, the help output is missing the commands from subc1 that should be included by virtue of subc2 having fallthrough enabled?

Exactly.

maxsharabayko:
A sample to reproduce the issue:

```c++

#include <CLI/CLI.hpp>
#include <iostream>
#include <string>

int main(int argc, char** argv) {
    CLI::App app{ "CLI11 issue #1269 MRE: fallthrough + nested subcommands" };

    // Level 1 subcommand with its own option (--opt1)
    auto* subc1 = app.add_subcommand("subc1", "First level command");
    int opt1 = 0;
    subc1->add_option("--opt1", opt1, "Option belonging to subc1");

    // Level 2 subcommand under subc1 with fallthrough enabled
    auto* subc2 = subc1->add_subcommand("subc2", "Second level command");
    subc2->fallthrough();  // allow unknown options to fall through to parent
    int opt2 = 0;
    subc2->add_option("--opt2", opt2, "Option belonging to subc2");

    // Make it easy to see that fallthrough parsing works:
    subc2->callback([&]{
        std::cout << "[callback] --opt1=" << opt1 << " --opt2=" << opt2 << "\n";
        });

    try {
        app.parse(argc, argv);
    }
    catch (const CLI::ParseError& e) {
        return app.exit(e);
    }
    return 0;
}
```

Output:

```shell
$> ./sample subc1 subc2 --help
Second level command

./sample subc1 subc2 [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit
          --opt2 INT          Option belonging to subc2
```

Issue #1268, opened 2025-12-19
Help for subcommands of subcommands


maxsharabayko:
A help for a subcommand `subc2` is not shown if `subc1` has a positional option and is not provided for the help message.

Consider the following structure
```
app subc1 ID subc2 [OPTIONS]
```
where subc2 is a subcommand of subc1.

The following command will show help message for subcommand 1 (subc1)
```
app subc1 subc2 --help
```

To see help for subc2 the following must be called:
```
app subc1 5 subc2 --help
```

maxsharabayko:
This issue persists in CLI11 v2.6.2.

maxsharabayko:
Below is the minimum reproductional example.
As you can see, .calling `/cli11_help_bug subc1 subc2 --help` provides help on subc1, not subc2.

```shell
$> ./cli11_help_bug subc1 --help
First subcommand


./cli11_help_bug subc1 [OPTIONS] ID [SUBCOMMAND]


POSITIONALS:
  ID INT REQUIRED             Object ID

OPTIONS:
  -h,     --help              Print this help message and exit

SUBCOMMANDS:
  subc2                       Nested subcommand
$> ./cli11_help_bug subc1 subc2 --help
First subcommand


./cli11_help_bug subc1 [OPTIONS] ID [SUBCOMMAND]


POSITIONALS:
  ID INT REQUIRED             Object ID

OPTIONS:
  -h,     --help              Print this help message and exit

SUBCOMMANDS:
  subc2                       Nested subcommand
```

### Reproduction example

```c++
#include <CLI/CLI.hpp>

int main(int argc, char** argv) {
    CLI::App app{"CLI11 help bug MRE"};

    auto* subc1 = app.add_subcommand("subc1", "First subcommand");

    int id{};
    subc1->add_option("ID", id, "Object ID")->required();

    auto* subc2 = subc1->add_subcommand("subc2", "Nested subcommand");

    bool flag{false};
    subc2->add_flag("--flag", flag, "Example flag");

    CLI11_PARSE(app, argc, argv);

    return 0;
}
```

Adding positional argument:
```
$ ./cli11_help_bug subc1 1 subc2 --help
Nested subcommand


./cli11_help_bug subc1 subc2 [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit
          --flag              Example flag
```

PR #1292, opened 2026-01-29
find options through fallthrough

phlptp:
When searching for options, use fallthrough if set to get the options in the parent


phlptp:
Fixes #1269 
Fixes #1268

phlptp:
@henryiii will merge 2/3 unless you have some suggestions.  
Then will start work on a new release 2.7

PR #1304, opened 2026-03-04
Update some documentation errors, and remove static keyword from pre…

### LINKED ISSUE #1301: [v2.6.2] Error in documentation
The option in the documentation is wrong.

https://github.com/CLIUtils/CLI11/blob/37bb6edc5317e99af72ef48405e65d9ca5218861/book/chapters/installation.md?plain=1#L203

This is the actually used value.

https://github.com/CLIUtils/CLI11/blob/37bb6edc5317e99af72ef48405e65d9ca5218861/tests/CMakeLists.txt#L369

### LINKED ISSUE #1302: [CMake] Why force static library
On Arch Linux at least, dynamic libraries are more common.

Can you remove the keyword `STATIC` from here?

https://github.com/CLIUtils/CLI11/blob/37bb6edc5317e99af72ef48405e65d9ca5218861/src/CMakeLists.txt#L10

Without the keyword, the default is still a static library but users can choose via `BUILD_SHARED_LIBS=(OFF|ON)`.

I maintain [this package](https://aur.archlinux.org/packages/cli11-compiled) for Arch Linux. Packagers would profit.

phlptp:
Update some documentation errors, and remove static keyword from precompiled library definition in CMAKE
Fixes #1302 
Fixes #1301

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -245,6 +245,23 @@ jobs:
         run: ctest --output-on-failure -L Packaging
         working-directory: build
 
+  install-precompiled-shared:
+    name: install tests precompiled shared
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Configure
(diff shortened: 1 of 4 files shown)


phlptp:
@spixmaster would this work for you?

henryiii (approved):

Spixmaster:
Both fixed in one pull request. Indeed, this is great!

Issue #334, opened 2019-10-17
Enhancement of ExistingFileValidator, ExistingDirectoryValidator, ExistingPathValidator with additional checks


chri50:
Hi,

appreciate if the default validators are able to check for empty,non-empty files, directories or symbolic links.
e.g. 
> cli11.add_option("--file", file, "filename")->required()->check(CLI::ExistingFile,CLI::NotEmpty);
> or 
> cli11.add_option("--file", file, "filename")->required()->check(CLI::NonEmptyExistingFile);

thx.

PR #1305, opened 2026-03-07
add a validator for the file size

phlptp:
add an  additional extra Validator that can check file sizes minimum and maximum


phlptp:
Fixes #334

phlptp:
will merge tomorrow morning unless there are additional concerns or comments

PR #1312, opened 2026-03-14
fix: remove en_US.UTF-8 from unicode locale fallback list

### LINKED ISSUE #1141: en_US.UTF-8 should not be check
setting en_US.UTF-8 would create strange behaviour for system with other default locale.

https://github.com/CLIUtils/CLI11/blob/2d792055216baa5ce7eca1d12efeb6180b8c7b5c/include/CLI/impl/Encoding_inl.hpp#L36

omerrbbulut:
## Summary

Remove `en_US.UTF-8` from the fallback locale list used by
`set_unicode_locale()`.

The current fallback list includes:
- `C.UTF-8`
- `en_US.UTF-8`
- `.UTF-8`

This change removes `en_US.UTF-8`, which can introduce unexpected behavior on
systems with a different default locale.

Closes #1141

### DIFF
--- a/include/CLI/impl/Encoding_inl.hpp
+++ b/include/CLI/impl/Encoding_inl.hpp
@@ -33,7 +33,7 @@ namespace detail {
 #if !CLI11_HAS_CODECVT
 /// Attempt to set one of the acceptable unicode locales for conversion
 CLI11_INLINE void set_unicode_locale() {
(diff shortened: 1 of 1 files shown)


phlptp:
@all-contributors add @omerrbbulut  for code

phlptp (approved):

Issue #541, opened 2020-11-25
Config subcommand validators are enforced even when the subcommand is not requested


eddyashton:
If I create a config for an app which contains subcommands, with `default_also = true`, the created config contains the default values of all options for all subcommands. When parsing this config, the `Validator`s for this option are run, even if the subcommand is not triggered. This means the config may not be useable, as the `Validator`s fail for options we weren't planning to use.

When writing the config, is it possible to only include the options on the active subcommand(s)? Or else to avoid parsing/processing options from config on inactive subcommand(s)?

Here's a small repro illustrating the problem, and why we have options with conflicting/unsatisfiable `Validator`s - one subcommand requires that a file is _not_ present, while another requires that it _is_ present.

```c++
#include <CLI11/CLI11.hpp>
#include <fstream>
#include <iostream>

int main(int argc, char** argv)
{
  CLI::App app{"repro"};
  app.set_config("--config", "", "Read config", false);
  app.allow_config_extras(false);

  auto create = app.add_subcommand("create");
  create->configurable();
  std::string create_path = "./my_file.txt";
  create->add_option("--create-path", create_path, "A file which must not exist")
    ->capture_default_str()
    ->check(CLI::NonexistentPath);

  auto load = app.add_subcommand("load");
  load->configurable();
  std::string load_path = "./my_file.txt";
  load->add_option("--load-path", load_path, "A file which must exist")
    ->capture_default_str()
    ->check(CLI::ExistingPath);

  CLI11_PARSE(app, argc, argv);

  std::cout << "Reached end of main!" << std::endl;
  const auto config_file = "./config.ini";
  std::cout << "Writing config to " << config_file << std::endl;
  std::fstream fs;
  fs.open(config_file, std::fstream::out | std::fstream::trunc);
  fs << app.config_to_str(true, false);
  fs.close();

  return 0;
}
```

```
$ ./a.out 
 Reached end of main!
 Writing config to ./config.ini
$ cat ./config.ini 
 create.create-path="./my_file.txt"
 load.load-path="./my_file.txt"
$ ./a.out --config ./config.ini 
 --load-path: Path does not exist: ./my_file.txt
 Run with --help for more information.
$ touch ./my_file.txt
$ ./a.out --config ./config.ini 
 --create-path: Path already exists: ./my_file.txt
 Run with --help for more information.
```

phlptp:
[...]

PR #1314, opened 2026-03-19
more options for config file generation

phlptp:
Add a way to control whether only active subcommand defaults are written to a config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1429,12 +1429,27 @@ configuration files if the `configurable` flag was set on the subcommand. Then
 the use of `[subcommand]` notation will trigger a subcommand and cause it to act
 as if it were on the command line.
 
-To print a configuration file from the passed arguments, use
-`.config_to_str(default_also=false, write_description=false)`, where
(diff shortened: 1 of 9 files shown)


phlptp:
Fixes #541

phlptp:
@henryiii I either need to be given permissions to modify the configuration on codacy for CLI11 or you need to disable the missing include check or I am going to disable all the codacy check.   Right now it is not useful given the number of false positives it is generating.

henryiii:
I disabled the pattern on one of them, I think/hope that will disable all of them on the next run. AFAICT you should have access already, if you sign in with GitHub; I think it just uses the GitHub permissions?

phlptp:
I can see everything but I can't change the settings for the projects as best as I can tell.   On my other projects I have access to a code_patterns tab and settings tab in Codacy where I can change that but on cli11 I can just see issues.   Hopefully that change propagates and it is a non-issue going forward.

phlptp:
I will merge tomorrow morning unless there are objections or comments.

PR #1316, opened 2026-03-21
Infinite recursion bug

### LINKED ISSUE #1315: Infinite recursion in CLI::App::get_option_no_throw
Hello.

I've encountered a fatal bug when using an option group within a subcommand while fallthrough is enabled. Tested on the latest commit.

```C++
#include <filesystem>
#include <CLI/CLI.hpp>

static std::filesystem::path path;

static bool flag = false;

  CLI::App app;

  app.add_flag("--flag,!--no-flag", flag, "Enable a flag")->default_val(false);

  // Disabling this prevents the issue:
  app.fallthrough(true);

  auto *sub = app.add_subcommand("sub", "Execute a subcommand");

  auto group = sub->add_option_group("GROUP");

  // Fails:
  group->add_option("-p,--path", path, "An option")->check(CLI::ExistingFile);

  sub->fallthrough(true);

  return 0;
}
```

phlptp:
fix a bug that led to infinite recursion on subcommand option groups with fallthrough enabled.

FIxes #1315

### DIFF
--- a/README.md
+++ b/README.md
@@ -1077,11 +1077,11 @@ option_groups. These are:
   along with the parent for subcommands with fallthrough.
 - `.get_option_no_throw(name)`: Get an option pointer by option name. This
   function will return a `nullptr` instead of throwing if the option is not
-  available.
+  available. This method will not search parents of option_options or nameless subcommands regardless of fallthrough status 🚧, this behavior is slightly different from `get_option`. 
 - `.get_options(filter)`: Get the list of all defined option pointers (useful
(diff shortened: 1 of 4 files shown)


phlptp:
I got curious how an AI would handle this solution  I gave codex the issue vs what I did without AI.   Discussion #1317 .  I am not entirely sure how I feel about that solution, I was considering going down that route as well, so I don't think it is wrong.   Curious what others think.

Issue #1318, opened 2026-03-23
Module doesn't work


mccakit:
Long opts don't work when I use the module but works when I use the header
```cpp
    App app {"cppbuild — Modern Umka-powered Build System"};
    std::string script_path {};
    std::string toolchain_path {};
    std::string build_dir {};
    std::string install_dir {};
    // Configure Subcommand
    auto config = app.add_subcommand("configure", "Configure the project");
    config->add_option("-S,--script-path", script_path, "Path to build script");
    config->add_option("-T,--toolchain-path", toolchain_path, "Compiler toolchain to use");
    config->add_option("-B,--build-dir", build_dir, "Build directory");
    // Build Subcommand
    auto build = app.add_subcommand("build", "Execute build targets");
    build->add_option("-B,--build-dir", build_dir, "Build directory");
    // Install Subcommand
    auto install = app.add_subcommand("install", "Install build artifacts");
    install->add_option("-B,--build-dir", build_dir, "Source build directory");
    install->add_option("-I,--install-dir", install_dir, "Installation prefix");
    // Scan Subcommand
    auto scan = app.add_subcommand("scan", "Scan for source changes/dependencies");
    scan->add_option("-B,--build-dir", build_dir, "Build directory");
    try
    {
        app.parse(argc, argv);
    }
    catch (const CLI::ParseError &e)
    {
        return app.exit(e);
    }
```

phlptp:
what error are you getting, I added this code to the module test and it seems to work fine

mccakit:
Just unexpected argument error.

phlptp:
can you post the build/run output the contains the error.   I added the code to the modules test and it compiles/ runs fine,  it is possible there are platform differences that we are not testing.

mccakit:
Sure, I'll reply as soon as I can

mccakit:
Here
```
import std;
import fmt;
import cli11;
using CLI::App;
auto main(int argc, char *argv[]) -> int
{
    App app {"cppscan — C++/P1689 scanner"};
    std::string file_path {};
    app.add_option("-f,--file", file_path, "Path to the C++ source file to scan");
    try
    {
        app.parse(argc, argv);
    }
    catch (const CLI::ParseError &e)
    {
        return app.exit(e);
    }
}

```

Compdb command
```
[
{
  "directory": "/home/mccakit/desktop/cppscan/build",
[...]

PR #1321, opened 2026-03-24
test the module with more complex parsing tests

phlptp:
Looking #1318,  adding a more complex test for modules using subcommands and different types of options


henryiii:

Pushed a touch-up: the new options were registered but never parsed (the only invocation was `module-test one`), so the #1318 failure mode (long options rejected in module builds) was not exercised. The test app now prints the parsed values, and three ctest cases pass long options, short options, and an option-group option. Also made `-o,--output` bind the previously unused `output` variable instead of `value`. Verified locally with LLVM clang 22 on macOS: all 5 tests pass.

Issue #1285, opened 2026-01-20
Colorful help


dkaszews:
Python's `argparse` library can produce colorful help messages as of 3.14, which would be a nice thing to have in CLI11 as well.

![Image](https://github.com/user-attachments/assets/9e48aae5-cb1d-49a3-b287-6cc44e7d2eb2)

I think it greatly improves readability, visually breaking up large blocks of text and highlighting positional, short and long options.

ghost:
That would be a nice addition.

Zhou-London:
I am looking into this.
Seems a very feasible improvement. Adding a new subclass of **Formatter**, say, ColorFormatter, injecting some ANSI codes in overrides, should be able to make it.

PR #1328, opened 2026-04-09
Support colored help output

Zhou-London:
This PR adds support for the colored help output regarding the dicuss in issue: #1285 
Functions are added into the `Default Formatter`. When `color_enabled_` is true, ANSI codes are injected into the stream to display various colors. 

Certain formatting logic has been modified in the colored branch (`color_enabled_ == true`):
- Count `visual_length` instead of the actual length, for that the latter one include the ANSI codes, which are not displayed.
- Colorize the header/option/etc. before putting them to the stream.
- The color palette is defined as macros in `Macros.h`

Usage:
```cpp
app.get_formatter()->enable_color();
```

An example and a new test has been added. Readme has been updated as well. All tests are passed on my x86 Linux (Arch)
machine.

ffb22712e2108d43bfcdb8943d82cf5b


Sckab:
please don't quit this PR, it's so good and helpful

Issue #1330, opened 2026-04-17
CLI::CheckedTransformer, throws error on parsing enum with stream operator<< overload and default value defined


OBarronCS:
When using CLI::CheckedTransformer and parsing an enum that has an `operator<<` overload for `ostream`, CLI11 is throwing a `CLI::ConversionError` when a `.default_val` is defined. I am using release v2.6.2.

The full code to trigger this is below, but here is an explanation:

```cpp
app.add_option("--color", color)
    ->default_val(Color::Red)
    ->transform(CLI::CheckedTransformer(STR_TO_COLOR_MAP));
```

The `STR_TO_COLOR_MAP` maps valid input strings to the enum. When no `std::ostream& operator<<(std::ostream& os, Color c)` is defined, this works correctly. When defining the overload (which converts the enum to string representing it), however, a ConversionError is thrown when attempting to parse the default_val.

```c++

#include "CLI11.hpp"
#include <unordered_map>

#define COLOR_LIST \
    X(Red) \
    X(Blue) \
    X(Green)  \

enum class Color {
#define X(name) name,
  COLOR_LIST
#undef X
};

const std::unordered_map<std::string, Color> STR_TO_COLOR_MAP {
#define X(name) {#name, Color::name},
  COLOR_LIST
#undef X
};

const std::unordered_map<Color, std::string> COLOR_TO_STR_MAP {
#define X(name) {Color::name, #name},
  COLOR_LIST
#undef X
};


int main(int argc, char** argv)
{
    CLI::App app{"Example"};

    Color color;

    app.add_option("--color", color)
        ->default_val(Color::Red)
        ->transform(CLI::CheckedTransformer(STR_TO_COLOR_MAP));
    
    CLI11_PARSE(app, argc, argv);
    
    std::cout << static_cast<int>(color) << std::endl;
}

// Comment this to stop CLI::ConversionError from being thrown
std::ostream& operator<<(std::ostream& os, Color c) {
    auto it = COLOR_TO_STR_MAP.find(c);
    if (it != COLOR_TO_STR_MAP.end())
        return os << it->second;

    return os << std::string("Unknown");
}

```

```sh
g++ -std=c++17 main.cpp -o main
```

Note that I also tried using `->default_str("Green")`, (instead of default_val), and although there is no error thrown, the value written to the variable appears to be random (not the expected underlying integer value of 0,1, or 2)

OBarronCS:
[...]

PR #1332, opened 2026-04-18
add test for enumeration conversion

phlptp:
Looking into #1330,

adding a test for the described issue.


oscarpeer-design:
Your tests are failing because variables are uninitialised:
`/__w/CLI11/CLI11/tests/NewParseTest.cpp:687:11: error: variable 'color' is not initialized [cppcoreguidelines-init-variables,-warnings-as-errors]
  687 |     Color color;
      |           ^
/__w/CLI11/CLI11/tests/NewParseTest.cpp:688:11: error: variable 'other_color' is not initialized [cppcoreguidelines-init-variables,-warnings-as-errors]
  688 |     Color other_color;
      |       `

henryiii:
That's allowed, but yes, it's tripping up an optional rule.

Issue #1334, opened 2026-04-21
Dangling References


andrewc18:
Hi,

Line [188](https://github.com/CLIUtils/CLI11/blob/29d8b203987c37eb12577fe85940a7250a385df8/include/CLI/impl/StringTools_inl.hpp#L188) thru [191](https://github.com/CLIUtils/CLI11/blob/29d8b203987c37eb12577fe85940a7250a385df8/include/CLI/impl/StringTools_inl.hpp#L191) are dangling references - bound to a temporary at namespace scope. The C++ standard does not extend the temporary's lifetime (only local `const T&` bindings extend lifetime). The temporary `std::string` is destroyed immediately, leaving these four references dangling.

On RHEL7 with GCC8 and C++17 _Segmentation Fault_ is reproducible using v2.6.2 and this simple code snippet:
```
int main(int argc, char** argv) {
    CLI::App app{"App description"};
    CLI11_PARSE(app, argc, argv);
    return 0;
}
```

Regards.

phlptp:
The differences in how  compilers treat this are fascinating.

PR #1335, opened 2026-04-22
static std::string & initializers

phlptp:
move the static inline const std::string& references to a helper function call


phlptp:
Fixes #1334

phlptp:
Will merge on 4/27 unless objections are raised

PR #1337, opened 2026-04-23
chore: improve support for agentic edits

henryiii:
This should be helpful when using AI. I'm generally torn on if AGENTS.md should be commited, but let's try adding committing it here.

:robot: Assisted-by: OpenCode:Kimi-K2.5 (for the AGENTS.md)
:robot: Assisted-by: Copilot:Claude-Haiku-4.5 (for the copilot CI environment)

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,40 @@
+name: "Copilot Setup Steps"
+
+# Automatically run the setup steps when they are changed to allow for easy validation, and
+# allow manual testing through the repository's "Actions" tab
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+
+permissions: {}
+
+jobs:
+  # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
+
+    # Set the permissions to the lowest permissions possible needed for your steps.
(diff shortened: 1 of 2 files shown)


phlptp:
I don't really have any problem with adding these, but curious what the intention is and what this enables?   I have not looked into this much,

henryiii:
Whenever you start an agentic workflow (OpenCode, Copilot, Pi, Claude Code, Codex, etc), AGENTS.md is loaded into the context window. This provides the LLM a general understanding and specific commands, enabling it to work faster and smarter (especially useful for smaller or local LLMs). Without it, it will have to hunt around and collect this information each run.

The other file is for `@copilot`, Github's built-in agent, so if you trigger it (by mentioning it, assigning something to it, or typing into the agents panel), it will have a few useful things already in the environment when it starts up.

I don't have anything specific to work on with these right now, but it should make it easier if something comes up.

henryiii:
The positives with checking in a AGENTS.md is that I did hand edit it a bit, so it's a little better than the autogenerated one (and OpenCode's `/init` instructions seem to be very good, better than pi/copilot, which are the only other two I've tried). And it's pretty human-readable, too.

The downsides are that it could become out of date (rerunning `/init` will refresh it, keeping the edits).

henryiii:
Also, this was somewhat prompted bit https://github.com/CLIUtils/CLI11/pull/1316#issuecomment-4104346661, Codex should be a bit smarter/faster with this present (unless you already generated a local one - note that most harnesses also have their own version, like CLAUDE.md, but they support the open standard AGENTS.md too).

phlptp (approved):

phlptp:
@henryiii  is this ready for merging

phlptp:
just going to merge this on Monday unless I hear otherwise

PR #1342, opened 2026-05-05
docs: mention Catch2 dependency for tests

zgPigeon:
## Summary

Document that the CMake test build uses Catch2 and may try to download the Catch2 header during configuration if Catch2 is not installed.

## Motivation

This helps users understand configuration failures in restricted-network environments, where downloading from GitHub releases may fail.

## Testing

- Ran `ctest --test-dir out/build/x64-debug --output-on-failure`
- 83 tests passed


phlptp (approved):

phlptp:
@all-contributors add @zgPigeon for docs

PR #1345, opened 2026-05-08
add test and fix for empty transformer description

### LINKED ISSUE #1338: Passing description to transform disables it
I'm not sure if this is an intended behaviour or not but it's at least misleading.

`Option::transform(CLI::Transformer(...), "", "")`

^ Just disables option instead of hiding description.

Example:
```cpp
static void CLITest(int argc, const char** argv)
{
	std::map<std::string, int> numbers_type_map{
		{ "one", 1 },
		{ "two", 2 },
		{ "three", 3 },
	};

	CLI::App app("MyApp");

	int result_number = 0;

	app.add_option("-n,--number", result_number, "My number")
		// This silently disables option
		->transform(CLI::Transformer(numbers_type_map), "", "")
		// This works as expected
		// ->transform(CLI::Transformer(numbers_type_map).name("").description(""))
		;

	app.parse(argc, argv);
}

int main()
{
	const char* test_argv[]{ "MyApp.exe", "--number", "two" };
	CLITest(sizeof(test_argv) / sizeof(test_argv[0]), test_argv);
	return 0;
}
```

phlptp:
Fixes #1338 

The issue was that the incorrect overload was being activated and the transformer itself was getting wrapped in another layer that masked the change.   The solution was to add another overload for the first argument being a Validator object, and updating it with the other arguments.

### DIFF
--- a/include/CLI/Option.hpp
+++ b/include/CLI/Option.hpp
@@ -465,7 +465,7 @@ class Option : public OptionBase<Option> {
     Option *transform(Validator_p validator);
 
     /// Adds a transforming Validator with a built in type name
-    Option *transform(Validator validator, const std::string &transform_name="");
+    Option *transform(Validator validator, const std::string &transform_name = "");
 
     /// Adds a transforming Validator with a built in type name and description
     Option *transform(Validator validator, const std::string &transform_description, const std::string &transform_name);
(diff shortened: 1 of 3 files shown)


phlptp:
will merge on Monday unless I hear of some issues

PR #1351, opened 2026-05-31
Fix/1350 cli existing file issue

### LINKED ISSUE #1350: CLI::ExistingFile
The following:

```
std::filesystem::path filename = "";
  app.add_option("--filename", filename, "Specifies the filename")
      ->check(CLI::ExistingFile);
```

does not seem to detect that file does not exist in the default case (empty).

So one needs to add later on, after parse:

```
if (materialModel.empty())
  return EXIT_FAILURE;
```

is this intended behavior or could this be done already by existingFile default validator?

oscarpeer-design:
**Summary**

`CLI::ExistingFile` did not explicitly handle empty string inputs, leading to inconsistent behavior when an empty filename was provided via the command line. This change makes the behavior explicit by rejecting empty filenames with a` "File name is missing."` validation error.

Fixes #1350

**Test Plan**

- [X] Added regression test coverage for explicit empty string file input
- [X] Verified default (unset option) still passes validation unchanged
- [X] Ran full CLI Test suite
- [X] All new and existing tests pass without modification

### DIFF
--- a/include/CLI/impl/Validators_inl.hpp
+++ b/include/CLI/impl/Validators_inl.hpp
@@ -168,7 +168,10 @@ CLI11_INLINE path_type check_path(const char *file) noexcept {
 #endif
 
 CLI11_INLINE ExistingFileValidator::ExistingFileValidator() : Validator("FILE") {
-    func_ = [](std::string &filename) {
+    func_ = [](std::string &filename) -> std::string {
+        if (filename.empty())
+            return std::string("File name is missing.");
+
         auto path_result = check_path(filename.c_str());
         if(path_result == path_type::nonexistent) {
             return "File does not exist: " + filename;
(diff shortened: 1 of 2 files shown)


ferdymercury (approved):
Thanks a lot for the fix.

Maybe, a new test case could be added for completeness using std::filesystem::path as input

oscarpeer-design:
Alright I added a new test using std::filesystem::path as input. Let me know if you need any other fixes.

phlptp:
@all-contributors add @oscarpeer-design for code

phlptp (approved):

phlptp:
Thanks,  anything more that should be added or is planning in this PR, otherwise I will merge it.  Will also fix the merge token  in the .allcontriborsrc in a separate PR and make sure that gets resolved.

phlptp:
@all-contributors add @oscarpeer-design for code

Issue #1357, opened 2026-06-11
CLI11 review for bugs, performance, simplifications, and modernizations (Claude Fable)


henryiii:
:robot: _AI text below_ :robot:

Spot-checks confirm the top findings are real. Here's the synthesized review.

# CLI11 Review

Five parallel review agents covered the parsing core, Option/TypeTools, string/encoding utilities, Config/Formatter, and validators/build system. Each finding below was verified by tracing the code (the agents compiled repro programs for most of the bugs, and I independently re-checked the top three). I've grouped them by impact.

## High-severity bugs

**1. `ignore_case()` + `ignore_underscore()` together break subcommand matching entirely** — `App_inl.hpp:991-1009`. In `check_name_detail`, the underscore step computes `local_name = detail::remove_underscore(name_)`, but the case step then recomputes from scratch with `local_name = detail::to_lower(name_)`, throwing away the underscore removal. A subcommand named `my_sub` with both flags set matches *nothing* — not even its exact spelling (`name_to_check` had its underscores removed, but `local_name` kept them). The alias loop just below chains the transforms correctly, which highlights the fix: `local_name = detail::to_lower(local_name);`.

**2. `Option::clear()` leaves stale `proc_results_`** — `Option.hpp:398-401`. `clear()` resets `results_` and the state machine but not `proc_results_`. The fast path in `results(T&)` (Option.hpp:725-726) prefers `proc_results_` whenever it's non-empty, so after `clear()` + `add_result("7")`, an option that previously had a processed result (e.g. Sum policy producing `"5"`) returns the old **5** instead of 7. Fix: add `proc_results_.clear();`.

**3. Non-codecvt `narrow_impl`/`widen_impl` ignore `str_size`** — `Encoding_inl.hpp:62-122`. Both do `(void)str_size;` and let `wcsrtombs`/`mbsrtowcs` run to the NUL terminator. That's fine for the `std::wstring` overloads, but the `narrow(const wchar_t*, size_t)` and C++17 `string_view` overloads can convert past the requested length (or read out of bounds on a non-terminated view), and embedded NULs truncate. This matters more over time: under C++26 the codecvt path is disabled (`Macros.hpp:108`), making this the default code path. Fix: copy to a local string to guarantee termination at `str_size`, or loop with `wcrtomb`/`mbrtowc` bounded by `str_size`.

[...]

PR #1358, opened 2026-06-11
fix: chain transforms in check_name_detail so ignore_case+ignore_underscore both apply

henryiii:

## Bug

When both `ignore_underscore()` and `ignore_case()` were set on a subcommand
whose name contained an underscore, `App::check_name_detail` silently discarded
the underscore-removal step for the app's own name.

**Root cause** (`include/CLI/impl/App_inl.hpp`, ~line 998):

```cpp
if(ignore_underscore_) {
    local_name = detail::remove_underscore(name_);  // stores result in local_name
    name_to_check = detail::remove_underscore(name_to_check);
}
if(ignore_case_) {
    local_name = detail::to_lower(name_);           // BUG: reads raw name_, not local_name
    name_to_check = detail::to_lower(name_to_check);
}
```

The alias loop just below chains the transforms correctly on `les`, but the
primary name path did not.

**Effect:** With both flags set, NO spelling matched — not even the exact name.
`add_subcommand("my_sub")->ignore_case()->ignore_underscore()` followed by
parsing `my_sub` threw `ExtrasError`; `mysub`, `MY_SUB`, `MySub` all failed too.

## Fix

One-character change: `detail::to_lower(name_)` → `detail::to_lower(local_name)`,
so the two transforms compose correctly.

## Test

Added `SubcommandIgnoreCaseAndUnderscoreCombined` in `tests/SubcommandTest.cpp`
which verifies that all four spellings (`my_sub`, `mysub`, `MY_SUB`, `MySub`)
match and that an unrelated name still fails.

All 549 assertions in 122 test cases pass.

Part of #1357

### DIFF
--- a/include/CLI/impl/App_inl.hpp
+++ b/include/CLI/impl/App_inl.hpp
@@ -995,7 +995,7 @@ CLI11_NODISCARD CLI11_INLINE App::NameMatch App::check_name_detail(std::string n
         name_to_check = detail::remove_underscore(name_to_check);
     }
     if(ignore_case_) {
-        local_name = detail::to_lower(name_);
+        local_name = detail::to_lower(local_name);
         name_to_check = detail::to_lower(name_to_check);
     }
 
--- a/tests/SubcommandTest.cpp
+++ b/tests/SubcommandTest.cpp
@@ -1840,6 +1840,40 @@ TEST_CASE_METHOD(TApp, "SubcommandAliasIgnoreCaseUnderscore", "[subcom]") {
     CHECK_THROWS_AS(run(), CLI::ExtrasError);
 }
 
+// Regression test: ignore_case() + ignore_underscore() together on a subcommand
+// whose name contains an underscore — all spellings should match (GH #1357)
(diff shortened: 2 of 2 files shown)


phlptp:
@henryiii how did you want to handle these AI generated PR's Some look good some I am somewhat concerned by,  did you want to review them and work through them?  For example this one is probably a legitimate bug

henryiii:
Yes, when I take them out of draft that means I've reviewed it and think it looks fine, you don't need to review until they come out of draft. By legitimate bug, you mean, this is a fix, or introducing a bug?

In general, I've been very happy with the quality of these, I've run on about 30 different repositories. Though most of the other ones were python, curious to see how it does here.

henryiii:
I was running a lot of these (https://hackmd.io/@henryiii/r1I96UObfg) because I thought I only had access to Fable for about 12 days. Turned out it was more like two or three days. https://www.anthropic.com/news/fable-mythos-access

phlptp:
I can't speak to all of them yet, but definitely some are legitimate bugs,  probably not ones that would come up all that often like this one but legitimate fixes.   For example this one looks fine to merge.

henryiii:
This is not a one character change. Maybe it's a one token change? :) Looks fine.

PR #1359, opened 2026-06-11
fix(encoding): bound non-codecvt narrow/widen by str_size

henryiii:

## Bug

In `include/CLI/impl/Encoding_inl.hpp`, the non-codecvt branches of `narrow_impl`/`widen_impl` did `(void)str_size;` and relied on `std::wcsrtombs`/`std::mbsrtowcs` converting up to a NUL terminator. That is fine for the `std::wstring`/null-terminated overloads, but wrong for the size-taking overloads (`narrow(const wchar_t *, std::size_t)`, the C++17 `narrow(std::wstring_view)` / `widen(std::string_view)`, etc.):

- A view into the middle of a larger buffer converted the entire remainder instead of `str_size` characters.
- A view over a non-NUL-terminated buffer read out of bounds (UB).
- Embedded NULs truncated the result.

The codecvt path already handled all of this correctly via `to_bytes(str, str + str_size)`. This non-codecvt path becomes the **default under C++26** (`CLI11_HAS_CODECVT` is 0 when `CLI11_CPP26` is set, per `include/CLI/Macros.hpp`), so it matters increasingly.

## Fix

In the non-codecvt branches, copy the input into a local, NUL-terminated buffer of exactly `str_size` characters and convert from that, so the conversion is bounded by `str_size` and never reads past the buffer. The existing locale `scope_guard` behavior and error reporting are preserved (offsets now reported relative to the local copy). C++11-compatible. Embedded NULs within the first `str_size` characters still terminate the C-style conversion early; this is documented in a comment as a best-effort limitation.

## Testing (dual-path)

Added `Encoding: WidenSize` / `Encoding: NarrowSize` regression tests in `tests/EncodingTest.cpp` that:
- convert a prefix of a longer buffer (e.g. `widen(longer.c_str(), 6)` / `narrow(longer.c_str(), 6)`) and assert only the prefix is converted;
- convert a non-NUL-terminated buffer of exactly `str_size` chars (would read OOB under the old code).

Both the default (codecvt) build and a `-DCMAKE_CXX_FLAGS=-DCLI11_HAS_CODECVT=0` build were configured and run; `EncodingTest` passes in both. The macro override was confirmed to select the non-codecvt branch (preprocessed output uses `wcsrtombs`/`mbsrtowcs` rather than `to_bytes`).

Part of #1357

### DIFF
--- a/include/CLI/impl/Encoding_inl.hpp
+++ b/include/CLI/impl/Encoding_inl.hpp
@@ -69,9 +69,15 @@ CLI11_INLINE std::string narrow_impl(const wchar_t *str, std::size_t str_size) {
 
 #endif  // _WIN32
 #else   // CLI11_HAS_CODECVT
-    (void)str_size;
+    // Copy into a local, NUL-terminated buffer so the conversion is bounded by str_size: a view into the
+    // middle of a larger buffer or a non-NUL-terminated buffer must not be read past str_size. Note that an
+    // embedded NUL inside the first str_size characters will still terminate std::wcsrtombs early; this matches
+    // the behavior of C-style narrow conversion and is treated as a best-effort limitation.
+    std::wstring input(str, str_size);
+    const wchar_t *src = input.c_str();
+    const wchar_t *it = src;
+
     std::mbstate_t state = std::mbstate_t();
-    const wchar_t *it = str;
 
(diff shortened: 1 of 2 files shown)


henryiii:
I'm not on Windows, but the tests pass, so will undraft.

PR #1367, opened 2026-06-11
chore: modernization and portability fixes

henryiii:

## Summary

A collection of modernization and portability fixes verified against the actual source. All 24 tests pass; both header-only and precompiled (`CLI11_PRECOMPILED=ON`) builds succeed.

### Changes

1. **Missing standard includes** — Added `<cerrno>`, `<cstdlib>`, `<cctype>` to `TypeTools.hpp` (uses `errno`, `ERANGE`, `EINVAL`, `std::strtoull/strtoll`, `std::isspace`); `<cerrno>` to `Option_inl.hpp` (uses `errno`); `<cstdint>` to `FormatterFwd.hpp` (uses `std::uint8_t`); qualified unqualified `uint8_t` → `std::uint8_t` in `ConfigFwd.hpp` (which already includes `<cstdint>`).

2. **`CLI::Number` missing from module build** — Added `CLI11_MODULE_INLINE` to the `Number` declaration in `ExtraValidators.hpp` (matching siblings like `ValidIPV4`, `NonNegativeNumber`); added `using CLI::Number;` to `src/modules/CLI11.cppm` export list.

3. **Intel Classic compiler exclusion removed** — Dropped `!defined(__INTEL_COMPILER)` from the C++ standard version detection guard in `Macros.hpp` (and the related occurrence in the `CLI11_NODISCARD` guard). Intel Classic (icc) is discontinued; its successor icx identifies as Clang and is unaffected.

4. **Inconsistent enable guards** — The outer guard in `ExtraValidators.hpp` tested `== 1`; the inner guard tested `!= 0`. Unified to `!= 0` throughout.

5. **`get_environment_value` missing `CLI11_INLINE`** — The only definition in `StringTools_inl.hpp` without the macro; added it.

6. **CMake fixes** — `VERSION_GREATER 3.19` → `VERSION_GREATER_EQUAL 3.19` (feature works FROM 3.19, not after); removed `-D` prefixes from all `target_compile_definitions` arguments in `src/CMakeLists.txt` (modern CMake style).

7. **Warning-list cleanup** — Removed Objective-C-only diagnostics (`-Wimplicit-atomic-properties`, `-Wstrict-selector-match`, `-Wundeclared-selector`) from both `cmake/CLI11Warnings.cmake` and `meson.build`; removed the duplicate `-Wshadow` from the Clang append in `CLI11Warnings.cmake` (it is already in the base `unix-warnings` list).

### Observation (not changed, flagged for maintainers)

`meson.build` unconditionally sets `-DCLI11_ENABLE_EXTRA_VALIDATORS=1` in `precompiled` build mode, while the CMake build defaults it off. These two build systems will produce different default behavior for the same library configuration. No change made here — this is a pre-existing inconsistency that warrants a deliberate decision.

Part of #1357

## Test plan

- [x] `cmake -S . -B build -DCMAKE_BUILD_TYPE=Debug -DCLI11_BUILD_EXAMPLES=OFF && cmake --build build -j4 && (cd build && ctest --output-on-failure -j4)` — 24/24 tests passed
- [x] `cmake -S . -B build-pre -DCLI11_PRECOMPILED=ON -DCLI11_BUILD_TESTS=OFF -DCLI11_BUILD_EXAMPLES=OFF && cmake --build build-pre -j4` — precompiled mode builds successfully
- [x] `prek -a --quiet` passes (clang-format and cmake-format auto-fixes applied in commit)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmake/CLI11Warnings.cmake
+++ b/cmake/CLI11Warnings.cmake
@@ -6,16 +6,7 @@ set(unix-warnings -Wall -Wextra -pedantic -Wshadow -Wsign-conversion -Wswitch-en
 # Clang warnings
 # -Wfloat-equal could be added with Catch::literals and _a usage
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  list(
-    APPEND
-    unix-warnings
-    -Wcast-align
-    -Wimplicit-atomic-properties
-    -Wmissing-declarations
-    -Woverlength-strings
-    -Wshadow
-    -Wstrict-selector-match
-    -Wundeclared-selector)
+  list(APPEND unix-warnings -Wcast-align -Wmissing-declarations -Woverlength-strings)
   # -Wunreachable-code Doesn't work on Clang 3.4
 endif()
 
--- a/include/CLI/ConfigFwd.hpp
+++ b/include/CLI/ConfigFwd.hpp
@@ -115,7 +115,7 @@ class ConfigBase : public Config {
     /// the character to use around single characters and literal strings
     char literalQuote = '\'';
     /// the maximum number of layers to allow
-    uint8_t maximumLayers{255};
+    std::uint8_t maximumLayers{255};
     /// the separator used to separator parent layers
     char parentSeparatorChar{'.'};
     /// comment default values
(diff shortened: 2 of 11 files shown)


henryiii:
Do we want to do "Intel Classic compiler exclusion removed"? We aren't testing on it, so tempted to say yes.

Edit: dropped that one as it was small.

Issue #1258, opened 2025-11-22
default_val validate fails if ostream operator differs from `CheckedTransformer` 'CLI::ConversionError' (regression?)


huweiATgithub:
## Environment

- **CLI11 Version:** 2.6.1

I was using `2.5.0` which does not have problem.

## Minimal Reproduction

```cpp
#include <CLI/CLI.hpp>
#include <iostream>
#include <map>

enum class Color { kRed, kBlue };

// operator<< outputs full enum name (standard practice)
std::ostream& operator<<(std::ostream& os, Color c) {
  return os << (c == Color::kRed ? "kRed" : "kBlue");
}

int main() {
  // Map uses simplified, user-friendly keys
  std::map<std::string, Color> color_map = {
    {"red", Color::kRed},    // User types "red"
    {"blue", Color::kBlue}   // User types "blue"
  };

  CLI::App app;
  Color color = Color::kRed;

  app.add_option("--color", color)
      ->transform(CLI::CheckedTransformer(color_map, CLI::ignore_case))
      ->default_val(Color::kRed);  // BUG: Validates "kRed" against {"red", "blue"}

  app.parse("");  // Should use default
  
  return 0;
}
```

### Output

The `val` passed to `default_val()` has been serialized first and compared with keys of `CheckedTransformer`.

```
ValidationError:
   --color: Check kRed value in {red->kRed,blue->kBlue} OR {kRed,kBlue} FAILED
```

### Expected Behavior

> Generate the default string from a value and validate that the value is also valid. For options that assign directly to a value type the value in that type is also updated. Value must be convertible to a string(one of known types or have a stream operator).

From my point of view, the value should be checked against the values of `CheckedTransformer`. But currently, it is serialized first and checked with key with `CheckedTransformer`.
Although the Exception messages says `OR {kRed,kBlue}`, it turns out the values are not being checked with.
 

## Why This Is a Problem

This breaks the common pattern of:
- User-friendly CLI input (short names: "red", "blue")
- Full enum names in code (prefixed: `kRed`, `kBlue`)

## Current Workarounds

Use `capture_default_str` doesn't seems to have problem.

phlptp:
In this issue are you expecting "kRed" and "kBlue" as strings to work as well?  in #1261 I fixed the error in this case which was the default_val conversion.  That conversion would not use the ostream method any more, which means a string of "kBlue" and "kRed" would not be recognized and should be included in the defined map in the transformer.

[...]

PR #1376, opened 2026-07-01
fix: do not run checks against default values (#1375)

### LINKED ISSUE #1375: add_option(...)->check(...)->default_val(...) throws immediately instead of when parsing
### MWE:
```cpp

	int someNumber;
	app.add_option("someNumber,-n,--someNumber", "Some number")
		->check(CLI::PositiveNumber)
		->default_val(0);
	CLI11_PARSE(app, argc, argv);
	return 0;
}
```
### Expected behaviour:
The exception thrown by `check(CLI::PositiveNumber)` is caught by the `catch` block in the `CLI11_PARSE` macro.

### Actual behaviour:
An uncaught exception occurs:
```
terminate called after throwing an instance of 'CLI::ValidationError'
  what():  --someNumber: Value 0 not in range [2.22507e-308 - 1.79769e+308]
```

With the current behaviour, I'd have to wrap `app.add_option(...)->check(...)->default_val(...)` into its own separate `try {...} catch` block, which doesn't seem ideal. Is this intended behaviour?

henryiii:

Fixes #1375.

## Problem

`add_option(...)->check(...)->default_val(...)` throws immediately during app **setup**, not during parsing:

```cpp
app.add_option("someNumber,-n,--someNumber", "Some number")
    ->check(CLI::PositiveNumber)
    ->default_val(0);   // terminate called after throwing 'CLI::ValidationError'
```

`default_val()` validated the default against every registered validator, so a default that fails a `check()` threw right there — outside the `CLI11_PARSE` try/catch, terminating the program. The behavior was also order-dependent: it only threw when the `check()` was registered *before* `default_val()`.

## Fix

Thread a `transform_only` flag through `run_callback` → `_validate_results` → `_validate`. When validating a default value, only **modifying** validators (transforms) are applied; non-modifying **checks** are skipped. This leans on the existing distinction that `check()` marks its validator `non_modifying()` while `transform()` does not.

Consequences:

- Transforms still run against defaults, so enum / `CheckedTransformer` defaults still convert correctly (preserves the #1258 fix).
- Conversion still runs for bound values, so a genuinely un-convertible default still raises `ConversionError` at config time.
- Pure checks no longer throw on a default, so the case above is accepted; the check still applies to values parsed from the command line and is caught by `CLI11_PARSE`.
- Behavior is now independent of `check()` / `default_val()` ordering.

## Behavior change

`default_val()` no longer validates a default against `check()`s. One existing assertion in `NumericTypeTest` (a bad default caught by `check(CLI::Number)` at config time) was updated to reflect this, with added coverage that the check still fires on parsed input.

## Tests

- Added `defaultValFailingCheck` regression test (mirrors the issue MWE); confirmed it failed before the fix.
- Updated `doubleVectorFunctionRunCallbackOnDefault`.
- Documented the behavior in `README.md` and `book/chapters/options.md`.
- Full suite passes (82/82); `prek` clean.


RL-S:
I think the `default_val` should still be checked against the registered validators - just not immediately, but during parsing. Not checking them at all would not be an improvement.

henryiii:
I was experimenting with this - actually registering deferred evaluations and running them during parsing would be a much bigger change. You control the default, though (and can manually call the validator), so I'm not sure if we should require the default be able to validate. It would mean you could set a default that can't be passed, but sometimes you might want that. (I'm a little surprised 0 is not a positive number, I'd have said it was both).

phlptp:
There is also a nonnegative validator that includes 0 if that is better suited to your situation

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

RL-S:
In my case, the default value is a subdirectory of the working directory. That's a runtime value, so it requires validation. Passing the option allows changing the subdirectory's name, so I do require all cases.

(Including zero is typically called non-negative, which also has a builtin validator)

PR #1379, opened 2026-07-04
fix: double indentation in help message

AverageBeerEnjoyer:
In some cases using `indent_block` after `streamOutAsParagraph` results in double indentation
due to the addition of one more prefix after the last line in function `streamOutAsParagraph`

to solve this, i moved prefix addition to the beginning of the loop iteration

``` c++
#include "CLI11.hpp"

    CLI::App app("test description");

    auto g = app.add_option_group("group", "common group");

    auto g1 = g->add_option_group("subgroup 1", "subgroup 1 description");
    g1->add_option("-a", "option a");

    auto g2 = g->add_option_group("subgroup 2", "subgroup 2 description");

    CLI11_PARSE(app, argc, argv);
}
```
this code produces double indentation before `[Option Group: subgroup 1]`
```
$ ./test-old -h
test description


./test-old [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit
[Option Group: group]
  common group

    [Option Group: subgroup 1]
    subgroup 1 description


    OPTIONS:
      -a                          option a
  [Option Group: subgroup 2]
    subgroup 2 description

    $
```

after fix
```
$ ./test -h
test description


./test [OPTIONS]


OPTIONS:
  -h,     --help              Print this help message and exit
[Option Group: group]
  common group

  [Option Group: subgroup 1]
    subgroup 1 description


    OPTIONS:
      -a                          option a
  [Option Group: subgroup 2]
    subgroup 2 description

$
```

### DIFF
--- a/include/CLI/impl/StringTools_inl.hpp
+++ b/include/CLI/impl/StringTools_inl.hpp
@@ -629,15 +629,20 @@ CLI11_INLINE std::ostream &streamOutAsParagraph(std::ostream &out,
                                                 std::size_t paragraphWidth,
                                                 const std::string &linePrefix,
                                                 bool skipPrefixOnFirstLine) {
-    if(!skipPrefixOnFirstLine)
-        out << linePrefix;  // First line prefix
-
     std::istringstream lss(text);
     std::string line = "";
+    bool first_iter = true;
     while(std::getline(lss, line)) {
         std::istringstream iss(line);
         std::string word = "";
         std::size_t charsWritten = 0;
+
+        if(!first_iter || !skipPrefixOnFirstLine) {
+            out << linePrefix;
+        }
+
+        if(first_iter)
+            first_iter = false;
 
         while(iss >> word) {
             if(charsWritten > 0 && (word.length() + 1 + charsWritten > paragraphWidth)) {
@@ -654,7 +659,7 @@ CLI11_INLINE std::ostream &streamOutAsParagraph(std::ostream &out,
         }
 
         if(!lss.eof())
-            out << '\n' << linePrefix;
+            out << '\n';
     }
     return out;
 }
--- a/include/CLI/impl/StringTools_inl.hpp
+++ b/include/CLI/impl/StringTools_inl.hpp
@@ -631,18 +631,14 @@ CLI11_INLINE std::ostream &streamOutAsParagraph(std::ostream &out,
                                                 bool skipPrefixOnFirstLine) {
     std::istringstream lss(text);
     std::string line = "";
-    bool first_iter = true;
     while(std::getline(lss, line)) {
         std::istringstream iss(line);
         std::string word = "";
         std::size_t charsWritten = 0;
 
-        if(!first_iter || !skipPrefixOnFirstLine) {
+        if(!skipPrefixOnFirstLine)
             out << linePrefix;
-        }
-
-        if(first_iter)
-            first_iter = false;
+        skipPrefixOnFirstLine = false;  // subsequent lines always get the prefix
 
         while(iss >> word) {
             if(charsWritten > 0 && (word.length() + 1 + charsWritten > paragraphWidth)) {
--- a/tests/HelpersTest.cpp
+++ b/tests/HelpersTest.cpp
@@ -218,21 +218,21 @@ TEST_CASE("StringTools: Modify3", "[helpers]") {
 }
 
 TEST_CASE("StringTools: streamOutAsParagraph", "[helpers]") {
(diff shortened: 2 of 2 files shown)


henryiii:
Claude review below. I pushed an update based on that.

My simplified form produces byte-identical output to the PR. I now have everything for the review.

---

# Code Review: PR #1379 — help message: fix double indentation

## Overview
When a nested option group's description was rendered, `streamOutAsParagraph` emitted a **trailing line prefix** after the last line. The parent group then re-indented that dangling prefix via `indent_block`, producing a doubly-indented `[Option Group: ...]` header. The fix moves prefix emission to the **start** of each line iteration and drops the trailing `linePrefix` after the final `'\n'`, so no dangling prefix is left behind.

**Verdict: the fix is correct and does what it claims.** I built the branch, reproduced the PR's exact scenario (4-space → 2-space header), and diffed old vs. new behavior across edge cases. All 320 `HelpTest` and 86 `FormatterTest` assertions still pass.

Old vs. new behavior (`prefix=">>"`, `\n` shown literally):

| input              | old                    | new                  |
| ------------------ | ---------------------- | -------------------- |
| `""`               | `>>`                   | *(nothing)*          |
| `"hello\n"`        | `>>hello\n>>`          | `>>hello\n`          |
| `"line1\nline2\n"` | `>>line1\n>>line2\n>>` | `>>line1\n>>line2\n` |

The dangling `>>` is exactly the double-indent source, and it's gone. The empty-string case changing from a bare prefix to nothing is a harmless improvement (no caller depends on an indent-only line).

## Code quality & style

- **Snake_case naming is inconsistent with local style.** `first_iter` doesn't match the surrounding camelCase (`charsWritten`, `skipPrefixOnFirstLine`, `paragraphWidth`, `linePrefix`). Prefer `firstLine`/`firstIter`.

- **The extra bool and two-statement dance are avoidable.** The added logic:
  ```cpp
  bool first_iter = true;
  ...
  if(!first_iter || !skipPrefixOnFirstLine) { out << linePrefix; }
  if(first_iter) first_iter = false;
  ```
  can reuse the existing `skipPrefixOnFirstLine` parameter as the flag, dropping a variable and the double negation:
  ```cpp
  if(!skipPrefixOnFirstLine)
      out << linePrefix;
  skipPrefixOnFirstLine = false;  // subsequent lines always get the prefix
  ```
  I verified this produces byte-identical output to the PR across all the cases above. It's smaller and reads more directly. (Note: it mutates a by-value parameter, which is fine and idiomatic here.)

- Minor: brace usage is inconsistent — `if(!first_iter ...)` uses braces around a single statement, `if(first_iter) first_iter = false;` doesn't. Not important if the simplification above is taken.

## Test coverage — main gap
- **No regression test is included.** The project's own guidance (CLAUDE.md: *"When fixing bugs, add the regression test first and confirm it fails without the fix"*) applies here, and there is currently **zero** direct coverage of `streamOutAsParagraph` and no `HelpTest` asserting nested-group indentation. This bug slipped through precisely because of that gap. Please add either:
  - a focused unit test on `detail::streamOutAsParagraph` (trailing-newline and multi-line cases — these would have failed pre-fix), or
  - a `HelpTest` case with a sub-option-group whose parent has a description, asserting the `  [Option Group: ...]` header has single indentation.

## Correctness / risk
- Low risk. The inner word-wrap path (`out << '\n' << linePrefix`) is unchanged; only per-line and trailing prefix emission moved. The paragraph branch in `Formatter::make_expanded` (and the option/positional description callers) benefit directly; footer callers use an empty prefix so are unaffected.
- One observation, **out of scope for this PR**: blank separator lines still carry indentation (e.g. `  ` after a group description), because `indent_block` prefixes blank lines. This PR doesn't introduce or worsen it — noting only so it isn't mistaken for a remaining regression.

## Recommendation
Approve after: (1) adding a regression test, and ideally (2) the small simplification + naming fix. The core change is sound.

AverageBeerEnjoyer:
Should this warning be fixed?
```
/home/runner/work/CLI11/CLI11/tests/HelpersTest.cpp:221:82: error: default argument specified for lambda parameter [-Werror=c++14-extensions]
  221 |     auto paragraph = [](const std::string &text, const std::string &prefix, bool skipFirst = false) {
      |                                                                             ~~~~~^~~~~~~~~~~~~~~~~
```

henryiii:

Yes — fixed in 6f3e872. Lambda default arguments are C++14, so the test lambda now takes `skipFirst` explicitly at each call site. Verified with a `-DCMAKE_CXX_STANDARD=11 -DCMAKE_CXX_EXTENSIONS=OFF` build: compiles clean and the test passes.

henryiii:
Thanks!

Issue #1378, opened 2026-07-04
Cannot compile with nvcc (infinite loop) up to at least CUDA 12.6 when using -std=c++20


RL-S:
Now, this is strictly a CUDA problem, not a CLI11 issue. However, it is much harder to ask some HPC system to update their CUDA version than to include a new header file.

### The issue

When compiling a CUDA file that includes the `CLI11.hpp` (v.2.6.2) with `nvcc` (tested v.12.6), the compiler progressively uses up all available memory until the process is killed by the OS.

### How to reproduce

Get a system with the offending compiler, e.g.  this Docker image: `nvidia/cuda:12.6.1-devel-ubuntu24.04`.
Get CLI11 header, e.g., latest release:
```wget https://github.com/CLIUtils/CLI11/releases/download/v2.6.2/CLI11.hpp```

Create test file:
```printf "#include "CLI11.hpp"\nint main(){return 0;}\n" > test.cu```

Compile with
```nvcc test.cu -o test_cu -std=c++20```

I don't know what part of CLI11 `nvcc` can't handle. But since I find this to be the best argument parser I've tried, I'd love to be able to use it in any project, including those that do GPU offloading. 
I wish it were limited to actual `.cu` files, but since we're using Kokkos for portability, any source file whose target links to Kokkos is affected.

I have reported this issue to NVIDIA as well.

henryiii:
Here's an investigation locally (@phlptp is adding more testing coverage in #1381) with Claude:

:robot: _AI text below_ :robot:

Thanks for the detailed report — reproduced and tracked down.

## tl;dr

This is **already fixed on `main`** and will be in the next release. If you need it now, either use the header built from `main` or apply the one change below to your local `CLI11.hpp`.

## Reproduction

Confirmed with your recipe (`nvidia/cuda:12.6.1-devel-ubuntu24.04`, nvcc 12.6.68, `-std=c++20`). Only the C++20 path fails — `-std=c++17`, `-std=c++14`, and the default all compile fine.

One difference by arch: on x86 nvcc chews through memory as you described, while on an arm64 (`sbsa-linux`) image the same build **segfaults `cudafe++` almost instantly**. `--verbose` shows the crash is in `cudafe++ --parse_templates` (the front-end), before `cicc`/`ptxas` — so it's purely a front-end bug, no GPU/driver involved.

## Root cause

[...]

PR #1381, opened 2026-07-07
add a cuda test for single header

phlptp:
Test Issue #1378

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -38,10 +38,12 @@ jobs:
     container: nvidia/cuda:12.6.1-devel-ubuntu24.04
     steps:
       - name: Add build tools
-        run: apt-get update && apt-get install -y git time wget
+        run: apt-get update && apt-get install -y cmake git python3 time
       - uses: actions/checkout@v4
-      - name: Download released single header
-        run: wget https://github.com/CLIUtils/CLI11/releases/download/v2.6.2/CLI11.hpp
+      - name: Configure single-header build
+        run: cmake -S . -B build -DCLI11_SINGLE_FILE=ON -DCLI11_BUILD_TESTS=OFF -DCLI11_BUILD_EXAMPLES=OFF
+      - name: Generate single header
+        run: cmake --build build --target CLI11-generate-single-file
       - name: Create reproduction source
         run: |
           cat <<'EOF' > test.cu
@@ -53,16 +55,16 @@ jobs:
       - name: Try reproducing nvcc single-header compile issue
         id: repro
         run: |
-          /usr/bin/time -v timeout --signal=KILL 3m nvcc test.cu -o test_cu -std=c++20
+          /usr/bin/time -v timeout --signal=KILL 3m nvcc test.cu -I "${GITHUB_WORKSPACE}/build/single-include" -o test_cu -std=c++20
       - name: Summarize reproduction attempt
         if: always()
         run: |
           if [ "${{ steps.repro.outcome }}" = "success" ]; then
-            echo "CUDA 12.6 single-header repro did not trigger; nvcc compiled the released v2.6.2 single header." >> "$GITHUB_STEP_SUMMARY"
+            echo "CUDA 12.6 single-header repro did not trigger; nvcc compiled the single header generated from this branch." >> "$GITHUB_STEP_SUMMARY"
           elif [ "${{ steps.repro.outcome }}" = "skipped" ]; then
             echo "CUDA 12.6 single-header repro did not run because an earlier setup step failed." >> "$GITHUB_STEP_SUMMARY"
           else
-            echo "CUDA 12.6 single-header repro triggered or timed out while compiling the released v2.6.2 single header." >> "$GITHUB_STEP_SUMMARY"
+            echo "CUDA 12.6 single-header repro triggered or timed out while compiling the single header generated from this branch." >> "$GITHUB_STEP_SUMMARY"
           fi
 
   cuda13-build:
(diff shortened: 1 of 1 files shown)


henryiii:
I think you need a newer base image or checkout@v1.

henryiii:
FYI, I can do a local reproduction pretty easily:

Yes — reproduced. Here's what I found.

## Reproduction confirmed (#1378)

Using the exact recipe from the issue (v2.6.2 single header, `test.cu` = `#include "CLI11.hpp"` + empty `main`, nvcc 12.6.68 in `nvidia/cuda:12.6.1-devel-ubuntu24.04`):

| `nvcc` flag  | Result    |
| ------------ | --------- |
| `-std=c++20` | **fails** |
| `-std=c++17` | ✅ EXIT=0 |
| `-std=c++14` | ✅ EXIT=0 |
| (default)    | ✅ EXIT=0 |

So the failure is specific to the C++20 path, matching the report.

**Manifestation differs by arch.** My machine is Apple Silicon, so Docker ran the **arm64 (`sbsa-linux`)** CUDA image. There nvcc doesn't slowly exhaust memory — it **segfaults almost instantly** (exit 139). The reporter is presumably on x86, where the same bug shows up as the runaway memory blowup they described. Same root cause, different symptom.

**Where it crashes:** `--verbose` shows the `gcc -E` preprocess step succeeds; the crash is inside **`cudafe++`** (the `--parse_templates` C++ front-end) before it ever reaches `cicc`/`ptxas`. That's a bug in NVIDIA's front-end template parser, triggered by CLI11's templates under C++20 — consistent with the reporter's note that this is "strictly a CUDA problem, not a CLI11 issue."

Notes on my setup:
- No GPU needed — this is a compile-time (front-end) crash, so the missing NVIDIA driver warning is irrelevant.
- I ran it with the container capped at 6–8 GB so a memory blowup couldn't hurt the host; the arm64 crash is fast enough that the cap never mattered.

## Want me to go further?
The issue asks implicitly whether CLI11 can add a header workaround. Since each failing compile crashes in <1s, it'd be cheap to **bisect which part of `CLI11.hpp` trips `cudafe++`** (isolate the offending template) and see if a small guarded change avoids it. I can also test whether current `main`'s single header still reproduces. Want me to do either?

henryiii:
Investigation ongoing, but it says current main is clean, and is git bisecting now.

henryiii:

Both tasks done. Tree is clean, back on `main`.

## Results

### 1. Does current `main` still reproduce? → **No.**
Regenerated the single header from source at HEAD (`3054373`) and it compiles cleanly under `-std=c++20` with nvcc 12.6 (confirmed twice). The v2.6.2 tag regenerated from source *does* crash — so this is a real source fix, not a stale-release artifact.

### 2. Which part of CLI11 trips `cudafe++`? → **Pinpointed to one construct.**

`git bisect` (49 commits, `v2.6.2..main`) identified the fixing commit:

> **`a5950f4` — "static std::string & initializers (#1335)"** by Philip Top

Before that commit, `include/CLI/impl/StringTools_inl.hpp` had four **namespace-scope `const std::string&` references bound to string-literal temporaries**:

```cpp
inline const std::string &escapedChars("\b\t\n\f\r\"\\");
inline const std::string &escapedCharsCode("btnfr\"\\");
inline const std::string &bracketChars("\"'`[(<{");
inline const std::string &matchBracketChars("\"'`])>}");
```

#1335 replaced them with Meyers-singleton helper functions (`static const std::string s{...}; return s;`), which is what inadvertently fixed the CUDA crash.

**Minimal standalone reproducer** (no CLI11 — three lines segfault `cudafe++` under `-std=c++20`, compile fine otherwise):
```cpp
#include <string>
inline const std::string &escapedChars("\b\t\n\f\r\"\\");
int main(){ return escapedChars.empty() ? 0 : 1; }
```

### Notes on manifestation
- On my arm64 (Apple Silicon) CUDA image the bug is a **fast `cudafe++` segfault (exit 139)**; the reporter's x86 system shows the same front-end bug as a runaway memory blowup. Same root cause — a bug in NVIDIA's `--parse_templates` front-end handling a lifetime-extended namespace-scope `std::string` reference under C++20.
- It's `-std=c++20`-only; c++17/c++14/default all compile in every case.

### Takeaway for the issue
The crash is **already fixed on `main`** and will resolve in the next release. The reporter can either wait for that or apply the #1335 change to their local v2.6.2 header. Worth posting the 3-line minimal reproducer on their NVIDIA bug report — it's far cleaner than the full CLI11.hpp.

Want me to draft a comment for #1378 with this summary and the minimal reproducer?

phlptp:
interesting,   What AI system are you using @henryiii.  And do we want to keep the CI job or just drop this PR since the issue is fixed in main already?

henryiii:
This seems like a good thing to test also, and it's pretty fast. Not sure why coverage is reporting a drop.

I use the Linux Assisted-By trailers, so you can tell if it's anything with a commit. :) Currently I have the Claude OSS sub, which is what I'm using here (Claude Code, Fable and Opus), I also have some amount of GPT 5.5 I use for reviewing and access to OSS models via NRP.ai.

henryiii:
Oh, and I write about things I do, generally: https://iscinumpy.dev/post/starting-with-agentic-ai/ (before Claude) and https://iscinumpy.dev/post/claude-code-reviews/ and https://henryiii.github.io/pu_rse_2026_07/

phlptp:
@henryiii I don't think the codecov is related to this PR, so I am thinking to merge this and then deal with the code coverage in a future PR

henryiii:
Pretty sure it's a flake with some of the runs not reporting. I've seen it on another PR too.

PR #1382, opened 2026-07-08
feat: add PrefixCommandMode::PositionalOnly (#1374)

### LINKED ISSUE #1374: `App::prefix_command(true)` changed behavior between // CLI11 2.5.0 and 2.6.2.
We have observed a regression in behavior when updating CLI11 in ParaView (by way of this [update](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/13309) in VTK) from 2.5.0 to 2.6.2.

`App::prefix_command(true)` is documented as: "Do not parse anything after the first unrecognized option (if true) all remaining arguments are stored in remaining args". We use this combined with `allow_extras(true)` to parse a command line that may contain options we haven't registered with the App *yet* (we register the rest of our options and re-parse later, once we know more about what's needed) without errors, while still picking out the handful of options we *have* registered so far, wherever they appear on the command line.

Under CLI11 2.5.0, encountering one unrecognized option-like token only marked that single token as "missing"/extra; parsing then continued
scanning the remaining tokens normally, so a later, registered option was still recognized regardless of position.

Under CLI11 2.6.2, the *first* unrecognized option-like token now causes ALL remaining tokens -- recognized or not -- to be swept into "missing" immediately, with no further attempt to match them. This makes whether a registered option is recognized depend on its position relative to unrecognized ones, which it did not before.

A minimal reproducer is attached here: 

[prefix_command_regression.cpp](https://github.com/user-attachments/files/29479553/prefix_command_regression.cpp)

Expected output (matches CLI11 2.5.0 behavior, and what this reproducer
asserts):
  order: --unknown --known => known=1
  order: --known --unknown => known=1

Actual output on CLI11 2.6.2:
  order: --unknown --known => known=0   <-- "--known" silently dropped
  order: --known --unknown => known=1

We can remedy the regression in ParaView by calling `App::prefix_command(false)` wherever we have a situation where only a few of the full list of arguments are defined. However, this change in behavior took some time to track down and was something of a surprise.

henryiii:

Fixes #1374.

Reworked per review: instead of reverting the 2.6.2 behavior, this keeps `prefix_command(true)` (`PrefixCommandMode::On`) as is — parsing stops at the first unrecognized item — and adds a new `PrefixCommandMode::PositionalOnly`.

In `PositionalOnly` mode, only a positional argument or the `--` separator triggers prefix mode; an unrecognized option is collected as an extra and parsing continues. This is the pre-2.6.2 behavior as an opt-in mode, which is what ParaView/VTK depended on (they can switch to `prefix_command(CLI::PrefixCommandMode::PositionalOnly)`).

Includes tests for the new mode and README/docstring updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1161,7 +1161,10 @@ option_groups. These are:
   `prefix_command(true)` and `prefix_command(false)`. Calling with
   `PrefixCommandMode::SeparatorOnly` will only trigger prefix command mode with
   the subcommand separator `--`; other unrecognized arguments are considered an
-  error unless `allow_extras` is enabled.
+  error unless `allow_extras` is enabled. Calling with
+  `PrefixCommandMode::PositionalOnly` will only trigger prefix command mode at
+  the first positional argument or the separator `--`; unrecognized options do
+  not stop processing and are collected in the remaining_arg list.
 - `.usage(message)`: Replace text to appear at the start of the help string
   after description.
 - `.usage(std::string())`: Set a callback to generate a string that will appear
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -84,8 +84,9 @@ enum class config_extras_mode : std::uint8_t { error = 0, ignore, ignore_all, ca
 
 /// @brief  enumeration of prefix command modes, separator requires that the first extra argument be a "--", other
 /// unrecognized arguments will cause an error. on allows the first extra to trigger prefix mode regardless of other
-/// recognized options
-enum class PrefixCommandMode : std::uint8_t { Off = 0, SeparatorOnly = 1, On = 2 };
+/// recognized options. positional_only triggers prefix mode only at the first positional argument (or "--");
+/// unrecognized options do not stop parsing and are collected as extras
+enum class PrefixCommandMode : std::uint8_t { Off = 0, SeparatorOnly = 1, On = 2, PositionalOnly = 3 };
 
 class App;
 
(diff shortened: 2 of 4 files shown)


henryiii:
@phlptp thoughts? This was changed in 2.6.2. We could introduce a new mode for the strict stop at any option mode. Or we could reintroduce the old mode, though that's more churn since this has only been out in one patch release. (This PR is just the revert, but any path is possible)

phlptp:
the `prefix_command` mode has caused a bit of confusion over time.   I remember several issues raised of it not working as expected and the previous behavior being confusing to people.  So the previous behavior caused some issues and confusion, especially when used in combination with `fallthrough` and `allow_extras`.  The current behavior made that clear to better match the documentation, but I guess there were those depending  on the nuances of the behavior.   I think the intended case is putting `prefix_command` on a subcommand with no option so that command dumps everything remaining since there are no valid options after that, or with a new mode using `--`  the complexity comes when adding `prefix_command` on an app/subcommand with existing options or `fallthrough` or `allow_extras` that it can interact with the parent options and get some odd behavior.   I think my preference would be to leave the current behavior as is, as I would argue that it is clearer than before,  and to address this issue add a PositionalOnly mode that turns off the while loop for unknown arguments.

henryiii:
Ahh, I wish it wouldn't auto-push. I'll comment or undraft when happy with it. Edit: looks pretty good, ready for review.

Reworked as suggested: the revert is dropped, `PrefixCommandMode::On` keeps the current (2.6.2) behavior, and a new `PrefixCommandMode::PositionalOnly` provides the stop-only-at-positionals/`--` behavior as an opt-in mode. Title and description updated.

PR #1396, opened 2026-07-30
perf: replace stringstream in detail::split with find-based loop

### LINKED ISSUE #1395: perf: replace stringstream in detail::split with find-based loop
## Summary
`detail::split` in `StringTools_inl.hpp` uses `std::stringstream` + `std::getline`,
which incurs heap allocation and stream overhead for a simple delimiter split.

## Proposed change
Replace with a `std::string::find` loop:

```
CLI11_INLINE std::vector<std::string> split(const std::string &s, char delim) {
    std::vector<std::string> elems;
    if(s.empty()) {
        elems.emplace_back();
        return elems;
    }
    std::size_t start = 0, end;
    while((end = s.find(delim, start)) != std::string::npos) {
        elems.push_back(s.substr(start, end - start));
        start = end + 1;
    }
    elems.push_back(s.substr(start));
    return elems;
}
```
Behavior

Identical to the original: empty string returns a single empty element;
otherwise splits on delim including trailing

Motivation

Avoids std::stringstream construction and locale-aware stream machinery
for a hot path that only needs basic string

ArbStray:
closed #1395


ArbStray:
Note on coverage drop

The Codecov report shows a missed line in ExtraValidators_inl.hpp:41:

if(result.size() != 4) {
    return std::string("Invalid IPV4 address: must have four parts (") + ip_addr + ')';
}

This branch is unreachable: the check on line 36 (cdot != 3) already guarantees exactly 3 dots, so split will always return exactly 4 parts. The coverage miss is pre-existing dead code unrelated to this PR's changes.

Happy to remove the dead branch in a follow-up or as part of this PR if preferred. @henryiii

henryiii:
The uncovered line is due to the behavior change in this PR: `a.b.` now returns three elements, the empty one after the delimiter, which the stringstream version dropped that final element. I think this new one is a better behavior. `Option_inl.hpp:748` skips empty items already, `Formatter_inl.hpp:333` never ends in a delimiter, and `ExtraValidators_inl.hpp:39` is why the coverage has gone down. I'd recommend deleting that check (lines 40-42 following that), as it's not needed anymore.

henryiii:
Other implications, like Python, follow the new behavior in this PR:

```python
>>> "1.2.".split(".")
['1', '2', '']
```

ArbStray:
thanks for the Python reference I'll remove lines 40–42 in ExtraValidators_inl.hpp now.

henryiii:
Thanks!

PR #1399, opened 2026-07-30
chore: prepare 2.7.0 release

henryiii:
Edit: okay, I think this (and we) are ready. @phlptp, are we okay to release?

🤖 _AI text below_ :robot:

Add the 2.7.0 changelog entry and bump the version to 2.7.0. Also update the copyright years in the license and the version in the installation docs.

PR #1403, opened 2026-07-30
ci: check the changelog entry on every build

henryiii:

Run the release notes extraction on every build, with the version from `scripts/ExtractVersion.py` when the run is not for a tag. A version bump that has no matching changelog section now fails the build, instead of making a release with empty notes.

This means the version bump and the changelog entry must land in the same PR, which matches how 2.6.2 and #1399 were done.


henryiii:
I made this a stack to try out GitHub's new stacked PRs, not because it really needed to be. :)

Issue #343, opened 2019-11-15
Autocomplete for CLI11


jbriales:
About "Autocomplete: This might eventually be added to both Plumbum and CLI11, but it is not supported yet."
Is there any plans or thoughts on this feature at this point?

I've been thinking about giving this a try myself for some time.
Before going for it, I'd appreciate any pointers/caveats/ideas in this regard :)

jbriales:
No thoughts here? :)

henryiii:
Yes, thoughts are just slow coming. :)

I looked into this a bit, and have some general ideas. I'll try to collect my ideas and would be happy to have someone work on this.  Ping me on Thursday-Friday if I don't do something by then (have to prepare materials to teach a class Wednesday).

jbriales:
Haha thanks, that sounds good! Now I can align my plans on this better too ;)

jbriales:
Hey @henryiii friendly ping here :)

henryiii:
I’ll write it up when I get home today. :) Edit: midday/evening tomorrow, too late now.

henryiii:
I've previously talked briefly with the author of argcomplete:

https://github.com/kislyuk/argcomplete/issues/116

It would be nice to be able to reuse as much as possible from argcomplete, possibly even have a general completer. Not sure how much would work for C++, since the procedure would have been different, though plumbum and other CLI toolkits for scripting languages could. I don't know how to ensure a text string is near the beginning of a binary; you probably cannot as that might be used for indexing. So that method is probably out.

How familiar are you with argcomplete? The basics there would be the same here. There would be a way to run the program to get just the completions, given a string and a location. It would predict the possible options that would make sense for that string and location. It would then return the list of completions. The hard parts are letting the system know that an executable supports completions; the trick of reading the first few bytes of a program would probably be quite hard in a compiled program.

So I think we would need:

* A general `.completions(string in, int loc)` method that could do the work
* A way to trigger and run this. Ideally matching argcomplete to the degree reasonable.
* A way to tell the shell this can be done.

Your thoughts?

I've worked through the beginnings of this in a branch somewhere in plumbum, quite a while ago.

jbriales:
I've used argcomplete in the past, so I have some very high-level idea how it works , although I've never looked into the internal implementation.
I'm not sure what's your idea of reusing argcomplete as a general completer. Are you thinking of emulating the logic there in C++ or how?

[...]

PR #1422, opened 2026-08-02
feat: dynamic shell completion, starting with bash

cachelous:
Part of #343. For now only bash is supported, I will work on zsh/fish on later PRs.

Instead of generating a static script that encodes the app's options, the script is a small fixed adaptor: it re-invokes the binary with the line being completed, and the binary replies with candidates and a directive. Rationale in #343. A detailed explanation is included in `book/chapters/completion.md`.

### How to try it out

To see the current state of completion in action (from a bash shell with `bash-completion`):

```bash
cmake --workflow default && cd build
PATH="$PWD/examples:$PATH"
source <(completion --completion bash)
completion <TAB>
# and start playing around with different args
```

To see the wire protocol directly, without a shell in the way:

```console
$ CLI11_COMPLETE=bash CLI11_COMPLETE_INDEX=1 CLI11_COMPLETE_PROTO=1 completion --l
--level       How hard to try
:2
```

### Status

Working end to end for bash:
- Subcommands, including nested ones and aliases
- Option names, with their descriptions
- Values from validators
- `--opt=value`, and values written onto short names and bundles
- Path completion, delegated to the shell
- `--` and `++`
- Positionals, at both multiplicities
- `/flag:value` under `allow_windows_style_options`
- Options and positionals declared in nameless option groups
- Name matching that honours `ignore_case` and `ignore_underscore` the way a parse does

Deliberately not in this PR, happy to follow up:
- zsh, fish and powershell generators
- Per `Option` completion callbacks
- Candidate grouping
- Dot notation (teaching completion that `--sub.opt` takes a value)

Reviewing commit-by-commit will be easier than the squashed diff, as each commit is one capability with its own tests.


henryiii:
Let me know if you want any help rebasing. Sorry for the conflicts, I'm done for a bit (was working on improving compile time performance).

cachelous:
No worries! Almost no conflicts :smile: 

The PR is ready for review now. Just one thing, not sure how to handle the 7 `unusedStructMember` warnings from cppcheck on the new completion structs (the members are only read from the `_inl.hpp` files, which cppcheck doesn't see. Any preference on how to deal with this?

cachelous:
Went ahead and silenced the `unusedStructMember` warnings with `cppcheck-suppress` inline comments, as they are used after all. No rush on the review, I know it's a big diff. If it helps, I am happy to split it into smaller PRs

henryiii:
I probably won't review till I'm back from vacation; doing little things here and there, but probably not big reviews. :)

cachelous:
Don't worry, let me know if there is anything I can do to help. For now I added some unittests to reach full coverage, I think it was the last thing missing

PR #1431, opened 2026-08-09
fix: preserve precision in floating-point default values

### LINKED ISSUE #1429: Work around issues with std::to_string
When converting floating-point values to string, the std::to_string implementation before c++26 behaved in an unexpected way (see https://en.cppreference.com/cpp/string/basic_string/to_string for details).

Something like:
```
double foo = 1e-8;
CLI::Option* foo_opt = args_parser.add_option("--foo", foo, "")->default_val(foo);
```

will result in a default value of 0, while something like
```
double foo = 1e-8;
CLI::Option* foo_opt = args_parser.add_option("--foo", foo, "")->default_val("1e-8");
```
will result in a default value of 1e-8.

I didn't test against the most recent version, though.

ThunderKhan:
Closes #1429 

`default_val()` uses `detail::value_string()` to convert values to strings before parsing them back. For floating-point values, this relied on `std::to_string`, which can lose precision for small values such as `1e-8`, producing `"0.000000"`.

This change separates integral and floating-point handling in `value_string()`:

- Integral types continue to use `std::to_string`.
- Floating-point types use a precision-preserving stream conversion with `std::numeric_limits<T>::max_digits10`.

A regression test was added to verify that small floating-point default values such as `1e-8` are preserved correctly.

Validation performed locally:
- Full test suite passes.
- Formatting checks pass with `prek`.
- Additional round-trip checks were performed for `float`, `double`, and `long double`.

### DIFF
--- a/include/CLI/TypeTools.hpp
+++ b/include/CLI/TypeTools.hpp
@@ -15,6 +15,8 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdlib>
+#include <iomanip>
+#include <limits>
 #include <memory>
 #include <sstream>
 #include <string>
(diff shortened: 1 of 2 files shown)


phlptp:
@all-contributors add @ThunderKhan  for code

phlptp (approved):

PR #1432, opened 2026-08-10
Fix/floating precision to string

ArbStray:
Extend #1431's precision fix to to_string so composite types with floating-point members also round-trip exactly


ArbStray:
#1431 fixed precision loss in value_string for floating-point scalars, but missed the same problem in detail::**to_string**, which is the underlying conversion used by **composite types** (pair, tuple, vector containing float/double members), by validators like AsNumberWithUnit and Bound that write results back via to_string, and by the third value_string overload that delegates to to_string for **non-arithmetic types** — all of these paths still stream floating-point values with **default (~6 digit) precision**, losing values like 1e-8 to "0.000000". This PR adds a **floating-point-specific** to_string overload using **max_digits10** precision.
Can I get a review on this? @henryiii

henryiii:
The test names are duplicated now, need deduplication.

ArbStray:
All checks have passed. Ready for your review @henryiii

phlptp:
@all-contributors add @ArbStray  for code

PR #1433, opened 2026-08-10
fix(windows): link shell32 via pragma for header-only MSVC (#1053)

### LINKED ISSUE #1053: MSVC error LNK2019: unresolved external symbol __imp_CommandLineToArgvW referenced in function CLI::detail::compute_win32_argv(void)
Hello!

I'm using CLI11 in a CMake project on Windows with:
```
# Don't link implicitly with gdi32.lib, winspool.lib, shell32.lib, etc
# https://stackoverflow.com/questions/41400536/why-cmake-adds-unnecessary-libraries-to-visual-studio-projects
set(CMAKE_CXX_STANDARD_LIBRARIES_INIT "")
set(CMAKE_CXX_STANDARD_LIBRARIES "")
```

This causes a linker error:
```
error LNK2019: unresolved external symbol __imp_CommandLineToArgvW referenced in function CLI::detail::compute_win32_argv(void)
```

Since CLI11 uses `CommandLineToArgvW()` it shouldn't rely on Shell32.lib being implicitly linked and instead do it explicitly.

Best regards,
Nikolai

hannahgreendesk:
## Summary

Fixes #1053.

CMake and Bazel already link `shell32` when consumers use the `CLI11` / `cli11_header_only` targets. That does not help MSVC (or clang-cl) projects that only include the headers, or that clear `CMAKE_CXX_STANDARD_LIBRARIES` so the usual implicit `shell32.lib` is gone — they still hit `LNK2019` on `CommandLineToArgvW`.

This PR adds `#pragma comment(lib, "shell32.lib")` next to the Windows argv helper so header-only MSVC builds pull in `shell32` explicitly. Existing CMake/`shell32` and Bazel `DEFAULTLIB:shell32.lib` links stay for MinGW and target-based builds.

## Test plan

- [ ] MSVC: build a tiny TU that only `#include <CLI/CLI.hpp>` / calls `CLI::detail::compute_win32_argv` (or `ensure_utf8`) with `CMAKE_CXX_STANDARD_LIBRARIES` cleared and without linking `CLI11::CLI11`; confirm link succeeds
- [ ] Existing Windows CI (MSVC / clang-cl) stays green
- [ ] Non-Windows builds unchanged

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,5 @@
 # Changelog {#changelog}
 
-## Version 2.7.3: Explicit shell32 for MSVC headers
-
-### Fixed
-
-- Link `shell32` for MSVC/clang-cl even when CLI11 is used as headers without
-  the CMake or Bazel target, via `#pragma comment(lib, "shell32.lib")` next to
-  `CommandLineToArgvW`. CMake/`shell32` and Bazel `DEFAULTLIB:shell32.lib`
-  remain for MinGW and target-based builds. Fixes [#1053][].
-
-[#1053]: https://github.com/CLIUtils/CLI11/issues/1053
-
 ## Version 2.7.2: Faster compiles
 
 This patch release makes the precompiled and experimental module modes cheaper
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog {#changelog}
 
+## Version 2.7.3: Explicit shell32 for MSVC headers
+
+### Fixed
+
+- Link `shell32` for MSVC/clang-cl even when CLI11 is used as headers without the
+  CMake or Bazel target, via `#pragma comment(lib, "shell32.lib")` next to
+  `CommandLineToArgvW`. CMake/`shell32` and Bazel `DEFAULTLIB:shell32.lib` remain
+  for MinGW and target-based builds. Fixes [#1053][].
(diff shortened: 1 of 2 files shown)


henryiii:
Looked it up here: https://learn.microsoft.com/en-us/cpp/preprocessor/comment-c-cpp?view=msvc-170#lib

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+
+### Fixed
+
+- Link `shell32` for MSVC/clang-cl even when CLI11 is used as headers without
+  the CMake or Bazel target, via `#pragma comment(lib, "shell32.lib")` next to
+  `CommandLineToArgvW`. CMake/`shell32` and Bazel `DEFAULTLIB:shell32.lib`
+  remain for MinGW and target-based builds. Fixes [#1053][].
+
+[#1053]: https://github.com/CLIUtils/CLI11/issues/1053
+
```
henryiii:
```suggestion
```

Let's not change this yet.

hannahgreendesk:
Thanks @henryiii — confirmed against the MSVC `#pragma comment(lib)` docs you linked. Left the pragma in place for header-only / cleared-stdlib MSVC consumers, cited that page in the comment, and kept CHANGELOG untouched as requested.

henryiii (approved):

PR #1435, opened 2026-08-13
feat(help): add header() for extended help text after description (#507)

hannahgreendesk:
## Summary

Adds `App::header()` / `get_header()` as the complement to the existing `footer()` API, addressing #507.

Extended help text is rendered after the short description and before the usage line in both normal and expanded help output. Subcommand listings in the main help still show only the short description.

## Changes

- `header(std::string)` and `header(std::function<std::string()>)` on `App`, mirroring `footer()`
- `Formatter::make_header()` integrated into `make_help()` and `make_expanded()`
- Header is inheritable from parent apps; cleared for option groups (like footer)
- Tests for header placement, callbacks, and inheritance

## Example

```cpp
auto *fmt = app.add_subcommand("fmt", "Format files in-place.");
fmt->header("If a file is given, it reformats in-place. Otherwise output goes to stdout.");
```

Fixes #507

## Test plan

- [x] `HelpTest`: header appears before Usage in help output
- [x] `HelpTest`: subcommand header not shown in subcommand list
- [x] `CreationTest`: header inherited by subcommands

PR #1436, opened 2026-08-13
feat(help): show option delimiter in help text (#1423)

### LINKED ISSUE #1423: [Enhancement] Setting a delimiter for container-like options should disable greedy matching
Currently, as far as I understand it, container-like options _always_ consume as many arguments as possible ((up to a bound set by `expected()`)). However, this greedy behavior still occurs when `delimiter()` is set, which becomes very unintuitive in combination with ((optional)) positional arguments. 

Consider a very simple example:
```cpp
std::filesystem::path inputPath{};
app.add_option(
	 "input_path",
	 inputPath
)->check(CLI::ExistingDirectory);

std::vector<std::string> strings{};
app.add_option("-s", strings)->delimiter(',');
```
If we call `program /some/dir -s abc,def` we get the intended behavior of `inputPath = "/some/dir"` and `strings = {"abc", "def"}` ((though the default help message techincally forbids passing positional arguments before options, so the user would need to figure that out first)), while calling the much more intuitive `program -s abc,def /some/dir` results in the _unintended_ `inputPath = ""` and `strings = {"abc", "def", "/some/dir"}`. This latter case the the one I'm concerned with, because even though the delimiter is conceptually binding the two arguments much more strongly, the following argument is still matched greedily and included into the container instead of the positional argument.

My proposal is for either the default behavior of `delimiter()` to disable this greedy matching, or to add a flag that allows disabling it ((as well as updating the default help text to display the delimiter)). From my perspective, there is no situation in which an end user calling `program -s abc,def /some/dir` _ever_ intends for `/some/dir` to be part of the arguments of `-s`, and I think just fully disallowing that case ((potentially even with an error)) would make the program behavior much less surprising to an end user. 

Potentially, setting a delimiter should just fully disable the ability to pass multiple arguments _without_ using the delimiter ((this could probably be implemented by simply only matching the argument directly after the option flag if `delimiter()` is set)), though that might be too much of a breaking change

hannahgreendesk:
## Summary

Shows the configured element delimiter in option help output, addressing the help-text portion of #1423.

When `->delimiter(',')` is set on a container option, help now includes `(Delimiter:',')` alongside other option metadata (similar to the existing `Env:` annotation).

## Example

```
--idx INT ... (Delimiter:',')
```

Fixes #1423 (help text)

## Test plan

- [x] `FormatterTest`: delimiter annotation appears in help for delimited options

### DIFF
--- a/include/CLI/impl/Formatter_inl.hpp
+++ b/include/CLI/impl/Formatter_inl.hpp
@@ -182,14 +182,6 @@ CLI11_INLINE std::string Formatter::make_usage(const App *app, std::string name)
     return out.str();
 }
 
-CLI11_INLINE std::string Formatter::make_header(const App *app) const {
-    std::string header = app->get_header();
-    if(header.empty()) {
-        return std::string{};
-    }
-    return header + "\n\n";
-}
-
 CLI11_INLINE std::string Formatter::make_footer(const App *app) const {
     std::string footer = app->get_footer();
     if(footer.empty()) {
@@ -215,11 +207,6 @@ CLI11_INLINE std::string Formatter::make_help(const App *app, std::string name, 
             out, make_description(app), description_paragraph_width_, "");  // Format description as paragraph
     } else {
         out << make_description(app);
-    }
-    if(is_description_paragraph_formatting_enabled()) {
-        detail::streamOutAsParagraph(out, make_header(app), description_paragraph_width_, "");
-    } else {
-        out << make_header(app);
     }
     out << make_usage(app, name);
     out << make_positionals(app);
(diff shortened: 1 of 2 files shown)


hannahgreendesk:
CI fix: rebuilt this branch from upstream `main` with **only** the delimiter annotation in `make_option_opts` (+ `Formatter: DelimiterInHelp` test).

The previous tip incorrectly included `make_header()` / `get_header()` calls without the rest of the header API from #1435, which broke compilation. Those stubs are removed; this PR no longer depends on `header()`.

Note: #1425 (maintainer) takes a different approach (delimiter on the type name). Happy to adjust or close if that lands instead.

hannahgreendesk:
CppLint was failing on the new DelimiterInHelp test (missing #include <vector> for vector<>). Added the include; tests were already green.

henryiii:
The one without an English word added was merged. Thanks though!

PR #1439, opened 2026-08-14
fix: support wide strings for app description and usage

### LINKED ISSUE #1130: Add support for Unicode string in app usage under windows
The following code doesn't work when passing wide char strings
```cpp

	CLI::App app{ L"Some unicode string"};
	app.usage(L"Some unicode string");
```

That means that under windows, these string cannot contain Unicode characters.

ThunderKhan:
Closes #1130.

## Summary

Adds support for wide strings in the `CLI::App` description constructor and `usage()` setter.

## Changes

* Add a `std::wstring` overload for the `CLI::App` description constructor.
* Add a `std::wstring` overload for `App::usage()`.
* Convert wide strings through CLI11's existing `narrow()` helper and delegate to the existing `std::string` implementations.
* Add a regression test covering wide string literals with non-ASCII Unicode characters.

## Testing

* Added a regression test reproducing the previously unsupported wide-string calls.
* Verified the regression test fails to compile without the new overloads.
* Built successfully with C++11.
* Ran the full test suite: 83/83 tests passed.
* Ran `prek` successfully.
* Ran `git diff --check` successfully.

### DIFF
--- a/include/CLI/App.hpp
+++ b/include/CLI/App.hpp
@@ -345,6 +345,9 @@ class App {
     /// Create a new program. Pass in the same arguments as main(), along with a help string.
     explicit App(std::string app_description = "", std::string app_name = "");
 
+    /// Create a new program with a wide-string description.
+    explicit App(std::wstring app_description);
+
     App(const App &) = delete;
     App &operator=(const App &) = delete;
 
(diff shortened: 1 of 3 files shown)


flagarde:
@ThunderKhan Please don't do that! wchar_t is an old and ugly thing. It predates unicode and utf-8 and it is not cross-platform as Windaube and UNIX etc don't agree on what wchar_t means. wchar_t is just useful in win© and that windows uglyness would be better burried in the code and not in the interface.

ThunderKhan:
@flagarde Thanks for raising this. These overloads delegate immediately to the existing `CLI::narrow()` helper, so they retain UTF-8 internally and reuse its platform-specific conversion. My intent was to address the calls reported in #1130, following the existing wide-input overloads. Whether to extend that convenience to description and usage is a fair API-design question; I’d welcome maintainer input on that versus documenting explicit conversion.

PR #1451, opened 2026-08-30
Fix duplicate option in TOML config output for + option groups

### LINKED ISSUE #1449: Option group with '+' prefix in name causes duplication in TOML
MWE:

```cpp
#include "CLI/CLI.hpp"
#include <iostream>
#include <string>

    CLI::App app;

    app.set_help_all_flag("-H,--help-all");

    // [config file]
    app.set_config("-c,--config", "", "Read a configuration file");
    app.allow_config_extras(false); // unknown CLI arguments trigger error

    CLI::Option_group *input = app.add_option_group("Input");
    bool is2d;
    bool is3d;
    std::string input_file;
    {
        auto dim = input->add_option_group("+dim");

        auto d2 = dim->add_flag("-2,--2d", is2d);
        auto d3 = dim->add_flag("-3,--3d", is3d);
        d3->excludes(d2);

        dim->require_option(1);

        input->add_option("-i,--input", input_file)->option_text("<file>");
        input->add_flag("--trans");
    }

    try {
        app.parse(argc, argv);
    } catch(const CLI::ParseError &e) {
        return app.exit(e);
    }

    std::cout << "# ==== config ==== #\n" + app.config_to_str(CLI::ConfigOutputMode::Active, /* write_description */ true);

    return 0;
}
```

```sh
$ ./issue -2
# ==== config ==== #
## 

## Input Options
## 
2d=true

## +dim Options
## 
2d=true # <--- duplicate
```

```sh
./issue -2 > test.toml && ./issue -c test.toml
2d: too many inputs for a flag
Run with --help or --help-all for more information.
```

Is this a bug?

afonsojanu:
Closes #1449.

A nested option group whose display name starts with `+` (used to flatten its options into the parent's own listing) has its options merged into the parent's option list by `App::get_options()`. But `Config::to_config()`'s subcommand recursion loop still walked into these groups unconditionally, so their options were written out a second time under a separate section.

Beyond the visible duplication in the output, this also breaks reading the config back in: the reported example ends with `2d: too many inputs for a flag`, since the flag's value appears twice in the file and the parser rejects that.

The fix mirrors the exact condition `get_options()` already uses to decide whether a group's options are folded into the parent (`!group.empty() && group.front() == '+'`), and skips the recursive `to_config` call for those groups since their content is already covered by the parent's own pass over its (already-merged) option list.

Added a regression test (`TomlOutputPlusGroupNoDuplication`) reproducing the reported scenario: it checks the duplicate line disappears from the output, and that the round trip through `parse_from_stream` no longer throws.

Ran the config, option-group, subcommand, and app test suites locally; all pass except a single pre-existing failure in `AppTest` (`ensure_utf8_twice`, about argv handling in a spawned helper process) that reproduces identically with or without this change, so it's unrelated to this fix.


phlptp (approved):

phlptp:
@all-contributors add @afonsojanu  for code

PR #1453, opened 2026-08-31
Fix Meson amalgamated tests

meator:
This PR's main focus is to make the test pass with `mode=amalgamated`. While reviewing the relevant parts of the build system, I fixed some minor issues too.

This PR might deserve more testing. The tests pass, I've tested that, but the other minor additions change behavior.

There's a warning from `MakeSingleHeader.py`:

```
/home/meator/git/CLI11/single-include/../scripts/MakeSingleHeader.py:62: UserWarning: Failed to find any matches in ../include/CLI/CLI.hpp
  warnings.warn(
```

I have yet to investigate that (could be a problem on my end).


henryiii:
I pushed some fixes based on the review below, maintainer edits don't seem to be on, so they are in `pr/fix-meson-amalgamated-tests`. Pull or cherry-pick bca8673fcafffcf1d924d176a559c48ade6cd6e9 for them.

Code review finished on `pr/fix-meson-amalgamated-tests`. All findings were verified empirically by the reviewer (all three Meson modes built and tested; the proposed fix for #2 was built in a scratch clone, 25/25 pass).

**Findings, most important first:**

1. **`single-include/meson.build:11` — `python` shebang breaks the amalgamated build.** `MakeSingleHeader.py` runs via `#!/usr/bin/env python`; on a machine with only `python3` you get `env: python: No such file or directory`. Since the PR makes the target `install: true`, this now fails by default. Fix: `[find_program('python3'), find_program('../scripts/MakeSingleHeader.py'), ...]`.

2. **`meson.build:68` — the amalgamated `cli11_dep` should carry the define and both include dirs.** Unlike CMake's `CLI11_SINGLE` target it exposes only the single-include dir and no `-DCLI11_SINGLE_FILE`, so a subproject consumer using the `#ifdef CLI11_SINGLE_FILE` idiom gets `'CLI/CLI.hpp' file not found`. Verified fix:
   ```meson
   cli11_dep = declare_dependency(sources: single_header,
     include_directories: [include_directories(meson.current_build_dir() / 'single-include'), cli11_inc],
     compile_args: '-DCLI11_SINGLE_FILE')
   ```
   With that, `tests/meson.build` can be restored byte-for-byte to `main` (TimerTest and link_test_2 back in the dict) and all 25 tests pass. This makes findings 4, 6, 7, 8 moot.

3. **`.github/workflows/build-tests.yml:132` — CI only runs `headeronly` mode**, so the path this PR fixes can regress silently. Add a matrix over `mode: [headeronly, amalgamated, precompiled]`.

4. **`tests/meson.build:92`** — TimerTest and link_test_2 are hand-rolled outside the `testnames` loop; per-entry `dependencies`/`link_with` overrides in the dict already suffice and the copies already diverge (TimerTest lacks `depends: app_tgts`/`workdir`).

5. **`meson.build:72`** — comment says the targets use `cli11_nonamalgamated_dep` "instead of" `cli11_dep`, but they get both via `testdep`, and it omits the real reason (`Timer.hpp` is deliberately not in the single header).

6. **`meson.build:76`** — `cli11_nonamalgamated_dep` assigned in all three branches; a one-liner after `endif` is simpler.

7. **`tests/meson.build:111`** — `link_with: kwargs.get('link_with', [])` is now dead code.

8. **`tests/meson.build:102`** — `app_cflags_single` re-reads `get_option('mode')` instead of using `buildmode`, and the `app_` prefix is misleading.

Not flagged (pre-existing): amalgamated mode generates no pkg-config file, `CLI/Timer.hpp` is never installed by Meson in any mode, and the "Failed to find any matches in CLI.hpp" warning matches CMake's behavior.

Recommendation: apply 1, 2 (which collapses the tests/meson.build workaround), and 3. Say the word and I'll make those changes.

meator:
I will react to some of the AI comments. But note that using AI on Meson is a really bad idea, (I assume that) common models do not have enough Meson training data.

> 1. **`single-include/meson.build:11` — `python` shebang breaks the amalgamated build.** `MakeSingleHeader.py` runs via `#!/usr/bin/env python`; on a machine with only `python3` you get `env: python: No such file or directory`. Since the PR makes the target `install: true`, this now fails by default. Fix: `[find_program('python3'), find_program('../scripts/MakeSingleHeader.py'), ...]`.

To quote Meson documentation:

> Meson will also autodetect scripts with a shebang line and run them with the executable/interpreter specified in it both on Windows (because the command invocator will reject the command otherwise) and Unixes (if the script file does not have the executable bit set). Hence, you must not manually add the interpreter while using this script as part of an array of commands. Since 0.50.0 if the "python3" program is requested and it is not found in the system, Meson will return its current interpreter.

The usage of `#!/usr/bin/env python` is not that big of a deal. Changing it to `python3` would lead to a (tiny bit) nicer behavior as described above, but that shouldn't be an issue.

If you really want to fix this, the proper way to do so is

```diff
diff --git a/scripts/MakeSingleHeader.py b/scripts/MakeSingleHeader.py
index d9cd7c5..83a643d 100755
--- a/scripts/MakeSingleHeader.py
+++ b/scripts/MakeSingleHeader.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from __future__ import print_function, unicode_literals
 
```

That script doesn't look like it would intentionally attempt to support python2, so this change shouldn't break anything.

> 2\. **`meson.build:68` — the amalgamated `cli11_dep` should carry the define and both include dirs.** [...]

AI's first claim, that amalgamated built CLI11 should export `CLI11_SINGLE_FILE` to dependants is interesting, so I tested it.

CLI11 CMake build system, which can (I believe) be considered CLI11's primary build system which should be used as reference when deciding behavior for other build systems like Bazel and Meson, has very rudimentary support for amalgamated/`CLI11_SINGLE_FILE` usage.

There is no indication whether amalgamated header file is available in the installed artifacts (there is nothing in pkg-config file and in the three installed `.cmake` files). These files also do not export the `CLI11_SINGLE_FILE` macro (which AI asks me to do for Meson).

The only detectable way of using amalgamated header (except for `__has_include` and configure time build system checks) is to vendor CLI11 with `add_subdirectory(CLI11)`. This seems to be one of the primary distribution methods supported by CLI11.

Vendored CLI11 does indeed export the `CLI11_SINGLE_FILE` macro, but not for the primary `CLI11::CLI11` target, it exports it when using the undocumented `CLI11_SINGLE` target.

I'll write more about this at the end of the comment.

> 3\. **`.github/workflows/build-tests.yml:132` — CI only runs `headeronly` mode**, so the path this PR fixes can regress silently. Add a matrix over `mode: [headeronly, amalgamated, precompiled]`.

The more CI and tests the better. No objections here.

> 4\. **`tests/meson.build:92`** — TimerTest and link_test_2 are hand-rolled outside the `testnames` loop; per-entry `dependencies`/`link_with` overrides in the dict already suffice and the copies already diverge (TimerTest lacks `depends: app_tgts`/`workdir`).

I took care to replicate CMake behavior. CMake also treats them specially (because they are special). I ommited some flags for individual targets when CMake also did not provide them. ~~But I rechecked it now and `TimerTest` should indeed have at least `app_tgts`, so I will recheck the code and fix any issues.~~ My changes are correct already. It is a special case because it needs `cli11_nonamalgamated_dep` and does not need `-DCLI11_SINGLE_FILE`.

> 8\. **`tests/meson.build:102`** — `app_cflags_single` re-reads `get_option('mode')` instead of using `buildmode`, and the `app_` prefix is misleading.

This is an inconsequential stylystic change which makes it less readable. I will not implement it.

I agree with the rest of AI's points and will ammend fixes soon.

# The amalgamated header situation
As I mentioned, I consider CMake the primary build system. If you really want to proceed with the proposed changes, I think CMake should be changed first to make things consistent.

The situation is complicated by the fact that CMake and Meson treat amalgamated header differently. CMake recognizes two states:

- normal build or precompiled build
- additionally build amalgamated header or not

Meson recognizes one state

- build normally _or_ build precompiled _or_ build amalgamated

This difference can cause confusion and problems down the line. I see several ways forward:

1. amalgamated header should not have native build system support, its main purpose should be for direct vendoring

   This means that CMake should still export nothing about `CLI11_SINGLE_FILE`, it should not even install the amalgamated file.

   AI's recommendation for Meson should not be followed.
2. amalgamated header should have native build system support

   I consider CMake the reference build system, so it should be changed first. An additional exported dependency (with appropriate `.cmake` and pkg-config files) should be added for amalgamated file. Dependants could therefore know whether amalgamated header is provided using standard dependency resolution mechanisms.

   `CLI11_SINGLE` should be documented better, perhaps it should be renamed to `CLI11::CLI11_SINGLE` for consistency with `CLI11::CLI11` (my main area of build system expertise is in Meson, I do not know whether this is a good change, because I do not know as much about CMake).

   Meson should split its one state into two to match CMake behavior. This can be considered a breaking change. Note that Meson has [good support for deprecating options](https://mesonbuild.com/Build-options.html#deprecated-options), so deprecating `mode=amalgamated` isn't hard.

   Meson should also export CLI11 amalgamated as a separate dependency.

   Bazel and any other build systems of CLI11 should follow suite. But I know nothing about Bazel, I can't provide any guidance here.

Both solutions assume improved documentation. I can work on that if there's interest. But I can't implement the CMake changes required for 2., since my knowledge of CMake is rudimentary.

meator:
Meson also fails to export `CLI11_COMPILE` macro for `mode=precompiled`, but it exports it in installed artifacts, so it's in a better state than amalgamated header.

henryiii:
> But note that using AI on Meson is a really bad idea, (I assume that) common models do not have enough Meson training data.

In general, what makes AI good is having ways to test things. Agentic AI is better at Rust than C, even though it has way more training data on C, because it runs the compiler and acts on the output, which is better and stricter. (Not true if there's no agentic loop, like when asking in chat only interfaces.) In general, any statement from AI that can't be proven (usually by the AI) is potentially hallucinated, but anything it can verify (assuming it does) is generally pretty good.

Anyway, AI is still better at Meson than I am, which isn't saying much. ;)

* (1) Yes, we probably should use python3 in the script, it's the official name of Python these days. python is not required to be present, python3 is. (Windows aside, but that uses a different system anyway).
* (2) Yes, cmake is the primary build system. Reworking support (and then Meson following) is quite reasonable. I can try to remember to do this (maybe we can open an issue).
* (4) Sure.
* (8) Not surprising, AI has a really bad idea of what is readable for humans. :)

Do you know why it's failing currently?

meator:
Even if I take what you say at face value, I doubt that any common AI models have Meson integration which would allow it to "test" things. And if the AI is just given a small VM to play around in with commands, I would trust AI to do proper testing even less. Properly testing a build system is not easy. It requires undetstanding the intricacies of different build environments, run environments, compilers etc.

If maintainers have limited capacity for maintaining the Meson build system, that could be an argument for its deprecation. Or it should be documented that it is maintained on a best-effort basis and that CMake is the "primary" build system (as you have confirmed). If you wish to document it, please do so in a really obvious place so that people like me can find it without combing through documentation or depleating resources to give that task to an AI. A good place would be a comment at the top of the `CMakeLists.txt` and `meson.build` top-level files.

> Windows aside, but that uses a different system anyway

Meson abstracts this away as explained in the Meson documentaton I quoted. That lets it bypass any Windows weirdness.

You did not respond to the most important point brought up in by comments, the disparity between Meson and CMake handling of CLI11's special build modes. Such changes may be better suited to a separate PR, but I think it is a worthwile change. The unoptimal implementation lead to the errors this PR wishes to fix. I'm willing to contribute, but a decision from maintainers is required to choose between the two solutions I provided or to devise something else.

I can make a more detailed writeup on the issue (perhaps by making a new GitHub issue) if requested.

> Do you know why it's failing currently?

https://github.com/CLIUtils/CLI11/pull/1452

